PR #8, opened 2023-07-04
Ans migration scripts

Kayanski:
This PR aims at providing simple diff and update scripts for the ANS_HOST variables.

It diffs from scraped asset lists (on github), diffs with the on-chain elements and updates if necessary

## TODO
- [ ] The pool update script is not perfectly functional especially in case updates are needed on a pooladdress (the pool update function doesn't work like the other ones)


Kayanski:
Should we remove the `channels`  file that doesn't have an associated scrapping file ? Or should we leave it as it was a.k.a manual update ?

CyberHoward:
> Should we remove the `channels` file that doesn't have an associated scrapping file ? Or should we leave it as it was a.k.a manual update ?

Yeah let's leave it as it was so we can still update it if we want.

CyberHoward (changes requested):
I'd like to see some tests for the diff logic. 

Maybe download the file and remove/add to it manually and then diff and assert? That way the test will also fail if the remote format/location changes.

Kayanski:
There is an issue left with pool data updates.
As there can be multiple pools with a single asset pair (and with multiple pool types as well), the updater script can't work as is. 
That's actually the way we handle pools that is at stake here. In the updater do we want to have only one pool registered in the ans, which is the pool with the most TVL ? That could be an excellent idea to get rid of this updater and pool choice issue

CyberHoward:
This functionality was moved to the [`abstract-name-service` repo ](https://github.com/AbstractSDK/abstract-name-service)

PR #13, opened 2023-07-05
Adapters should have crates.io dependencies

Kayanski:
This PR aims at allowing users to import abstract from a crates.io dependency and adapter from this git monorepo

### DIFF
--- a/adapter-packages/Cargo.toml
+++ b/adapter-packages/Cargo.toml
@@ -46,11 +46,11 @@ semver = "1.0"
 cw-semver = { version = "1.0" }
 cw-orch = { version = "0.13" }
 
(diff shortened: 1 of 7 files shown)


CyberHoward:
I'll just remove the path deps from adapters when publishing the crate

CyberHoward:
because if we keep the deps to crates the code won't compile

CyberHoward:
We're not merging this, we'll keep the adapter path deps for development and remove them when we create a release.

CyberHoward:
Does that work? I think the adapter integrations will also require it, sucks really hard tho

Kayanski:
Why are you saying this ? This works as is now, using abstract 0.16.1

CyberHoward:
Fixed in #14

PR #14, opened 2023-07-05
Deploy 0.17.0

CyberHoward:
Deployment of v0.17 and monorepo re-structuring.


applications/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["contracts/*"]
+members = ["contracts/apps/*", "packages/bundles/wyndex"]
```
Buckram123: YESSS All on-chain apps together now 🔥. Don't forget to include "contracts/adapters/*"

scripts/wasm-all.sh
```diff
@@ -4,60 +4,48 @@
 if [[ $(arch) == "arm64" ]]; then
 image="cosmwasm/rust-optimizer-arm64"
 workspace_image="cosmwasm/workspace-optimizer-arm64"
-abstract_image="abstractsdk/workspace-optimizer-arm64"
+abstract_image="abstractmoney/workspace-optimizer-arm64"
```
Buckram123, line 7: I guess workspace image not used anymore?

Buckram123 (commented):
Seems much better!

applications/Cargo.toml
Buckram123: And not sure about wyndex bundle

scripts/wasm-all.sh
CyberHoward: It's still the workspace image, it's just a fork that allows featured builds!

Buckram123: I meant if we plan to use `abstract_image` over `workspace_image` why keep both :laughing:

PR #16, opened 2023-07-07
Instantiate admins to native contracts

Buckram123:


cyberhoward (commented):
I think I would prefer the admin to be required in the deploy_on function and not be optional in the Instantiate messages. 

It's still very easy to set the admin to the sender in that case of we want to, just by providing the address of the current sender in the deploy function. WDYT?

framework/contracts/native/account-factory/src/contract.rs
```diff
@@ -33,8 +33,12 @@ pub fn instantiate(
     cw2::set_contract_version(deps.storage, ACCOUNT_FACTORY, CONTRACT_VERSION)?;
 
     CONFIG.save(deps.storage, &config)?;
-    // Set up the admin as the creator of the contract
-    cw_ownable::initialize_owner(deps.storage, deps.api, Some(info.sender.as_str()))?;
+    // Set up the admin
+    let owner = match msg.admin.as_deref() {
```
Kayanski, line 37: This code is repeated 4 times, could we have a function instead that initializes the owner base on a optional admin field ?

Kayanski (approved):

framework/contracts/native/account-factory/src/contract.rs
Buckram123: True, thanks

Buckram123:
> I think I would prefer the admin to be required in the deploy_on function and not be optional in the Instantiate messages.

Agree, `admin: None` could also mean no-admin, which is not in our case

PR #18, opened 2023-07-10
Sub Account Logic

Kayanski:
This PR aims at creating a subaccount logic on the abstract accounts

These sub accounts are normal accounts that are owned by 2 entites : 
1. The account manager
2. The owner of the original account

It should be possibe to 
- [x] Create a sub_account from the manager directly
- [x] Execute an arbitrary message on the sub-account from the manager
--> ExecuteOnSubAccount
- [x] Execute an arbitrary message on the sub-account from the owner of the original account
--> Modified admin check logic on the manager only. 
--> Added a GovernanceDetails member to account for that subaccount logic
--> Should we also change things on the proxy ?
- [ ] Query all sub-accounts on the account (to avoid having to keep track) ?

# Checklist

- CI is green.
- Changelog updated.


CyberHoward (commented):
The way I was thinking about the sub-account execution was to not have it as a separate endpoint or account type but just to swap out the `assert_admin` logic with a function that: 

1. First asserts admin like normal
2. Queries the `owner` of the caller (expected to resolve if it's a sub-account, else fails)
3. (Repeat 2 until it fails?) I.e. the top-level account owner. 
4. assert that the sender is that owner (like you already do) 

I see a scenario where we need pass-through execution only if we don't do the recursive admin lookup. So basically if we remove the proxied message execution and replace it with the recursive admin lookup we should be gold. 

Thought?

I do like the `SubAccountMonarchy` type tho, maybe we can just have it be `SubAccount` with a `owner_account_manager` field?

framework/contracts/account/manager/src/commands.rs
```diff
@@ -808,22 +779,43 @@ pub fn update_internal_config(deps: DepsMut, info: MessageInfo, config: Binary)
...
-        GovernanceDetails::SubAccountMonarchy { monarch } => {
-            // We try to query the ownership of the monarch account
-            let Ownership::<String> { owner, .. } = deps
-                .querier
-                .query_wasm_smart(monarch, &QueryMsg::Ownership {})?;
-            // We assert that the sender is the owner
-            if *sender != owner.unwrap() {
-                return Err(ManagerError::Admin(cw_controllers::AdminError::NotAdmin {}));
+        GovernanceDetails::SubAccount {
+            owner_account_manager,
```
CyberHoward, line 801: hmm this is confusing. 

`owner_account_manager` -> The owner of the account manager (which?)
OR 
`owner_account_manager` -> The manager of the account that owns this account (whit we want) 

Maybe there's a better name to apply here

Kayanski: I can replace that with `manager` only, because that's what it indicates --> This is a subaccount, managed by an account manager, which we should be able to query an owner of

framework/contracts/account/manager/src/commands.rs
```diff
@@ -735,6 +781,56 @@ pub fn update_internal_config(deps: DepsMut, info: MessageInfo, config: Binary)
...
+                        // If not, we try again with the queried owner
+                        current = deps.api.addr_validate(&owner)?
+                    }
+                } else {
+                    return Err(ManagerError::Admin(cw_controllers::AdminError::NotAdmin {}));
+                }
+                if {
+                    i += 1;
+                    i
+                } > MAX_ADMIN_RECURSION
```
Buckram123, line 822: I would prefer for it to even not possible to create >5 level deep sub-account, we can save "depth" in manager config

framework/contracts/account/manager/src/commands.rs
```diff
@@ -735,6 +781,56 @@ pub fn update_internal_config(deps: DepsMut, info: MessageInfo, config: Binary)
...
+    }
+
+    let account_info = INFO.load(deps.storage)?;
+    match account_info.governance_details {
+        // If the account has SubAccount governance, the owner of the manager also has admin rights over this account
+        GovernanceDetails::SubAccount { manager } => {
+            // We try to query the ownership of the manager monarch account if the first query failed
+            let mut current = manager;
+            let mut i = 0;
+            loop {
```
Buckram123, line 806: Why not `while i < 5`

Buckram123 (changes requested):
Maybe saving how deep this sub-account can be useful, but not sure, wdyt?

CyberHoward (changes requested):

framework/contracts/account/manager/src/commands.rs
Kayanski: I'm not sure that's the way to go.
Because we are using the cw-ownable library for ownership, we allow for ownership of accounts indirectly and this is not followed inside abstract. I don't think enforcing subaccount maximum depth helps us here.
And we are not limiting the depth at runtime, only the depth at which you can call a subaccount action.
WDYT ?

CyberHoward (approved):

PR #21, opened 2023-07-12
Set adapter addr on install_adapter

Buckram123:
- Sets and returns Addr of installed adapter
- Tests for install_adapter functionality with tests for install_app

# Checklist

- CI is green.
- Changelog updated.


CyberHoward (commented):
I think we should get rid of / make private the `install_module` function on the account, so there's only `install_app` or `install_adapter`.

CyberHoward (approved):

PR #22, opened 2023-07-18
Integrate dao-dao with Abstract

CyberHoward:
This integration allows dao's to be merged with abstract accounts. 

## Process

The process is triggered by a dao migration. During the migration the required account-specific data is set on the proxy by the account factory (account-id). In essence only the version control contract should be provided in the migrate message as the other data can be queried from that address. A new endpoint on the factory should be introduced that allows a proxy contract to register itself to the account registry. The validity of the proxy contract can be checked by asserting that the contract's code-id is available in the registry under a specific key (abstract:dao-proxy).

The dao proxy contract is a mix of the dao-dao and abstract proxy contracts. During migration a new manager is created and its owner is set to the proxy contract. This allows for dao-managed account execution as well as sub-accounts owned by the dao (proxy).

The beauty about this way of implementing the integration is that the code to maintain is minimal (we import and re-use all the execute and query logic). It's also fully compatible with both codebases as the entry point messages are merged.

## Goal

Dao's will be able to manage and monetize dao-owned code and create and manage applications as sub-accounts.

### DIFF
--- a/framework/contracts/account/proxy/src/lib.rs
+++ b/framework/contracts/account/proxy/src/lib.rs
@@ -1,7 +1,7 @@
-mod commands;
+pub mod commands;
 pub mod contract;
-mod error;
-mod queries;
+pub mod error;
+pub mod queries;
 pub mod reply;
 
 #[cfg(test)]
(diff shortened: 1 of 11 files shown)


Kayanski:
We have an issue with this integration, we need to make sure that storage fields don't overlap with each other between the 2 contract implementations. Should we prefix all of our contract storage ?

adairrr:
<https://linear.app/abstract-sdk/issue/ABS-102>

PR #23, opened 2023-07-19
Feature/Query handler for module data

Buckram123:
During instantiation a ModuleData struct is stored.

This data is not exposed through any query yet. So this PR aims adding a variant to the app and adapter base query messages that retrieves that returns that data. 

# Checklist

- CI is green.
- Changelog updated.


framework/packages/abstract-core/src/adapter.rs
```diff
@@ -121,6 +124,17 @@ pub enum BaseQueryMsg {
     /// Returns [`AuthorizedAddressesResponse`].
     #[returns(AuthorizedAddressesResponse)]
     AuthorizedAddresses { proxy_address: String },
+    /// Returns module data
+    #[returns(AdapterModuleDataResponse)]
+    BaseModuleData {},
```
adairrr, line 129: Might consider naming this `ModuleData` instead - what value does `Base` add?

framework/packages/abstract-core/src/adapter.rs
```diff
@@ -121,6 +124,17 @@ pub enum BaseQueryMsg {
...
+}
+
+#[cosmwasm_schema::cw_serde]
+pub struct AdapterModuleDataResponse {
+    pub module: String,
```
adairrr, line 134: Perhaps `module_id` would be more descriptive.

adairrr (changes requested):

adairrr:
This is going to be really helpful for the frontend!

framework/packages/abstract-core/src/adapter.rs
Buckram123: We want to avoid name collision for schema generation, for example if we will have some adapter, which will have `ModuleData` query it won't be able to generate schema. IMO `ModuleData` is pretty generic name, wdyt?

framework/packages/abstract-core/src/app.rs
```diff
@@ -75,6 +78,17 @@ pub enum BaseQueryMsg {
     /// Returns the admin.
     #[returns(AdminResponse)]
     BaseAdmin {},
+    /// Returns module data
+    #[returns(AppModuleDataResponse)]
+    BaseModuleData {},
+}
+
+#[cosmwasm_schema::cw_serde]
+pub struct AppModuleDataResponse {
```
CyberHoward, line 87: `AppModuleDataResponse` and `AppModuleDataResponse` are the same and represent the same underlying structure (ModuleData). Can we instead add a single response type along with de definition of the ModuleData struct and return that in both implementation?

framework/packages/abstract-core/src/adapter.rs
CyberHoward: I think `ModuleData` is quite specific so agree with Adair on this one :)

CyberHoward (changes requested):
One little change and we good :)

CyberHoward (approved):

CyberHoward:
@Buckram123 Any idea why the modules tests fail? I can't reproduce because i'm on M1...

Buckram123:
> @Buckram123 Any idea why the modules tests fail? I can't reproduce because i'm on M1...

Oh yeah, we don't have wasms with `install_modules` field built! In which PR should I build them?

CyberHoward:
> > @Buckram123 Any idea why the modules tests fail? I can't reproduce because i'm on M1...
> 
> Oh yeah, we don't have wasms with `install_modules` field built! In which PR should I build them?

Ahh damn, you can add them here if you want! Or create a PR with this branch as a base.

PR #24, opened 2023-07-20
Update osmosis cw-staking adapter to query the pool id via ans pool

adairrr:
# Checklist

- CI is green.
- Changelog updated.


integrations/osmosis/packages/abstract-adapter/src/lib.rs
```diff
@@ -1,4 +1,4 @@
 pub const OSMOSIS: &str = "osmosis";
-pub const AVAILABLE_CHAINS: &[&str] = &["osmosis"];
+pub const AVAILABLE_CHAINS: &[&str] = &["osmosis", "osmo-test"];
```
Buckram123, line 2: :rocket:

Buckram123 (approved):
Looks good, would love to see that in action! Do you think we should cover this new feature by tests?

integrations/osmosis/packages/abstract-adapter/src/lib.rs
adairrr: And we should probably use `abstract_sdk::core::registry::OSMOSIS` here, as long as we include `osmo-test` there.

adairrr:
> Looks good, would love to see that in action! Do you think we should cover this new feature by tests?

Yes! We should definitely be testing this even if it's just some unit tests with mocked ans data. 

We also should likely need to add some better error messages for when the pool isn't found or just otherwise to clarify the stack trace.

Buckram123:
Update: we need a way to replace LockupQuerier, because it seems like any `/osmosis.lockup.Query/*` is not allowed for contracts on osmosis :cry:

CyberHoward:
@Buckram123 `/osmosis.lockup.Query/*` is now enabled, could we move forward with finishing this pr?

CyberHoward:
@Buckram123 is this still relevant?

Buckram123:
Merged in #30

PR #28, opened 2023-07-27
Bump to 0.17.2 to add neutron to abstract_sdk registry

adairrr:
This change adds neutron (and archway) to the registry and publishes 0.17.2.

### DIFF
--- a/framework/packages/abstract-core/src/registry.rs
+++ b/framework/packages/abstract-core/src/registry.rs
@@ -19,9 +19,10 @@ pub const ICS20: &str = "ics-20";
 
 // chain-id prefixes based on `https://cosmos.directory/`
 pub const JUNO: &[&str] = &["juno", "uni"];
-pub const OSMOSIS: &[&str] = &["osmosis", "osmo"];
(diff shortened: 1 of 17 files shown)


framework/packages/abstract-core/src/registry.rs
```diff
@@ -19,10 +19,11 @@ pub const ICS20: &str = "ics-20";
 
 // chain-id prefixes based on `https://cosmos.directory/`
 pub const JUNO: &[&str] = &["juno", "uni"];
-pub const OSMOSIS: &[&str] = &["osmosis", "osmo"];
+pub const OSMOSIS: &[&str] = &["osmosis", "osmo", "osmo-test"];
```
Buckram123, line 22: interesting, so "osmo" prefix doesn't cover "osmo-test"?

Buckram123 (approved):
LGTM

Buckram123:
Oopsie, came a little late to the party

framework/packages/abstract-core/src/registry.rs
adairrr: No because of the way that it's "calculated" it just removes the string after the last hyphen.

PR #31, opened 2023-07-28
Add ETF module to monorepo

adairrr:
This change migrates the ETF app from the old repository (https://github.com/AbstractSDK/apps/tree/main/contracts) to here


modules/contracts/apps/etf/src/state.rs
```diff
@@ -0,0 +1,14 @@
...
+/// State stores LP token address
+/// BaseState is initialized in contract
+#[cosmwasm_schema::cw_serde]
+pub struct State {
+    pub share_token_address: Addr,
+    pub manager_addr: Addr,
+}
+
+pub const STATE: Item<State> = Item::new("\u{0}{5}state");
+pub const FEE: Item<Fee> = Item::new("\u{0}{3}fee");
```
Buckram123, line 14: Would be nice to add comment why we start state key with NULL

Buckram123 (changes requested):
Happy to finally see this contract! :rofl:

modules/contracts/apps/etf/rustfmt.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# unstable... should we require `rustup run nightly cargo fmt` ?
+# or just update the style guide when they are stable?
+#fn_single_line = true
+#format_code_in_doc_comments = true
+#overflow_delimited_expr = true
+#reorder_impl_items = true
+#struct_field_align_threshold = 20
+#struct_lit_single_line = true
+#report_todo = "Always"
```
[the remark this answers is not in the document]
adairrr: ha! Will remove these since they're just leftover from the previous package.

modules/contracts/apps/etf/src/handlers/instantiate.rs
```diff
@@ -0,0 +1,60 @@
...
+pub fn instantiate_handler(
+    deps: DepsMut,
+    env: Env,
+    _info: MessageInfo,
+    _etf: EtfApp,
+    msg: EtfInstantiateMsg,
+) -> EtfResult {
+    let state: State = State {
+        share_token_address: Addr::unchecked(""),
+        manager_addr: deps.api.addr_validate(msg.manager_addr.as_str())?,
```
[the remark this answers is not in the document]
adairrr: That would certainly be best practice. Thanks for pointing it out.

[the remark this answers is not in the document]
adairrr: We should honestly convert this over to using token-factory anyway...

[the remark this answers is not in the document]
adairrr: Perhaps we should wait until #36 is merged and then use it as a usage example?

modules/contracts/apps/etf/src/state.rs
adairrr: Is it common practice to do so? I haven't seen that elsewhere 👀

modules/contracts/apps/etf/src/handlers/instantiate.rs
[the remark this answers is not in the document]
adairrr: > > We should honestly convert this over to using token-factory anyway...
> 
> Is token factory supported on every chain we use?

Not Archway :/

modules/contracts/apps/etf/src/state.rs
Buckram123: > Is it common practice to do so? I haven't seen that elsewhere eyes

Same, haha. Saw just random bytes before(to decrease key length), but not a normal key starting with NULL

adairrr: Ah, yeah that used to be required in CW denoting how long the key was. state = 5 chars, fee = 3 chars.

Buckram123: Oh wow, how old is this contract

adairrr: > Oh wow, how old is this contract

Howard wrote it as one of the first Abstract modules >1 year ago.

CyberHoward:
Lets merge this as an example contract for now and leave it as it is until there's a use-case for it.

cyberhoward (approved):

PR #33, opened 2023-08-01
Separate kujira integration

Buckram123:
Fixes ABS-60
# Checklist

- CI is green.
- Changelog updated.


framework/packages/utils/src/util.rs
```diff
@@ -28,5 +28,7 @@ pub fn coins_in_assets(assets: &[Asset]) -> Vec<Coin> {
             coins.push(Coin::new(asset.amount.u128(), denom.clone()));
         }
     }
+    // TODO: Hopefully some day tokenfactory coins can be last and we can get rid of this nonsense
+    coins.sort_unstable_by_key(|c| std::cmp::Reverse(c.denom.len()));
```
adairrr, line 32: factory/kujira1e7hxytqdg6v05f8ev3wrfcm5ecu3qyhl7y4ga73z76yuufnlk2rqd4uwf4/d
factory/kujira1ltvwg69sw3c5z99c6rr08hal7v0kdzfxz07yj5/demo
ibc/57503D7852EF4E1899FE6D71C5E81D7C839F76580F86F21E39348FC2BC9D7CE2
ukuji

adairrr (changes requested):

adairrr:
<https://linear.app/abstract-sdk/issue/ABS-60>

framework/packages/utils/src/util.rs
```diff
@@ -28,7 +28,7 @@ pub fn coins_in_assets(assets: &[Asset]) -> Vec<Coin> {
             coins.push(Coin::new(asset.amount.u128(), denom.clone()));
         }
     }
-    // TODO: Hopefully some day tokenfactory coins can be last and we can get rid of this nonsense
-    coins.sort_unstable_by_key(|c| std::cmp::Reverse(c.denom.len()));
+    // https://github.com/cosmos/cosmos-sdk/blob/d5b40fc427f530b5ab078c61531d521b85505a1e/types/coin.go#L259-L261
+    coins.sort_unstable_by(|a, b| a.denom.cmp(&b.denom));
```
Buckram123, line 32: At what level do we want that sorting to happen?

adairrr: Oh wow. Crazy that this is actually baked into the SDK as well. What do you mean which level?

integrations/kujira/packages/abstract-adapter/src/dex.rs
```diff
@@ -321,3 +321,238 @@ pub fn decimal2decimal256(dec_value: Decimal) -> StdResult<Decimal256> {
...
+            } => {
+                assert_eq!(max_slippage, None);
+                assert_eq!(callback, None);
+            }
+            _ => panic!("Expected a provide liquidity variant"),
+        }
+
+        let funds = get_wasm_funds(msgs[0].clone());
+        assert_eq!(funds.len(), 2);
+        assert_eq!(funds[0], coin(amount_demo, DEMO),);
```
adairrr, line 521: Weird that there's a trailing comma here and the linter didn't yell at you.

modules/contracts/adapters/dex/examples/deploy.rs
```diff
@@ -12,8 +12,12 @@ use semver::Version;
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 fn deploy_dex(network: ChainInfo) -> anyhow::Result<()> {
```
adairrr, line 14: @Kayanski will this deploy command automatically select the correct feature-flagged build? How are we going about that?

Kayanski: No it's going to select whatever first artifacts it founds in the directory (alphabetically I guess). So that's not good. 
@Buckram123 had a solution somewhere, don't remember what he proposed

Kayanski: Oh, found it. We wanted to have an interface that is not automatically generated and that would take an extra field an instantiation for specifying the chain it was supposed to go on. 
My solution was using environment variables, but that's not Great

integrations/kujira/packages/abstract-adapter/src/dex.rs
Kayanski: This sorting stuff is craaaazy, why didn't we know that ? That's weird

framework/packages/utils/src/util.rs
Buckram123: Agree, nevermind, it's gonna be too shady if we modify the message in any place.

modules/contracts/adapters/dex/examples/deploy.rs
Buckram123: > Oh, found it. We wanted to have an interface that is not automatically generated and that would take an extra field an instantiation for specifying the chain it was supposed to go on. My solution was using environment variables, but that's not Great

We still use not the automatically generated the solution was:
image

Kayanski (approved):

modules/contracts/adapters/dex/examples/deploy.rs
Buckram123: @adairrr Thoughts on that approach? We will have to wait a bit longer due to block_info request tho. Should be still faster than manually setting it tbh

framework/packages/utils/src/util.rs
CyberHoward: So we want this sort here? Why is it required?

modules/contracts/adapters/dex/src/adapter.rs
```diff
@@ -212,7 +212,7 @@ pub trait DexAdapter: AbstractNameService + Execution {
         let pair_address = exchange.pair_address(
             deps,
             ans.host(),
-            (paired_assets.swap_remove(0), paired_assets.swap_remove(1)),
+            (paired_assets.swap_remove(0), offer_asset.name.clone()),
```
CyberHoward, line 215: It is unclear if this method expects the user to provide the offer asset with the paired assets or not.

framework/packages/utils/src/util.rs
Buckram123: > So we want this sort here? Why is it required?

There is a link in a comment, so if coins not sorted transaction will fail

PR #34, opened 2023-08-02
Added namespace and base asset definitions in account creation

Kayanski:
This PR aims at allowing users to claim their namespace and add their base asset directly from the create account message.

# Checklist

- CI is green.
- Changelog updated.

### DIFF
--- a/framework/contracts/native/account-factory/src/commands.rs
+++ b/framework/contracts/native/account-factory/src/commands.rs
@@ -1,4 +1,5 @@
-use abstract_core::objects::ABSTRACT_ACCOUNT_ID;
+use abstract_core::objects::price_source::UncheckedPriceSource;
+use abstract_core::objects::{AssetEntry, ABSTRACT_ACCOUNT_ID};
 use abstract_core::{manager::ExecuteMsg, objects::module::assert_module_data_validity};
 use cosmwasm_std::{
     to_binary, wasm_execute, Addr, CosmosMsg, DepsMut, Empty, Env, MessageInfo, QuerierWrapper,
@@ -32,6 +33,7 @@ pub const CREATE_ACCOUNT_MANAGER_MSG_ID: u64 = 1u64;
 pub const CREATE_ACCOUNT_PROXY_MSG_ID: u64 = 2u64;
 
 /// Function that starts the creation of the Account
+#[allow(clippy::too_many_arguments)]
 pub fn execute_create_account(
     deps: DepsMut,
     env: Env,
@@ -40,6 +42,8 @@ pub fn execute_create_account(
     name: String,
     description: Option<String>,
     link: Option<String>,
+    namespace: Option<String>,
+    base_asset: Option<AssetEntry>,
 ) -> AccountFactoryResult {
     let config = CONFIG.load(deps.storage)?;
 
@@ -58,6 +62,11 @@ pub fn execute_create_account(
             account_manager_address: None,
             manager_module: Some(module.clone()),
             proxy_module: None,
+
+            additional_config: AdditionalContextConfig {
+                namespace,
+                base_asset,
+            },
         },
     )?;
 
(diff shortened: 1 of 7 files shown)


framework/contracts/native/account-factory/src/commands.rs
```diff
@@ -32,6 +33,7 @@ pub const CREATE_ACCOUNT_MANAGER_MSG_ID: u64 = 1u64;
 pub const CREATE_ACCOUNT_PROXY_MSG_ID: u64 = 2u64;
 
 /// Function that starts the creation of the Account
+#[allow(clippy::too_many_arguments)]
```
Buckram123, line 36: We have 6 arguments of create_account action, I guess we can pack them into struct, wdyt?

framework/contracts/native/account-factory/src/commands.rs
```diff
@@ -224,6 +233,36 @@ pub fn after_proxy_add_to_manager_and_set_admin(
...
+            Ok::<_, StdError>(CosmosMsg::Wasm(WasmMsg::Execute {
+                contract_addr: proxy_address.to_string(),
+                funds: vec![],
+                msg: to_binary(&ProxyExecMsg::UpdateAssets {
+                    to_add: vec![(a, UncheckedPriceSource::None)],
+                    to_remove: vec![],
+                })?,
+            }))
+        })
+        .transpose()?;
```
Buckram123, line 249: Nice

framework/packages/abstract-core/src/native/account_factory.rs
```diff
@@ -74,10 +83,14 @@ pub enum ExecuteMsg {
         governance: GovernanceDetails<String>,
         // Account name
         name: String,
+        // Optionally specify a base asset for the account
```
Buckram123, line 86: What's your thought on `CreateAccount(CreateAccountMsg)`?

framework/contracts/native/version-control/src/commands.rs
```diff
@@ -296,7 +296,11 @@ pub fn claim_namespace(
     // verify account owner
     let account_base = ACCOUNT_ADDRESSES.load(deps.storage, account_id)?;
     let account_owner = query_account_owner(&deps.querier, &account_base.manager, account_id)?;
-    if msg_info.sender != account_owner {
+
+    let account_factory = FACTORY.get(deps.as_ref())?.unwrap();
+
+    // The account owner as well as the account factory contract are able to claim namespaces
+    if msg_info.sender != account_owner && msg_info.sender != account_factory {
```
Buckram123, line 303: nice

Buckram123 (approved):
Really like that change. Would be nice to pack big ExecuteMsg variants into a single struct, not sure if it's blocked by anything tho

framework/packages/abstract-core/src/native/account_factory.rs
Kayanski: The issue is that we need to verify the governance details, so that's a bit of a hassle when unpacking an repacking the msg

Kayanski: We can do the governance verification inside the execute function, that's better

Kayanski: The issue is that it's not really compatible with cw_orch !

Buckram123: > The issue is that it's not really compatible with cw_orch !

Ah, I was wondering if that would be an issue, thanks. We should add support on those in the future then

Kayanski: Yes. However, it's not really that simple as the proc_macro engine doesn't work like that and cannot interpret things that are not defined in the structured that is being parsed

adairrr (approved):

PR #37, opened 2023-08-07
Abstract IBC implementation

Kayanski:
This PR aims at introducing IBC capabilities into Abstract.

Refer to [IBC.md] (https://github.com/AbstractSDK/abstract/blob/3a138adb4ce54dcf245f9758e31aa2ea2b79dc94/framework/development/IBC.md) for details on how this implementation functions

The center of this implementation are the 2 ibc-client and ibc-host contracts that work in tandem in providing ibc capabilities to users

# Next steps
- [x] Integrate  [Polytone](https://github.com/DA0-DA0/polytone/)  inside the Abstract IBC implementation.
- [x] Testing the IBC account interaction (tests/src/ibc.rs) `cd framework/tests && cargo run --bin setup && cargo run test test_create_ibc_account` 
- [ ] Integrate actions with sending funds (ibc-hooks, callbacks) (Not doable for now, we need [IBC 7.3.0](https://github.com/cosmos/ibc-go/releases) integrated in cosmwasm and on our target chains)
  This is coming soon, [Terra will integrate it in their next release](https://github.com/terra-money/core/pull/162/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R19)

# Documentation

The IBC execution flow is outlined here : https://www.notion.so/IBC-flow-08d769e142f74b6db782953a7a206874
A brief visual overview of the contract structure can be found here : https://www.figma.com/file/4v9l43ELlElmFvrlewGuqH/Abstract-IBC?type=whiteboard&node-id=5-14&t=EeiHnS0ANp0XTFrX-0
The deployment of abstract and the new function to deploy connectors between chains can help one understand the steps needed to connect 2 abstract remote instances

# Checklist

- CI is green.
- Changelog updated.


framework/packages/abstract-sdk/src/apis/ibc.rs
```diff
@@ -54,10 +56,126 @@ pub struct IbcClient<'a, T: IbcInterface> {
...
+            },
+            vec![],
+        )?
+        .into())
+    }
+
+    */
+
+    /// A simple hepler to create and register a distant account
+    pub fn create_distant_account(
```
adairrr, line 82: How did we come to `distant` as opposed to `remote` 😂

modules/contracts/adapters/cw-staking/src/handlers/execute.rs
```diff
@@ -52,14 +53,15 @@ fn handle_local_request(
 }
 
 /// Handle a request that needs to be executed on a remote chain
+/// TODO, this doesn't work as is. This should be corrected when working with ibc hooks ?
 fn handle_ibc_request(
```
CyberHoward, line 57: How do we see IBC requests being managed by these adapters?

IMO the chain that the user would like to use should be provided by the user (kinda like it is right now). 
Thoughts?

integrations/astroport/packages/abstract-adapter/Cargo.toml
CyberHoward, line 1: Do you have ftoml or something similar? I've seen this diff pop up multiple times and idk why.

framework/tests/src/tokenfactory.rs
```diff
@@ -0,0 +1,139 @@
...
+    prelude::{interchain_channel_builder::InterchainChannelBuilder, Daemon, TxHandler},
+    starship::Starship,
+    state::ChainState,
+};
+use ibc_relayer_types::core::ics24_host::identifier::PortId;
+use tokio::runtime::Runtime;
+
+/// Creates a new denom using the token factory module.
+/// This is used mainly for tests, but feel free to use that in production as well
+pub async fn create_denom(daemon: &Daemon, token_name: &str) -> Result<(), DaemonError> {
```
CyberHoward: Could you elaborate a bit on why this was added?

framework/tests/src/ibc.rs
```diff
@@ -0,0 +1,215 @@
...
+    // We need to register the ibc client as a module of the manager (account specific)
+    origin_abstract
+        .account
+        .manager
+        .install_module(IBC_CLIENT, &Empty {}, None)?;
+
+    // Now we send a message to the client saying that we want to create an account on osmosis
+    let register_tx = origin_abstract.account.manager.exec_on_module(
+        to_binary(&abstract_core::proxy::ExecuteMsg::IbcAction {
+            msgs: vec![abstract_core::ibc_client::ExecuteMsg::Register {
```
CyberHoward: Did we make a decision on removing this requirement? I think having the ability to do it is useful but if you perform an action on a remote account that doesn't exist yet it should be created for the user.

framework/tests/src/bin/token_bridge.rs
```diff
@@ -0,0 +1,123 @@
+// This scripts helps create a test environment for sending native tokens over an IBC connection between 2 chains
```
CyberHoward, line 1: Lol, cw-orch + starship = OP

CyberHoward:
As discussed during stand-up the size of this PR makes it hard to review. Would it be possible to split it up into 

- Core changes: All the changes that were made to the contracts but excluding the IBC contracts. (for example the changes to `AccountId`) 
- IBC contracts: All the IBC contracts
- IBC example tests:  I saw you added a `tests` directory for IBC testing specifically. I think this could use a PR on its own to see how we can streamline IBC testing. 

What do you think?

framework/packages/abstract-core/src/proto/ibc.rs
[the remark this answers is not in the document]
Kayanski: This was done because the dependencies we have (cosmrs for isntance) don't include the full proto OR are not compatible with proto tooling (Msg trait for instance). 
The ibc one was done because the memo field is absent from current repos.

This is supposed to be a temporary fix for (just like tokenfactory for instance) until maintainers include the necessary traits or fields. 
This is only used in the ICS20 case though and not used here for now

integrations/astroport/packages/abstract-adapter/Cargo.toml
Kayanski: I guess is cal like so because I added a workspace member and thus the string is too long to fit on one line !

framework/tests/src/ibc.rs
Kayanski: Yes, working on that now

framework/tests/src/bin/token_bridge.rs
Kayanski: it is ! Creating such testing evnironment was so easy using cw-orch !

framework/contracts/native/ibc-client/src/commands.rs
```diff
@@ -51,27 +59,109 @@ pub fn execute_update_config(
...
+    // auth check
+    ADMIN.assert_admin(deps.as_ref(), &info.sender)?;
+
+    // We need to make sure we are not over-writing anything
+    // TODO
+
+    let note = deps.api.addr_validate(&note)?;
+    POLYTONE_NOTE.save(deps.storage, &chain, &note)?;
+    REVERSE_POLYTONE_NOTE.save(deps.storage, &note, &chain)?;
+    REMOTE_HOST.save(deps.storage, &chain, &host)?;
```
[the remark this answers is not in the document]
CyberHoward: I'm not sure if a note can make multiple connections to different chains or if it's one contract to connect to one chain.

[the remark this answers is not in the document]
Kayanski: I am sure that a note can only connect to a single chain (just like a voice)

framework/contracts/native/ibc-client/src/commands.rs
```diff
@@ -51,27 +59,109 @@ pub fn execute_update_config(
...
+                vec![],
+            )?
+            .into()],
+            callback: callback_request,
+            timeout_seconds: env
+                .block
+                .time
+                .plus_seconds(PACKET_LIFETIME)
+                .seconds()
+                .into(),
```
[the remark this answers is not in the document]
Kayanski: Thank you

framework/packages/abstract-core/src/native/ibc_host.rs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+#[cosmwasm_schema::cw_serde]
+pub struct MigrateMsg {}
+
+#[cosmwasm_schema::cw_serde]
+#[non_exhaustive]
+pub enum InternalAction {
+    /// Registers a new account from a remote chain
+    Register {
```
CyberHoward, line 71: I think it would be really cool to add two things here. 

1: The ability to provide what modules to install on creation as well as to set a base account (all the fields that the create_account endpoint exposes on the factory)

2: The ability to trigger actions on any set of traces. I.e. I want to create an ICAA on Juno and Terra but there's no direct connection from Terra to my chain. So I call `Register`` with a set of traces on which I want to register accounts.

Another usecase is calling an action on a module that's installed on all my interchain accounts. 

The idea would be to have something like:

```rust
struct InterchainAbstractAccountAction {
    // HashSet of traces that the action should be performed for.
    // When the host receives this it constructs the trace for the current environment (trace of client + local chain name)
    // If that trace is in the set, execute the action (could be registration or any type of action)
    // If traces is not empty, forward the call to the local client which will forward it to the other chains if possible.
    traces: HashSet<AccountTrace>
    // Action to take on each trace for this account.
    action: HostAction
}
```

So basically we want to solve the problem of "I have an action I want to perform on chain X, Y and Z in one transaction."

1. Define the action (HostAction)
2. Define the chains (Traces)
3. Execute what you can locally and propagate the rest

framework/contracts/native/account-factory/src/error.rs
```diff
@@ -37,6 +37,12 @@ pub enum AccountFactoryError {
     #[error("No payment received")]
     NoPaymentReceived {},
 
+    #[error("Can not create remote accounts without configured IBC host.")]
+    IbcHostNotSet,
+
```
Buckram123, line 42: nit: let's use universal format of no args error 
```suggestion
    #[error("Can not create remote accounts without configured IBC host.")]
    IbcHostNotSet {},

```

framework/tests/src/setup.rs
```diff
@@ -0,0 +1,80 @@
...
+    #[test]
+    fn ibc_setup() -> AnyResult<()> {
+        logger_test_init();
+
+        let rt = Runtime::new()?;
+        let sender = Addr::unchecked("sender");
+        let mock_interchain = MockInterchainEnv::new(vec![(JUNO, &sender), (STARGAZE, &sender)]);
+
+        // We just verified all steps pass
+        let (abstr1, abstr2) = ibc_abstract_setup(&rt, &mock_interchain, JUNO, STARGAZE)?;
```
Buckram123, line 66: this is so cool!

Buckram123 (commented):
Awesome job!

framework/packages/abstract-core/src/objects/chain_name.rs
```diff
@@ -32,6 +37,20 @@ impl ChainName {
...
+        // We verify there is exactly two elements (no more chain delimiters than 1 and exactly one)
+        ensure_eq!(chain.len(), 2, AbstractError::Assert(
+            format!("An IBC ANS entry should contain a chain identifier followed by the chain delimiter {} ", CHAIN_DELIMITER)
+        ));
+        // Then we check the chain name
+        let chain_name = Self(chain[0].to_string());
+        chain_name.verify()?;
+        Ok(chain_name)
+    }
+
```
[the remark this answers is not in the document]
Kayanski: The format is `juno>wyndex` so constructing the chain name is with the first entry. Yeah it was a helper for a functionality in modules, that we decided to push to another PR (later)

framework/contracts/native/ibc-host/tests/integration.rs
```diff
@@ -0,0 +1,191 @@
...
+    let abstr = Abstract::load_from(chain.clone())?;
+
+    let account_sequence = 1;
+    let chain = "juno";
+
+    // We need to set the sender as the proxy for juno chain
+    admin_abstr
+        .ibc
+        .host
+        .register_chain_proxy(chain.into(), sender.to_string())?;
```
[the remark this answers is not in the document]
Kayanski: We don't want to connect chains that don't have the same function, like a testnet and a mainnet, or else. I don't understand what you mean here ?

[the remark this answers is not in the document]
Buckram123: The question is: do we register the chain on which ibc-host on? And it will still have a "trace"?

framework/packages/abstract-core/src/objects/chain_name.rs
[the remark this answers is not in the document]
Buckram123: Thanks, will look into it!

framework/contracts/native/ibc-client/src/queries.rs
```diff
@@ -1,60 +1,92 @@
...
-    host_chain: String,
-    account_id: AccountId,
-) -> StdResult<LatestQueryResponse> {
-    let channel = CHANNELS.load(deps.storage, &host_chain)?;
-    LATEST_QUERIES.load(deps.storage, (&channel, account_id))
+    start: Option<(AccountId, String)>,
+    limit: Option<u32>,
+) -> IbcClientResult<ListAccountsResponse> {
+    let start = start
+        .map(|s| {
```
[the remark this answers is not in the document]
Buckram123: > That would require an indexed map in our current designs and those are a bit of a mess to deal with. (Though possible)

Can't you just make a type where `AccountId` would be a key Prefix and AccountTrace of this AccountId would be a SubPrefix?
I think it avoids using indexed map

framework/contracts/native/ibc-host/tests/integration.rs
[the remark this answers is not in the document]
Kayanski: Ah, this is only a unit test here, and this should code should never be used. This is used to mock the presence of an IBC connection.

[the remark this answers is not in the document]
Kayanski: And yeah it's registered at the ibc-client level

framework/contracts/native/ibc-client/src/queries.rs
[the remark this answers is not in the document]
Kayanski: yes sure, that's a little different on the storage. What use case do you see on having this query available ?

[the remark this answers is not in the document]
Buckram123: > yes sure, that's a little different on the storage. What use case do you see on having this query available ?

No idea right now, xD! Let's say we have an app that wants to batch ibc txs to all abstract accounts

framework/contracts/native/ibc-host/tests/integration.rs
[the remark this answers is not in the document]
Buckram123: Got it, thanks

framework/packages/abstract-core/src/native/ibc_host.rs
```diff
@@ -0,0 +1,167 @@
...
+        account_id: AccountId,
+        msgs: Vec<CosmosMsg>,
+    },
+    /// Allows for remote execution from the Polytone implementation
+    #[cfg_attr(feature = "interface", fn_name("ibc_execute"))]
+    Execute {
+        account_id: AccountId,
+        /// The address of the calling account id. This is used purely for the send-all-back method.
+        /// We include it in all messages one-the-less to simpify the users life
+        proxy_address: String,
```
CyberHoward, line 130: If this is only used for send_all_back, can we store it in the host somehow?

Kayanski: No, the idea is that accountIds don't change, so we can denote accounts using those. However, proxy addresses may change. So we can't store that information and trust it

framework/contracts/native/ibc-client/src/ibc.rs
```diff
@@ -1,526 +1,102 @@
...
+            // We need to get the address of the remote proxy from the account creation response
+            if let Callback::Execute(Ok(response)) = &callback.result {
+                let account_creation_result = response.result[0].clone();
+
+                let wasm_event = account_creation_result
+                    .events
+                    .into_iter()
+                    .find(|e| e.ty == "wasm")
+                    .ok_or(IbcClientError::IbcFailed(callback.clone()))?;
+                let remote_proxy_address = wasm_event
```
CyberHoward, line 65: Hmmmmm this seems a bit vulnerable but I don't see a good way to do it another way.

framework/packages/abstract-core/src/native/ibc_host.rs
CyberHoward: Because of the possibility of chain resets? I.e. uni-5 -> uni-6 ?

framework/packages/abstract-core/src/native/ibc_host.rs
```diff
@@ -0,0 +1,167 @@
...
+        account_factory_address: Option<String>,
+        version_control_address: Option<String>,
+    },
+    /// Register the Polytone proxy for a specific chain.
+    /// proxy should be a local address (will be validated)
+    RegisterChainProxy { chain: String, proxy: String },
+    /// Remove the Polytone proxy for a specific chain.
+    RemoveChainProxy { chain: String },
+    /// Create an account internally (used for account creation before account action)
+    InternalRegisterAccount {
```
CyberHoward, line 114: Can we move this into a `Callback` variant like with the other contracts?

framework/contracts/native/ibc-host/src/endpoints/packet.rs
```diff
@@ -0,0 +1,115 @@
...
+    info: MessageInfo,
+    proxy_address: String,
+    account_id: AccountId,
+    host_action: HostAction,
+) -> HostResult {
+    // We verify the caller is indeed registered for the calling chain
+    let client_chain = REVERSE_CHAIN_PROXYS.load(deps.storage, &info.sender)?;
+
+    // We execute the action
+    _handle_host_action(
```
CyberHoward, line 31: can we merge this into one function? seems overkill to split them.

framework/contracts/native/ibc-host/src/endpoints/packet.rs
```diff
@@ -0,0 +1,115 @@
...
+                        action,
+                        client_proxy_address: proxy_address,
+                        account_id: remote_account_id,
+                        chain_name: client_chain,
+                    },
+                )?;
+
+                // We add a submessage after account creation to dispatch the action
+                let sub_msg =
+                    SubMsg::reply_on_success(create_account_message, INIT_BEFORE_ACTION_REPLY_ID);
```
CyberHoward, line 108: What do you think about moving this reply logic into the `receive_register(with_reply: bool, ...)` function?

That would remove the need to call this message internally and should save some gas.

framework/packages/abstract-core/src/native/ibc_host.rs
```diff
@@ -0,0 +1,167 @@
...
+#[cosmwasm_schema::cw_serde]
+#[derive(QueryResponses)]
+#[cfg_attr(feature = "interface", derive(cw_orch::QueryFns))]
+pub enum QueryMsg {
+    /// Returns [`ConfigResponse`].
+    #[returns(ConfigResponse)]
+    Config {},
+    #[returns(RegisteredChainsResponse)]
+    RegisteredChains {
+        start: Option<String>,
```
CyberHoward, line 145:
```suggestion
        start_after: Option<String>,
```

framework/contracts/native/ibc-host/tests/integration.rs
CyberHoward, line 1: We're going to need wayyy more tests

framework/packages/abstract-core/src/native/account_factory.rs
```diff
@@ -111,6 +114,8 @@ pub enum ExecuteMsg {
         description: Option<String>,
         // Account link
         link: Option<String>,
+        /// Account id on the remote chain. Will create a new id (by incrementing), if not specified
+        account_id: Option<AccountId>,
```
CyberHoward, line 118: Correct me if I'm wrong but if we migrate the current version of this contract then sub-account creation from old factory contracts should still work because it won't provide this field which is interpreted as None.

framework/packages/abstract-core/src/native/ibc.rs
```diff
@@ -0,0 +1,60 @@
...
+
+// CallbackInfo from modules, that is turned into an IbcResponseMsg by the ibc client
+#[cosmwasm_schema::cw_serde]
+pub struct CallbackInfo {
+    /// Used to identify the callback that is sent (acts like the reply ID)
+    pub id: String,
+    /// Used to add information to the callback.
+    /// This is usually used to provide information to the ibc callback function for context
+    /// Use `EMPTY_BINARY` if you don't need this field
+    pub msg: Binary,
```
CyberHoward, line 13: Can this be Option?

framework/packages/abstract-core/src/native/ibc_host.rs
Kayanski: Yes or for whatever reason the proxy address might change. It doesn't seem reasonable to put a constraint on the proxy address when the account already has much constraints on it

framework/contracts/native/ibc-host/src/endpoints/execute.rs
```diff
@@ -0,0 +1,138 @@
...
+            info,
+            ans_host_address,
+            version_control_address,
+            account_factory_address,
+        ),
+        ExecuteMsg::RegisterChainProxy { chain, proxy } => {
+            register_chain_proxy(deps, info, chain, proxy)
+        }
+        ExecuteMsg::RemoveChainProxy { chain } => remove_chain_proxy(deps, info, chain),
+        ExecuteMsg::RecoverAccount {
```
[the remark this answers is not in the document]
Kayanski: On channel closure, we can reopen a channel and be rid of polytone.
The way everything is coded doesn't really depend on the reliability of the ibc channel. 
So we can remove that

framework/packages/abstract-core/src/native/account_factory.rs
Kayanski: Yes, that works !

framework/packages/abstract-core/src/native/ibc_host.rs
CyberHoward: I don't think we'll ever want to enable a re-mapping of account-id to manager/proxy.

framework/packages/abstract-core/src/native/ibc_host.rs
CyberHoward: Or remove this alltogether by moving the reply into the account creation call (like mentioned in another comment)

framework/packages/abstract-core/src/objects/chain_name.rs
```diff
@@ -32,6 +37,20 @@ impl ChainName {
         Ok(chain_name)
     }
 
+    /// Gets the chain name from a raw ans-entry.
+    pub fn from_ibc_entry(entry: String) -> AbstractResult<Self> {
```
CyberHoward, line 41: This doesn't seem to be used anywhere? What's the goal of this function?

Kayanski: It's used in the modules folder. It's used to get "juno" from "juno>wyndex", to know what app to route to

framework/packages/abstract-core/src/base.rs
```diff
@@ -11,6 +12,9 @@ pub enum ExecuteMsg<BaseMsg, CustomExecMsg, ReceiveMsg = Empty> {
     /// An app request defined by a base consumer.
     Module(CustomExecMsg),
     /// IbcReceive to process IBC callbacks
+    /// In order to trust this, they should verify this comes from the a note contract.
```
CyberHoward, line 15: Oh so the callbacks come from the note and not the ibc client?

You can implement that check here: https://github.com/AbstractSDK/abstract/blob/aaf44ec8e0a284479908f19c89e5a18638759b39/framework/packages/abstract-app/src/endpoints/ibc_callback.rs

Kayanski: Yes, that is a mistake in the comment, the callbacks come from the ibc client (because the ibc-client registers them to itself and redispatches them)

framework/packages/abstract-core/src/native/ibc_host.rs
Kayanski: Ok, so we store the proxy address in the ibc host ?

framework/packages/abstract-core/src/objects/chain_name.rs
CyberHoward: Let's remove it. I don't like that it accepts a String type. We should have some overarching type for this (`Trace`?)

Because assets also have it... and account ids as well...

`TraceString`?

CyberHoward (approved):

PR #40, opened 2023-08-15
Staking pair with no bonding period

Buckram123:
Adding no bonding staking pair to `wyndex-bundle`:
- Added abstr token
- Added abstr-usd pair and pool
- Added one small test of it

Closes https://github.com/AbstractSDK/integration-bundles/issues/27

Note: period is 0 instead of None, don't think we want to change this method https://github.com/AbstractSDK/abstract/blob/main/integrations/wyndex/packages/wyndex-adapter/src/staking.rs#L244-L253

# Checklist

- CI is green.
- Changelog updated.

### DIFF
--- a/modules/contracts/adapters/cw-staking/tests/stake.rs
+++ b/modules/contracts/adapters/cw-staking/tests/stake.rs
@@ -19,6 +19,7 @@ use cosmwasm_std::{coin, Addr, Empty, Uint128};
 use cw_asset::AssetInfoBase;
 use cw_orch::prelude::*;
 use speculoos::*;
+use wyndex_bundle::ABSTR_USD_LP;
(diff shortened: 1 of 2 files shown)


Buckram123:
@MbBrainz Does that work for you?

MbBrainz:
@Buckram123 thanks for this effort!

There is just one issue im having and that is that the stakingInfo for that pool still returns a unbonding period being `Some(Time(0))`.  In our contract, we assume that a zero unbonding period is a none, which imo makes the most sense. I just adjusted one of our ignored tests to use the abstr_usd pool using this branch as a dependency, however It then returned a the `Some(...)` instead of `None`.

The staking ingo response is defined to have an optional `unbonding_periods` variable which is what we base instantiate logic upon.

https://github.com/AbstractSDK/abstract/blob/82ced8a6a1fbe9ac7aadb2753f2c2541e5dfedbe/framework/packages/staking/src/msg.rs#L134-L139

I guess we can adjust the contract a bit so that it handles the Time(0) case.

PR #41, opened 2023-08-15
Redo fee-grant api

adairrr:
# Checklist

- CI is green.
- Changelog updated.

### DIFF
--- a/framework/packages/abstract-sdk/src/apis/grant.rs
+++ b/framework/packages/abstract-sdk/src/apis/grant.rs
@@ -1,223 +1,185 @@
-//! # Grant
-//! Interacts with the feegrant module of cosmos
-//!
+//! # Fee Granter
+//! This module provides functionality to interact with the feegrant module of Cosmos.
+//! It allows for granting fee expenditure rights to other accounts.
 
 use std::time::Duration;
 
-use crate::features::AccountIdentification;
-
-use cosmos_sdk_proto::{cosmos::base, cosmos::feegrant, traits::Message, Any};
-use cosmwasm_std::{to_binary, Addr, Coin, CosmosMsg, Timestamp};
+use cosmos_sdk_proto::{Any, cosmos::base, cosmos::feegrant, traits::Message};
+use cosmwasm_std::{Addr, Binary, Coin, CosmosMsg, Timestamp};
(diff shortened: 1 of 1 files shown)


framework/packages/abstract-sdk/src/apis/feegrant.rs
```diff
@@ -0,0 +1,348 @@
...
+        Any {
+            type_url: feegrant::v1beta1::AllowedMsgAllowance::TYPE_URL.to_string(),
+            value: feegrant::v1beta1::AllowedMsgAllowance {
+                allowance: self.allowance.as_ref().map(|a| a.to_any()),
+                allowed_messages: self.allowed_messages.clone(),
+            }
+            .encode_to_vec(),
+        }
+    }
+}
```
adairrr, line 348: We should probably use `Into<Any>`

CyberHoward:
> Osmosis test tube does not support contracts fee-granting. This sucks and we need to find a way around it. However, you can test out the app yourself here: https://console.abstract.money/junotestnet/account/0/modules/abstract:gas-station?activeTab=interactions

Is it just test-tube or Osmosis as a chain?
@adairrr

framework/packages/abstract-sdk/src/apis/feegrant.rs
```diff
@@ -0,0 +1,348 @@
...
+///
+/// let grant: FeeGranter  = module.fee_granter();
+/// ```
+/// */
+#[derive(Clone)]
+pub struct FeeGranter<'a, T: GrantInterface> {
+    base: &'a T,
+    deps: cosmwasm_std::Deps<'a>,
+    // We use a RefCell here to allow for the granter to be set after the FeeGranter is created
+    granter: RefCell<Option<Addr>>,
```
CyberHoward, line 64: This should just be owned type. The struct can be `mut` if the user wants to set the granter.

I.e. change `fn granter(&self, ...)` to `fn granter(&mut self, ...)`

framework/packages/abstract-sdk/src/apis/feegrant.rs
Buckram123: Unfortunately it's used in `dyn` trait object and `Into` is marked with `Sized`, which makes it not object-safe

PR #44, opened 2023-08-21
Fix/standalone modules

Buckram123:
Standalone modules now only check that version matches in case cw2 implemented or no checks otherwise
# Checklist

- CI is green.
- Changelog updated.


framework/contracts/account/manager/Cargo.toml
```diff
@@ -55,6 +55,8 @@ abstract-app = { workspace = true, features = ["test-utils"] }
 abstract-testing = { workspace = true }
 abstract-macros = { workspace = true }
 
+cw20-base = { version = "1.1.0" }
```
CyberHoward, line 58: Can we use the workspace for this dependency?

Edit: I see that you depend on this version for the test. I think it would be better maintainability wise to have a macro like the `mock_app!` and `mock_adapter!` that lets us define standalone contracts easily and test with them without having to depend on an external lib (and its version). 

What do you think?

CyberHoward (changes requested):
Small comment on standalone related testing.

framework/contracts/account/manager/Cargo.toml
Buckram123: > I think it would be better maintainability wise to have a macro like the `mock_app!` and `mock_adapter!` that lets us define standalone contracts easily and test with them without having to depend on an external lib (and its version).
> 
> What do you think?

Yeah we can add that as well, but that sounds like a rigged test to me.  Don't we want to make sure that we can install `external standalone contract`?
We can add `mock_app/mock_adapter` standalone tests in addition to this one.

Buckram123: It felt like macros for standalone modules would be overkill, since we only care about cw2

framework/contracts/account/manager/tests/common/mock_modules.rs
```diff
@@ -128,3 +128,58 @@ pub mod app_1 {
...
+        deps: ::cosmwasm_std::Deps,
+        env: ::cosmwasm_std::Env,
+        msg: MockMsg,
+    ) -> cosmwasm_std::StdResult<cosmwasm_std::Binary> {
+        Ok(cosmwasm_std::Binary::default())
+    }
+}
+
+// standalone 2 does not have cw2
+pub mod standalone_2 {
```
CyberHoward, line 171: Maybe the name can indicate that it doesn't have a cw2 instead of using a number?

framework/contracts/account/manager/Cargo.toml
CyberHoward: Sure, this should work for now :)

framework/contracts/account/manager/tests/common/mock_modules.rs
Buckram123: Sure, just didn't want them to stand out in this file. :eyes:
adapter_1, adapter_2, app_1, app_2, standalone_cw2, standalone_no_cw2

CyberHoward (approved):

PR #45, opened 2023-08-22
Fix builds and bump cw-orch

CyberHoward:
Fixes the CI and bumps osmosis-std to 0.17.0 and cw-orch to 0.15.0.

# Checklist

- [ ] CI is green.
- [ ] Changelog updated.

PR #46, opened 2023-08-22
Add ability to install modules on account creation

Buckram123:
After fighting for a few hours with `cw-ownable`. I decided to add `forced` field on `ManagerExecuteMsg::SetOwner{}` method, LMK if you have a better idea here.
# Checklist

- CI is green.
- Changelog updated.


framework/contracts/account/manager/src/commands.rs
```diff
@@ -224,6 +225,10 @@ pub fn create_subaccount(
             name,
             description,
             link,
+            install_modules,
+            // TODO?
+            base_asset: None,
```
CyberHoward, line 230: This is fixed in https://github.com/AbstractSDK/abstract/pull/45

framework/contracts/account/manager/src/commands.rs
```diff
@@ -297,17 +305,22 @@ pub fn set_owner(
...
-        &info.sender,
-        cw_ownable::Action::TransferOwnership {
-            new_owner: new_owner_addr.into_string(),
-            expiry: None,
-        },
-    )?;
-
+    let ownership = if forced {
+        // Force move ownership
+        // Note: this will not require claim
```
CyberHoward, line 310: And this is required because the factory sets the owner after sub-account creation?

framework/contracts/account/manager/src/commands.rs
Buckram123: Yep will fix it after it gets merget into main

framework/contracts/account/manager/src/commands.rs
Buckram123: TL;DR: 
1. We can't do it in manager intantiate because account doesn't have a proxy
2. We can't make account factory owner first and move ownership after that, because it will require claim from owner

CyberHoward: Yeah for case 2 I think we should come up with a different solution. We **know** that on instantiation the factory is the caller so maybe we can have a `CACHE_ADMIN` or whatever that uses the regular Admin type and is set and used as long as it is set. On its first admin change the cache is cleared and the contracts moves over to the cw-ownable admin. wdyt?

Buckram123: Ok, if that's your preference!

Buckram123: I like that approach, do you think we check for cached admin first or after cw_ownable didn't match? Thinking to create `remove_cache_admin` method for this case instead of transfer admin after instantiation, thoughts?

CyberHoward: > @CyberHoward Is there a noticeable downside for adding forced mode to move ownership, that I'm missing?

It's kind of ugly in my opinion (because it changes the contract's API) as it's only a requirement for one specify scenario that can only take place on instantiation.


> I like that approach, do you think we check for cached admin first or after cw_ownable didn't match? Thinking to create `remove_cache_admin` method for this case instead of transfer admin after instantiation, thoughts?

You mean adding a new endpoint to the execute message? I think we can get by without changing the API of the contract.

Buckram123: I don't get how to achieve what you propose without changing the API, when do we clear cache admin?

Buckram123: Ok, I'm just going to allow account_factory to install modules, it already has power to update module addresses, don't think it creates new vector of attacks

framework/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased] - yyyy-mm-dd
 
 ### Added
+- Install modules on account or Sub-account creation
+- Forced field for moving ownership of manager without claiming
```
CyberHoward, line 12: This forced field is no longer needed.

CyberHoward (approved):
LGTM!

PR #47, opened 2023-08-22
ABS-166: Create accountability module

0xmovses:
## Description
See the original ticket [ABS-166](https://linear.app/abstract-sdk/issue/ABS-166/create-accountability-module). 

The main purpose of this was to get familiar with the SDK and create a new module end-to-end. I tried to keep the implementation as simple as possible for this reason. Whilst creating the module several things came up as possible enhancements that may want to be added before merging. Also I took some notes about some areas where we could have nicer doc comments and api improvements. 

This is work to lay the foundation for a forthcoming Tutorial section in the docs, which would guide a developer through writing a module end-to-end.

## Changelog
- Created a new module under `/apps/challenge`
- Created a new cosmwasm smart contract, handlers, error and msg. 
- Created two test files `wasm.rs` and `integration.rs`. 

## Testing
- I tested deploying the wasm bin contract
- There are 11 new integration tests for the contract. You can run them with `just test`

## Enhancements / Considerations
- After PR review I'll generate the schemas.
- In general if this was intended to be used on mainnet I would want to write more elaborate tests, coming up with various edge cases. 
- Incentivisation Mechanic: Where if the admin complete's the challenge, they get some token (fungible or non)
- Croncat functionality
  - There are several ways we may want to add functionality for this. One way is to use Croncat to periodically call `count_votes`
  - On a call with Adair yesterday we added the type :
 ```rust
#[cosmwasm_schema::cw_serde]
pub enum Penalty {
    FixedAmount {
        asset: OfferAsset,
    },
    Daily {
        asset: OfferAsset,
        split_between_friends: bool,
    },
}
```
I'm not sure exactly what the functionality should be for `Daily`. Is it that the total collateral is divided among the total days for the challenge and distributed equally to all friends every time there is a false vote? If yes I will need to add a field for `ChallengeEntry` that keeps track of total days and add more tests for this. 

- Proof of Challenge. 
  - Another idea we had was to add some functionality that allows the admin to prove they completed their challenge. For example uploading a picture of them bench pressing ;) thusly becoming the greatest NFT ever.

## CI
I've incorporated clippy changes but I am seeing some other tests failing:
```
test osmosis_test::stake_lp ... FAILED
test osmosis_test::claim_all ... FAILED
test osmosis_test::staking_inited ... FAILED
test osmosis_test::unstake_lp ... FAILED
```
Any ideas why this might be?

### CODE REVIEW

File: modules/contracts/apps/challenge/Cargo.toml [hunk 1]
```diff
@@ -41,6 +41,8 @@ abstract-core = { workspace = true }
 abstract-app = { workspace = true }
 abstract-sdk = { workspace = true }
 
+chrono = { workspace = true }
```
Buckram123, line 44: Can you compile and check if wasm deployable on cosmos. Might be wrong, but I remember I've had issues with this crate after `wasmbind` feature got enabled by default. Cosmwasm has limited list of accepted methods

modules/Cargo.toml
```diff
@@ -51,6 +51,7 @@ thiserror = { version = "1.0" }
 protobuf = { version = "2", features = ["with-bytes"] }
 
 clap = { version = "4.0.32", features = ["derive"] }
+chrono = "0.4"
```
adairrr, line 54: Is this package able to be wasmed?

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    // validate the address begins with 0x
+    // @TODO - this should be a more robust validation, for easy
+    // testing we are just checking for 0x
+    if !friend_address.starts_with("0x") {
+        return Err(AppError::Std(StdError::generic_err(
+            "Friend address must begin with 0x",
+        )));
+    }
```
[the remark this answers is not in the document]
0xmovses: nice. That's done

modules/Cargo.toml
CyberHoward, line 54: I think this could potentially be a big dependency to add. Do you have a rough idea on how much it impacts wasm size? @Buckram123 I remember you saying something about this as well.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+    info: MessageInfo,
+    app: ChallengeApp,
+    challenge: ChallengeEntry,
+) -> AppResult {
+    // Only the admin should be able to create a challenge.
+    app.admin.assert_admin(deps.as_ref(), &info.sender)?;
+
+    // Generate the challenge id
+    let id = NEXT_ID.update(deps.storage, |id| AppResult::Ok(id + 1))?;
+    let challenge_id = format!("challenge_{id}");
```
CyberHoward, line 80: What's the reason for formatting this into a fixed string? I'd either just stick to a number (takes less storage) or let the user customize the ID as a string on creation (which could be nice!)

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    app: ChallengeApp,
+    challenge_id: String,
+    challenge: ChallengeEntry,
+) -> AppResult {
+    app.admin.assert_admin(deps.as_ref(), &info.sender)?;
+
+    // will return an error if the challenge doesn't exist
```
CyberHoward, line 101: The errors returned by CosmWasm are notorious for being uninformative so often it's better to parse them into something more human-readable.

For updating values in storage we often opt for an `Option` based struct. I.e. a `ChallengeEntryUpdate` struct of the like that has all its field be `Option`. That way when updating a value you don't have to re-affirm the values you dont want to change.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+    info: MessageInfo,
+    app: ChallengeApp,
+    challenge_id: String,
+) -> AppResult {
+    app.admin.assert_admin(deps.as_ref(), &info.sender)?;
+
+    CHALLENGE_LIST.load(deps.storage, challenge_id.clone())?;
+    CHALLENGE_LIST.remove(deps.storage, challenge_id.clone());
+
+    Ok(Response::new().add_attribute("action", "cancel_challenge"))
```
CyberHoward, line 122: Could be nice to have this return the ID as well in the attributes.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+        .any(|f| &f.address == friend_address)
+    {
+        return Err(AppError::Std(StdError::generic_err(
+            "Friend already added for this challenge",
+        )));
+    }
+
+    let friend_address = deps.api.addr_validate(friend_address)?;
+
+    let friend = Friend {
```
CyberHoward, line 150: Every 3-letter agency will be all over this xd

modules/contracts/apps/challenge/src/msg.rs
```diff
@@ -0,0 +1,114 @@
...
+        challenge: ChallengeEntry,
+    },
+    UpdateChallenge {
+        challenge_id: String,
+        challenge: ChallengeEntry,
+    },
+    CancelChallenge {
+        challenge_id: String,
+    },
+    AddFriendForChallenge {
```
CyberHoward, line 55: This variant is redundant IMO as one can call `AddFriendsForChallenge` just the same. Same as `RemoveFriendForChallenge`. I'd prefer to see a `RemoveFriendsForChallenge` variant or you could merge those two and have an `UpdateFriendsForChallenge` which takes in friends to add and friends to remove. In general minimizing the API to the functional minimum is the approach we've been trying to stick to.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+    env: Env,
+    info: MessageInfo,
+    _app: &ChallengeApp,
+    vote: Option<bool>,
+    challenge_id: &str,
+) -> AppResult {
+    let today = date_from_block(&env);
+
+    // If Admin checked in today, friends can't vote
+    if DAILY_CHECK_INS
```
CyberHoward, line 267: This could get weird for different timezones...

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+    info: MessageInfo,
+    _env: Env,
+    app: &ChallengeApp,
+    challenge_id: String,
+) -> AppResult {
+    let votes_for_challenge = VOTES
+        .load(deps.storage, challenge_id.clone())
+        .unwrap_or_else(|_| Vec::new());
+
+    let any_false_vote = votes_for_challenge.iter().any(|v| v.vote == Some(false));
```
CyberHoward, line 311: This could get abused by his friends, maybe some treshold would be appropriate? And/or the ability for the admin to veto a count. In the end it's the admin's good will so if he really deems his opinion thruthworthy he should be able to overwrite his friend's descision. The ability for him to do so itself will push his friend to collaborate.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+fn charge_penalty(
+    deps: DepsMut,
+    info: MessageInfo,
+    app: &ChallengeApp,
+    challenge_id: String,
+) -> AppResult {
+    let challenge = CHALLENGE_LIST.load(deps.storage, challenge_id.clone())?;
+    let friends = CHALLENGE_FRIENDS.load(deps.storage, challenge_id.clone())?;
+
+    let admin_address = deps.api.addr_validate(info.sender.as_ref())?;
```
CyberHoward, line 328: Why would we assume the caller is the admin?

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+            let transfer_actions: Result<Vec<_>, _> = friends
+                .into_iter()
+                .map(|friend| bank.transfer(vec![asset_per_friend.clone()], &friend.address))
+                .collect();
+
+            let transfer_msgs = executor.execute(transfer_actions?);
+
+            Ok(Response::new()
+                .add_messages(transfer_msgs)
+                .add_attribute("action", "charge_fixed_amount_penalty"))
```
CyberHoward, line 358: The total amount charged would be useful to see here as well

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+        }
+        Penalty::Daily {
+            asset,
+            split_between_friends: _,
+        } => {
+            // Not sure what the exact implementation should be here.
```
CyberHoward, line 364: I'm not sure either... An optional end-date for the challenge would be nice to have tho.

modules/contracts/apps/challenge/src/handlers/query.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::state::{CHALLENGE_FRIENDS, CHALLENGE_LIST, DAILY_CHECK_INS, VOTES};
+use cosmwasm_std::{to_binary, Binary, Deps, Env};
+
+pub fn query_handler(
+    deps: Deps,
+    _env: Env,
+    app: &ChallengeApp,
+    msg: ChallengeQueryMsg,
+) -> AppResult<Binary> {
+    match msg {
```
CyberHoward, line 14: For automation purposes we could add a query here that returns if the challenge failed and a pay-out is required. A cron-cat job could be triggered on that query result.

modules/contracts/apps/challenge/src/state.rs
```diff
@@ -0,0 +1,54 @@
...
+    },
+    Daily {
+        asset: OfferAsset,
+        split_between_friends: bool,
+    },
+}
+
+#[cosmwasm_schema::cw_serde]
+pub struct Friend {
+    pub address: Addr,
```
CyberHoward, line 30: Using `Friend` in the public API is dangerous as the `Addr` type assumes the data is already validated. 

Consider changing the field to a generic T: AddressLike and adding a `check` method to `Friend<String>` that returns `Friend<Addr>`.

CyberHoward (changes requested):
Awesome and clear code! Thanks for the effort. 

I left a bunch of comments just to be thorough. Some of them are more important then others, especially if we don't move this to mainnet (tho I think we should). 

Looking forward to the feedback items you have on the docs + api!

modules/Cargo.toml
Buckram123: Yeah, it's pretty big with default features. I've had some issue with it, since it was exposing non default cosmwasm methods, when `default-features=true`
Decided to just check sizes (with optimizer) this contract, results:
- 502K with chrono
- 362K with chrono (default-features = false)
- 359K without chrono

So I think just disabling default features would be good enough!

modules/contracts/apps/challenge/examples/deploy.rs
```diff
@@ -0,0 +1,28 @@
...
+    tokio::runtime::Runtime,
+};
+
+use abstract_challenge_app::{contract::CHALLENGE_APP_ID, ChallengeApp};
+use abstract_interface::AppDeployer;
+use semver::Version;
+
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+fn main() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
Buckram123: Since this contract has the "abstract" namespace don't think we want to create account and claim namespace here. Unless for debugging that it deploys(and will have to change namespace before executing)

modules/contracts/apps/challenge/src/handlers/execute.rs
Buckram123: > The errors returned by CosmWasm are notorious for being uninformative so often it's better to parse them into something more human-readable.

I guess the usual approach would be to use `.may_load().ok_or()?`. Because we want to have different error in case storage is screwed somehow

modules/contracts/apps/challenge/src/handlers/execute.rs
Buckram123: Is there any reason to load it by the way? `remove()` won't do anything if there was no value by this key

Buckram123: > Could be nice to have this return the ID as well in the attributes.

Hmm, I would disagree with that, since `challenge_id` will already be logged in tx, no need to repeat it

CyberHoward: I guess, not sure if indexers can easily make use of those values tho.

modules/contracts/apps/challenge/tests/integrations.rs
```diff
@@ -0,0 +1,461 @@
...
+use abstract_interface::{Abstract, AbstractAccount, AppDeployer, *};
+use cosmwasm_std::{coin, Uint128};
+use cw_asset::AssetInfo;
+use cw_orch::{anyhow, deploy::Deploy, prelude::*};
+use lazy_static::lazy_static;
+
+// consts for testing
+const ADMIN: &str = "admin";
+const DENOM: &str = "TOKEN";
+lazy_static! {
```
Buckram123, line 30: Lazy static in tests :exploding_head:

Buckram123 (changes requested):
Nice progress! It's gonna be so cool app to onboard people with this app into both abstract and cosmos!

modules/contracts/apps/challenge/metadata.json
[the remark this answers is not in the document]
Buckram123: > Yes! see https://github.com/AbstractSDK/abstract/blob/main/app-template/metadata.json

Where can I get list of icons?

modules/contracts/apps/challenge/src/handlers/execute.rs
0xmovses: @Buckram123 using `ok_or()?` was causing a compiler error because the returned error type does not directly convert to `AppError`, so I used `map_err()?` instead.

modules/contracts/apps/challenge/metadata.json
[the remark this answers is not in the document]
CyberHoward: > 

@adairrr

modules/contracts/apps/challenge/src/handlers/execute.rs
0xmovses: I went with generated`u64` id, less storage is better and less error prone with potential malicious user input, possible dups and validation.

modules/contracts/apps/challenge/src/msg.rs
```diff
@@ -0,0 +1,114 @@
...
+impl Frequency {
+    pub fn to_interval(self) -> CronCatInterval {
+        match self {
+            Frequency::EveryNBlocks(blocks) => CronCatInterval::Block(blocks),
+            Frequency::Daily => unimplemented!(),
+            Frequency::Weekly => unimplemented!(),
+            Frequency::Monthly => unimplemented!(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
0xmovses: Made a ticket to track this [here](https://linear.app/abstract-sdk/issue/ABS-187/implement-frequency-onto-croncat)!

modules/contracts/apps/challenge/src/msg.rs
```diff
@@ -0,0 +1,114 @@
...
+    },
+}
+
+#[cosmwasm_schema::cw_serde]
+#[cfg_attr(feature = "interface", derive(cw_orch::QueryFns))]
+#[cfg_attr(feature = "interface", impl_into(QueryMsg))]
+#[derive(QueryResponses)]
+pub enum ChallengeQueryMsg {
+    #[returns(ChallengeResponse)]
+    Challenge { challenge_id: String },
```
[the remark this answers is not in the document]
0xmovses: Defo will be useful, will add.

modules/contracts/apps/challenge/src/msg.rs
```diff
@@ -0,0 +1,114 @@
...
+
+#[cosmwasm_schema::cw_serde]
+pub struct CheckInResponse {
+    pub check_in: Option<CheckIn>,
+}
+
+#[cosmwasm_schema::cw_serde]
+pub struct VotesResponse {
+    pub votes: Option<Vec<Vote>>,
+}
```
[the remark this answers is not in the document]
0xmovses: I think empty Vector we don't need Option types here.

modules/contracts/apps/challenge/src/state.rs
```diff
@@ -0,0 +1,54 @@
+use abstract_dex_adapter::msg::OfferAsset;
+use cosmwasm_std::Addr;
+use cw_storage_plus::{Item, Map};
+
+#[cosmwasm_schema::cw_serde]
+pub struct Config {
+    pub native_denom: String,
```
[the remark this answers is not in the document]
0xmovses: Yes I think this a nice addition to support cw20.

modules/contracts/apps/challenge/src/state.rs
```diff
@@ -0,0 +1,54 @@
...
+    Daily {
+        asset: OfferAsset,
+        split_between_friends: bool,
+    },
+}
+
+#[cosmwasm_schema::cw_serde]
+pub struct Friend {
+    pub address: Addr,
+    pub name: String,
```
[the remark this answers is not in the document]
0xmovses: haha nothing immutable abuse.

modules/contracts/apps/challenge/src/state.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct Vote {
+    pub voter: String,
+    pub vote: Option<bool>,
+    pub challenge_id: String,
+}
+
+#[cosmwasm_schema::cw_serde]
+pub struct CheckIn {
+    pub last_checked_in: Option<String>,
+    pub next_check_in_by: u64, //block number
```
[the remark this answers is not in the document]
0xmovses: Went with just block u64, got write of String as time in general.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+        .may_load(deps.storage, challenge_id.clone())?
+        .unwrap_or_default();
+
+    if friends_for_challenge
+        .iter()
+        .any(|f| &f.address == friend_address)
+    {
+        return Err(AppError::Std(StdError::generic_err(
+            "Friend already added for this challenge",
+        )));
```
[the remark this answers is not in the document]
0xmovses: Gotya, i'll add this as a variant to the error enum.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    // Check if Admin has already checked in today
+    if DAILY_CHECK_INS
+        .load(deps.storage, today.clone())
+        .map_or(false, |check_in| {
+            check_in.last_checked_in.as_deref() == Some(&today)
+        })
+    {
+        return Err(AppError::AlreadyCheckedIn {});
+    }
```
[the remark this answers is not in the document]
0xmovses: Will update.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    let _blocks_per_day = 1440; // dummy value, check this
+    let next_check_in_block = env.block.height + 10;
```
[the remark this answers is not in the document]
0xmovses: removed

modules/contracts/apps/challenge/src/msg.rs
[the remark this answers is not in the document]
Buckram123: This one is always hard to choose for me, lol

modules/contracts/apps/challenge/src/state.rs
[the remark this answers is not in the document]
CyberHoward: Then you should have a look at using [`cw-asset`](https://crates.io/crates/cw-asset)

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,385 @@
...
+                    "No friends found for the challenge.",
+                )));
+            }
+
+            // Calculate each friend's share
+            let amount_per_friend = asset.amount.u128() / num_friends;
+            let asset_per_friend = OfferAsset {
+                name: asset.name,
+                amount: Uint128::from(amount_per_friend),
+            };
```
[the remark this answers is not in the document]
0xmovses: The rounding is automatically done towards zero since its integer division. If there is a remainder, it is discarded.

modules/contracts/apps/challenge/src/handlers/query.rs
0xmovses: In the current implementation, the pay-out happens automatically when `count_vote` is called, the votes are counted and if there are any false votes, `charge_penalty is called. Currently `count_vote` has to be manually called. 

One way to use croncat to automate this could be to have a `ChallengeStatus` enum for a `status` field on `ChallengeEntry`. 

Then we can add an `end_block` on `ChallengeEntry`, if we are `>=` than the `end_block` when `daily_check_in` is called, the `ChallengeEntry.status` is set to `Finished`. Then we can use the query to kick off a croncat job which calls `count_vote`

modules/contracts/apps/challenge/src/handlers/execute.rs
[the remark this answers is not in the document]
Buckram123: Yeah, but let's imagine we have challenge, where the challenge creator should be charged for 3 coins, and he only has 2 friends. Who should get 3th coin?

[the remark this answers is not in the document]
0xmovses: Just tried to convert to `f64` and divide so that each friend would get `1.5` but `AnsAsset` wont allow this. We could randomly distribute the remainder to one of the friends, lucky prize.

modules/contracts/apps/challenge/src/handlers/query.rs
CyberHoward: As long as there's no way for the owner to veto a vote IMO there's probably no need for cron-cat here because you can just call `count_vote` when one person votes against. I think one of the design choices that is most important to this kind of app is the timeline design. 

- When do votes start/end?
- When does the tallying occur? (if only possible after a specific time-stamp then cron-cat can help here)
- Is there a veto period? 

From an app-design perspective I think the day-centric way of administering the challenges is an approach that doesn't go well with on-chain logic.  Instead of working with a 24h refresh period (which needs to be in sync with user's timezone) we could work with a "ticketing" system where a vote either goes to the last ticket created (if within a customizable delay) or a new ticket is created (with a set end-date for voting and a veto-period for administration) 
That would also allow for shorter or longer duration challenges as well as automation for cleanup by cron-cat after the ticket's end-time.

modules/contracts/apps/challenge/src/handlers/execute.rs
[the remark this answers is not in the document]
Buckram123: Or burn it :rofl:

modules/contracts/apps/challenge/src/handlers/query.rs
0xmovses: I'm going to go with simple blockheight for time within the contract first, `end_block` is set on creation of challenges and can be updated after creation. I'l get these changes up with some new tests. If anything this just allows me easier iteration with the tests until the design and flow is set. 

We can change this time modelling to a timestamp `String` later.

I have an idea for when votes start/end, which is based on the new challenge.status. Also in my new changes, counting of votes, or tallying, occurs only after `challenge.status = ChalengeStatus::OverAndUncounted`. 

Once these changes are up I think it will become clearer to us the most suitable path to go.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,475 @@
...
+    _env: Env,
+    app: &ChallengeApp,
+    challenge_id: u64,
+) -> AppResult {
+    let votes = VOTES.may_load(
+        deps.storage,
+        (challenge_id.to_owned(), info.sender.to_owned()),
+    )?;
+
+    let any_false_vote = votes.iter().any(|v| v.approval == Some(false));
```
0xmovses, line 379: @Buckram123 By the way the tally happens here, we simply check if there is a false vote present. If there is then it's a failed challenge. This was the simplest possible strategy. 

Other tally strategies could be chosen and set in the `ChallengeEntry`, then the associated tally function called.

CyberHoward: Yup! For that strategy the tally function could be called directly after, saves a tx.

modules/contracts/apps/challenge/src/handlers/execute.rs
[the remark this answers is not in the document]
0xmovses: So I originally opted for random distribute. But then didn't want to bring in `rand` as it would bloat the contract. Attempted to write something in raw rust, but it would be computationally expensive and not *that* random. 

So instead went with this, the remainder will just stay in the admin's account. And added a 
```rust
.add_attribute("remainder is", reaminder.to_string()))
```
In the response

modules/contracts/apps/challenge/src/handlers/query.rs
0xmovses: With my latest updates, these are the answers to your questions above:

- When do votes start/end?
  - votes start after `challenge_entry.end_block` has passed. 
  - votes end after `tally_votes` is called (this is only callable after `challenge_entry.end_block` has passed. `tally_votes` will set the `challenge_entry.status` to either `OverAndFailed` (friends won) or `OverAndCompleted` (admin won)
  
  - When does the tallying occur? (if only possible after a specific time-stamp then cron-cat can help here)
    - When `tally_votes` is called, it will tally only if the `challenge_status == ChallengeStatus::OverAndPending`. 
    
 - Is there a veto period?
   - In the current implementation, No.

adairrr:
> Any ideas why this might be?
Could be that the osmosis dex cw-staking wasm is out of date or missing. Happens more often than not.

modules/contracts/apps/challenge/examples/deploy.rs
```diff
@@ -0,0 +1,28 @@
...
+    dotenv().ok();
+    env_logger::init();
+    let chain = parse_network("uni-6");
```
adairrr, line 16: Hm we should update the template to take in the network as the argument for this script.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,568 @@
...
+    deps: DepsMut,
+    info: MessageInfo,
+    app: &ChallengeApp,
+    challenge_id: u64,
+) -> AppResult {
+    app.admin.assert_admin(deps.as_ref(), &info.sender)?;
+    let mut challenge = CHALLENGE_LIST.load(deps.storage, challenge_id.clone())?;
+    if challenge.status != ChallengeStatus::Active {
+        return Err(AppError::WrongChallengeStatus {});
+    }
```
adairrr, line 149: Would we want to be able to cancel a challenge if it is in the over status?

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,568 @@
...
+    CHALLENGE_VOTES, DAILY_CHECK_INS, NEXT_ID, VOTES,
+};
+
+pub fn execute_handler(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    app: ChallengeApp,
+    msg: ChallengeExecuteMsg,
+) -> AppResult {
```
adairrr, line 22: Since it seems every api is admin controlled, we *could* do the admin assertion here, though it wouldn't change semantics.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,568 @@
...
+            ChallengeStatus::Active => {
+                challenge.status = ChallengeStatus::OverAndPending;
+                CHALLENGE_LIST.save(deps.storage, challenge_id.clone(), &challenge)?;
+                return Ok(Response::new()
+                    .add_attribute("action", "check_in")
+                    .add_attribute(
+                        "message",
+                        "Challenge has ended. You can no longer register a daily check in. 
+                         ChallengeStatus has been set to OverAndPending.",
+                    ));
```
adairrr: We should use the app.tag_response here so that the wasm-abstract event is included.

adairrr: Same comment below

adairrr: We can also create the response at the top of the method, mutating it throughout, then return it at the very bottom.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,568 @@
...
+
+        // The admin has missed the deadline for checking in, they are given a strike.
+        // The contract manually sets the next check in time.
+        now if now >= check_in.next_check_in_by => {
+            for strike in challenge.admin_strikes.iter_mut() {
+                if !*strike {
+                    *strike = true;
+                    break;
+                }
+            }
```
adairrr, line 300: So we change every strike to be true?

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,568 @@
...
+                tally_result: None,
+            };
+            check_ins.push(check_in);
+            DAILY_CHECK_INS.save(deps.storage, challenge_id, &check_ins)?;
+
+            return Ok(Response::new()
+                .add_attribute("action", "check_in")
+                .add_attribute(
+                    "message",
+                    "You missed the deadline for checking in. You have been given a strike.",
```
adairrr, line 322: Nice. This is not a common pattern and will be great for the frontend.

modules/contracts/apps/challenge/src/handlers/execute.rs
adairrr: With the exception of the friends methods, so nevermind.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,568 @@
...
+
+fn cast_vote(
+    deps: DepsMut,
+    env: Env,
+    _info: MessageInfo,
+    _app: &ChallengeApp,
+    vote: Vote<String>,
+    challenge_id: u64,
+) -> AppResult {
+    let mut vote = vote.check(deps.as_ref())?.optimisitc();
```
adairrr, line 356: nit: optimistic is spelled incorrectly

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,568 @@
...
+
+    // If all friends have voted, we tally the votes.
+    if CHALLENGE_VOTES
+        .load(deps.storage, challenge_id.clone())?
+        .len()
+        == CHALLENGE_FRIENDS.load(deps.storage, challenge_id)?.len()
+    {
+        tally_votes_for_check_in(deps, env, challenge_id)?;
+    }
+    Ok(Response::new().add_attribute("action", "cast_vote"))
```
adairrr, line 398: We should be using tag_response here

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,568 @@
...
+    //@TODO only update ChallengeStatus if the challenge is OverAndPending
+    // update check_in status
+    // update check_in tally_result
+    //
+    // Then charge penalty and there check if the check_in tally_result is Some(false)
+    // if yes divide the collateral among how many days there are between now
+    // and challenge.end and split that amount between friends.
+
+    if any_false_vote {
+        challenge.status = ChallengeStatus::OverAndFailed;
```
adairrr: So the challenge is failed if one day's check-in is failed?

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,568 @@
...
+    }
+}
+
+fn charge_penalty(
+    deps: DepsMut,
+    info: MessageInfo,
+    app: &ChallengeApp,
+    challenge_id: u64,
+) -> AppResult {
+    deps.api.addr_validate(info.sender.as_ref())?;
```
adairrr: `info.sender` is already an `Addr` and validated.

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,568 @@
...
+}
+
+fn veto_vote(
+    deps: DepsMut,
+    info: MessageInfo,
+    _env: Env,
+    challenge_id: u64,
+    vote: Vote<String>,
+) -> AppResult {
+    if info.sender.to_string() != ADMIN.load(deps.storage)? {
```
adairrr, line 493: Do we have 2 admins? The primary and this secondary one?

modules/contracts/apps/challenge/src/handlers/execute.rs
0xmovses: whoops, will fix.

modules/contracts/apps/challenge/src/handlers/execute.rs
0xmovses: this wasn't fully implemented when you reviewed, it's been updated now!

modules/contracts/apps/challenge/src/handlers/execute.rs
0xmovses: This was old code, the `ADMIN` state is gone now.

modules/contracts/apps/challenge/src/handlers/execute.rs
0xmovses: I think no, we want the history to persist. Challenge's should only be cancelable if they are active right?

Buckram123 (changes requested):
AWESOME UPDATES! Just few small nits

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,496 @@
...
+        }
+    }
+
+    // validate the String addresses and convert them to Addr
+    // before saving
+    let friends: Vec<Friend<Addr>> = friends
+        .iter()
+        .cloned()
+        .map(|friend| friend.check(deps.as_ref()).unwrap())
+        .collect();
```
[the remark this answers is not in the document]
0xmovses: think it was left over from some previous thing. Danke!

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,496 @@
...
+        ))));
+    }
+
+    // Check if the voter has already voted
+    if VOTES
+        .may_load(deps.storage, (check_in.last.nanos(), vote.voter.to_owned()))
+        .map_or(false, |votes| votes.iter().any(|v| v.voter == vote.voter))
+    {
+        return Err(AppError::AlreadyVoted {});
+    }
```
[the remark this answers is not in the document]
0xmovses: Updated the state to
```rust
/// Key is a tuple of (challenge_id, check_in.last_checked_in, voter_address).
/// By using a composite key, it ensures only one user can vote per check_in.
pub const VOTES: Map<(u64, u64, Addr), Vote<Addr>> = Map::new("votes");
```

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,496 @@
...
+
+        // The admin has missed the deadline for checking in, they are given a strike.
+        // The contract manually sets the next check in time.
+        now if now >= check_in.next => {
+            for strike in challenge.admin_strikes.iter_mut() {
+                if !(*strike) {
+                    *strike = true;
+                    break;
+                }
+            }
```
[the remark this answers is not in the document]
0xmovses: Done we now have 
```rust
#[cosmwasm_schema::cw_serde]
pub struct ChallengeEntry {
    pub name: String,
    pub collateral: OfferAsset,
    pub description: String,
    pub end: EndType,
    pub total_check_ins: u128,
    pub status: ChallengeStatus,
    pub admin_strikes: StrikeConfig,
}

#[cosmwasm_schema::cw_serde]
pub struct StrikeConfig {
    /// The number of striked the admin has incurred.
    pub num_strikes: u8,
    /// When num_strikes reached the limit, the challenge will be cancelled.
    pub limit: u8,
}

impl Default for StrikeConfig {
    fn default() -> Self {
        StrikeConfig {
            num_strikes: 3,
            limit: 3,
        }
    }
}
```

modules/contracts/apps/challenge/examples/deploy.rs
0xmovses: Made a ticket to track that here: https://linear.app/abstract-sdk/issue/ABS-192/update-template-to-in-network-as-arg

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,497 @@
...
+    }
+    if let Some(collateral) = new_challenge.collateral {
+        loaded_challenge.collateral = collateral;
+    }
+    if let Some(description) = new_challenge.description {
+        loaded_challenge.description = description;
+    }
+    if let Some(end) = new_challenge.end {
+        loaded_challenge.end = end;
+    }
```
Buckram123, line 118: Is it okay to update collateral and end for the challenge?

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,497 @@
...
+            return Err(AppError::AlreadyAdded {});
+        }
+    }
+
+    // validate the String addresses and convert them to Addr
+    // before saving
+    let friends: Vec<Friend<Addr>> = friends
+        .into_iter()
+        .map(|friend| friend.check(deps.as_ref()).unwrap())
+        .collect();
```
Buckram123, line 199: Let's remove this unwrap and collect Result here

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,497 @@
...
+                        "Challenge has ended. You can no longer register a daily check in.",
+                    ),
+                    "daily check in",
+                ));
+            }
+            _ => {
+                return Err(AppError::Std(StdError::generic_err(format!(
+                    "Challenge has ended. Challenge end_timestamp is {:?} current timestamp is {}",
+                    challenge.end.seconds(),
+                    env.block.time.seconds()
```
Buckram123, line 273: One of them is debug output second isn't :eyes:

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -0,0 +1,497 @@
...
+                ))));
+            }
+        }
+    }
+
+    let mut check_ins = DAILY_CHECK_INS.load(deps.storage, challenge_id)?;
+    let check_in = check_ins.last().unwrap();
+    let next = Timestamp::from_seconds(env.block.time.seconds() + DAY);
```
Buckram123, line 281: Nice, looks much cleaner `+ DAY`

Buckram123 (changes requested):
LFG, that will be last nits, I promise!

modules/contracts/apps/challenge/src/handlers/execute.rs
0xmovses: whoops, nice catch.

modules/contracts/apps/challenge/src/handlers/execute.rs
0xmovses: Removed collateral and end as yes, they could be abused.

CyberHoward:
@0xmovses chrono needs to be <= v0.4.27. See: 
https://github.com/osmosis-labs/osmosis-rust/pull/87

Buckram123 (approved):
LFG

PR #52, opened 2023-08-25
Fix abs 164

CyberHoward:
Update the account creation logic to:
- Save a module queue in the manager for later installation. 
- Update the message execution ordering in the account factory to unsure permissions are changed at the correct time.
- Update the internal config logic of the manager to only allow account factory calls when the account is being instantiated. After instantiation the module queue is loaded and any queued modules are installed. 

Question @Buckram123: How do we want to manage payments related to these module installations?

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/contracts/account/manager/src/commands.rs
+++ b/framework/contracts/account/manager/src/commands.rs
@@ -40,7 +40,6 @@ use cosmwasm_std::{
     DepsMut, Empty, Env, MessageInfo, Response, StdError, StdResult, Storage, WasmMsg,
 };
 use cw2::{get_contract_version, ContractVersion};
-use cw_controllers::AdminError;
 use cw_ownable::Ownership;
 use cw_storage_plus::Item;
 use semver::Version;
--- a/framework/contracts/account/manager/src/commands.rs
+++ b/framework/contracts/account/manager/src/commands.rs
@@ -1,6 +1,6 @@
 use crate::{contract::ManagerResult, error::ManagerError, queries::query_module_cw2};
 use crate::{validation, versioning};
-use abstract_core::manager::state::ACCOUNT_FACTORY;
+use abstract_core::manager::state::{ACCOUNT_FACTORY, MODULE_QUEUE};
 
 use abstract_core::adapter::{
     AuthorizedAddressesResponse, BaseExecuteMsg, BaseQueryMsg, ExecuteMsg as AdapterExecMsg,
@@ -40,6 +40,7 @@ use cosmwasm_std::{
     DepsMut, Empty, Env, MessageInfo, Response, StdError, StdResult, Storage, WasmMsg,
 };
 use cw2::{get_contract_version, ContractVersion};
+use cw_controllers::AdminError;
 use cw_ownable::Ownership;
 use cw_storage_plus::Item;
 use semver::Version;
@@ -83,7 +84,7 @@ pub fn update_module_addresses(
     Ok(ManagerResponse::action("update_module_addresses"))
 }
 
-// Attempts to install a new module through the Module Factory Contract
+/// Attempts to install a new module through the Module Factory Contract
 pub fn install_module(
     deps: DepsMut,
     msg_info: MessageInfo,
@@ -109,9 +110,9 @@ pub fn install_module(
     Ok(response)
 }
 
-// Generate message for installing module
+/// Generate message for installing module
 pub(crate) fn install_module_internal(
-    storage: &mut dyn Storage,
+    storage: &dyn Storage,
     module: ModuleInfo,
     module_factory_address: Addr,
     init_msg: Option<Binary>,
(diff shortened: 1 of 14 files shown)


CyberHoward:
No idea why the formatting fails...

Buckram123:
> No idea why the formatting fails...

1.72 rustfmt :rocket:

Buckram123:
> Question @Buckram123: How do we want to manage payments related to these module installations?

Idea 1: 
- in the install_module array add `Option<Coin>`, so user can specify how much he's okay to attach for each module
- check that sum of those coins matches to the attached
- check in VC that it's enough to cover
- rest goes to the account balance
Idea 2:
- Check that attached coins is enough to cover installation for every module
- rest goes to the account balance
  
From a user perspective: 
- first idea:  pain
- second idea: some crazy app can have $69420 cost, and user wanted to create account with $69420 funds, but now has 0 but scam app installed

In my opinion it's bad to preinstall apps with monetization, wdyt? @CyberHoward

CyberHoward:
I think we should allow it as in nr. 2. Yes it could be bad but we can mitigate some of these issues by exposing a query on the module or account factory where you can get the total cost of installing a set of modules. That way the FE can easily provide the user with the cost of creating the application. 

I think that re-designing the module factory api to allow for multiple modules to be installed could also make use of that functionality. The flow would then be: 

- User queries account or module factory for the cost of creating the modules
- User provides x amout of funds, larger or equal to the total cost of installing the modules. 
- Factory performs the same query and sends the funds required for module payment to the manager. Forwards the rest to the proxy. 
- Manager forwards all the funds received in this last call from the factory with its call to the module factory to install the modules. 
- The module factory checks that the received funds add up to the required total and distributes the payment as it installs and instantiates the modules.

Wdyt? Hardest part will be the multi-module install on the module factory I think.

framework/packages/abstract-core/src/core/manager.rs
```diff
@@ -86,6 +87,8 @@ pub mod state {
     /// Stores the dependency relationship between modules
     /// map module -> modules that depend on module.
     pub const DEPENDENTS: Map<ModuleId, HashSet<String>> = Map::new("dependents");
+    /// Stores a queue of modules to install on the account after creation.
+    pub const MODULE_QUEUE: Item<Vec<(ModuleInfo, Option<Binary>)>> = Item::new("mqueue");
```
Buckram123, line 91: :tada:

framework/packages/abstract-core/src/core/manager.rs
```diff
@@ -114,6 +117,8 @@ pub struct InstantiateMsg {
     pub name: String,
     pub description: Option<String>,
     pub link: Option<String>,
+    // Optionally modules can be provided. They will be installed after account registration.
+    pub install_modules: Vec<(ModuleInfo, Option<Binary>)>,
```
Buckram123, line 121:
```suggestion
    // Optionally modules can be provided. They will be installed during account registration.
    pub install_modules: Vec<(ModuleInfo, Option<Binary>)>,
```
Just noting that if any of app installation fails whole account creation fails as well

Buckram123 (approved):
Thanks for saving me again :heart:

Buckram123:
> I think we should allow it as in nr. 2. Yes it could be bad but we can mitigate some of these issues by exposing a query on the module or account factory where you can get the total cost of installing a set of modules. That way the FE can easily provide the user with the cost of creating the application.
> 
> I think that re-designing the module factory api to allow for multiple modules to be installed could also make use of that functionality. The flow would then be:
> 
>     * User queries account or module factory for the cost of creating the modules
> 
>     * User provides x amout of funds, larger or equal to the total cost of installing the modules.
> 
>     * Factory performs the same query and sends the funds required for module payment to the manager. Forwards the rest to the proxy.
> 
>     * Manager forwards all the funds received in this last call from the factory with its call to the module factory to install the modules.
> 
>     * The module factory checks that the received funds add up to the required total and distributes the payment as it installs and instantiates the modules.
> 
> 
> Wdyt? Hardest part will be the multi-module install on the module factory I think.

Yeah, if we support multi-install everywhere, not only on account creation that will be great and would make sense to add assisting query as well :)

Are you ok to move that to another issue? You can assign me on that as well!

CyberHoward:
Sure! We'll get this merged first.

framework/contracts/account/manager/src/commands.rs
```diff
@@ -774,21 +787,63 @@ pub fn update_account_status(
...
+                    init_msg,
+                    funds,
+                )
+            })
+            .collect::<ManagerResult<Vec<CosmosMsg>>>();
+
+        // clear the queue
+        MODULE_QUEUE.remove(deps.storage);
+
+        Ok(ManagerResponse::action("manager_after_init").add_messages(install_msgs?))
```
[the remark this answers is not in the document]
CyberHoward: Yes, will do!

framework/contracts/native/account-factory/src/commands.rs
```diff
@@ -29,8 +29,8 @@ use crate::{
     response::MsgInstantiateContractResponse, state::*,
 };
 
-pub const CREATE_ACCOUNT_MANAGER_MSG_ID: u64 = 1u64;
-pub const CREATE_ACCOUNT_PROXY_MSG_ID: u64 = 2u64;
+pub const CREATE_ACCOUNT_PROXY_MSG_ID: u64 = 1u64;
+pub const CREATE_ACCOUNT_MANAGER_MSG_ID: u64 = 2u64;
```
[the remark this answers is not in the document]
CyberHoward: No, I don't think we need to. This works fine!

framework/packages/abstract-core/src/core/manager.rs
CyberHoward: Yup and that's ok!

PR #54, opened 2023-08-25
Fix sub account admin assertion

Buckram123:
This change allows modules of the owning account to execute and interact with the sub-accounts.
### Checklist

- [x] CI is green.
- [x] Changelog updated.

### DIFF
--- a/framework/packages/abstract-core/src/objects/gov_type.rs
+++ b/framework/packages/abstract-core/src/objects/gov_type.rs
@@ -21,6 +21,8 @@ pub enum GovernanceDetails<T: AddressLike> {
     },
     /// Used when the account is a sub-account of another account.
     SubAccount {
+        /// The manager of the account of which this account is the sub-account.
+        manager: T,
         /// The proxy of the account of which this account is the sub-account.
         proxy: T,
     },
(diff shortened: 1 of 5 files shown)


framework/packages/abstract-core/src/objects/gov_type.rs
```diff
@@ -21,8 +21,8 @@ pub enum GovernanceDetails<T: AddressLike> {
     },
     /// Used when the account is a sub-account of another account.
     SubAccount {
-        /// The manager of the account of which this account is the sub-account.
-        manager: T,
+        /// The proxy of the account of which this account is the sub-account.
+        proxy: T,
```
CyberHoward, line 25: Instead of this only storing the proxy here, maybe you can store the `Core` struct that contains both proxy and manager addresses? Would save a query.

Buckram123: How would you imagine implementation of `Addresslike` for the `Core`?

CyberHoward: Currently it's defined as 

```rust
/// Contains the minimal Abstract Account contract addresses.
#[cosmwasm_schema::cw_serde]
pub struct AccountBase {
    pub manager: Addr,
    pub proxy: Addr,
}
```

We can make the manager and proxy `AddressLike` as suggested. We can also move the type to the `abstract-core::core::mod.rs` instead of it being in the `version_control.rs`

PR #55, opened 2023-08-28
Register the sub account on the main account on creation

Buckram123:
Added sub-accounts to manager's state.

 Last question is who should be able to call the `set_owner` method? Right now only direct owner of this manager can do it(proxy of this manager for sub-accounts in latest implementation). 
 
Do we want this to still be the case or top-level owner should be able to set_owner as well?

### Checklist

- [x] CI is green.
- [x] Changelog updated.


Buckram123:
@CyberHoward Can you think of any flow where users can put corrupt `manager_addr` or `proxy_addr`?
Was trying to break it, but couldn't

Buckram123:
Ah, it's technically not a callback of this contract, but the callback of sub-account, not sure if we just let it sit in the `ExecuteMsg`

CyberHoward:
> Ah, it's technically not a callback of this manager contract, but the callback of sub-account, not sure if we just let it sit in the `ExecuteMsg`

If it's not a callback of the manager itself then it should just be an execute message variant.

framework/packages/abstract-core/src/core/manager.rs
```diff
@@ -180,14 +188,23 @@ pub enum ExecuteMsg {
         // Provide list of module to install after sub-account creation
         install_modules: Vec<(ModuleInfo, Option<Binary>)>,
     },
+    /// Unregister sub-account
```
CyberHoward, line 191: Small suggestion (idk if we want to do this) but instead of having all the sub-account actions on the top level we could create a new `enum SubAccountAction` and have a single variant here `SubAccount(SubAccountAction)`. 

It would screw over cw-orch's function generation tho :/

CyberHoward (changes requested):
I think we could get away with having no replies if we update the instantiation logic of the contract. Curious to hear your thoughts!

framework/contracts/account/manager/src/contract.rs
```diff
@@ -216,6 +228,33 @@ pub fn handle_callback(mut deps: DepsMut, env: Env, info: MessageInfo) -> Manage
     Ok(Response::new())
 }
 
+#[cfg_attr(feature = "export", cosmwasm_std::entry_point)]
+pub fn reply(deps: DepsMut, _env: Env, msg: cosmwasm_std::Reply) -> ManagerResult {
```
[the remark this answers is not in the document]
Buckram123: Only gas, I think!
You don't pay gas for loading contract from reply

framework/packages/abstract-core/src/core/manager.rs
Buckram123: Sure, I like it more! And by the way, it will get fixed in the next cw-orch version :rocket: 
https://github.com/AbstractSDK/cw-orchestrator/pull/196/commits/0cafc1c176b0adb7b249c9c93399ced90aac3a2a

framework/contracts/account/manager/src/contract.rs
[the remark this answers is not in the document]
Buckram123: Removed reply, as we suggested, to have more simplicity.

framework/packages/abstract-core/src/core/manager.rs
CyberHoward: Yeah but it will require us to construct the inner `SubAccountAction` type to do the call.

CyberHoward (approved):
Great stuff sir!!

framework/packages/abstract-core/src/core/manager.rs
```diff
@@ -145,6 +145,19 @@ pub enum InternalConfigAction {
     },
 }
 
+#[cosmwasm_schema::cw_serde]
+#[non_exhaustive]
+pub enum SubAccountAction {
```
CyberHoward, line 150: I thought we could also move the account creation here?

Buckram123: Maybe, but that's not the action of sub-account, it's action of this manager

PR #56, opened 2023-08-30
Update optimizer and add wasms to gitattributes

CyberHoward:
see title


.github/workflows/release-artifacts.yml
```diff
@@ -30,12 +29,12 @@ jobs:
           docker run --rm -v "$(pwd)":/code \
           --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
           --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
-          cosmwasm/workspace-optimizer:0.12.13
+          abstractmoney/workspace-optimizer:0.14
```
Buckram123, line 32: Does that mean it would build with 0.14.1 on CI if we publish it?

app-template/artifacts/checksums.txt
```diff
@@ -1 +1,2 @@
-c2ea249ee0244d88defe24e09de2ba63f829614cb8866e7d74c9daa2ede7f4d2  app.wasm
+239c1b05822d728e92751215ca4e0a5ea273a3cbc884eea389503663267ea846  app-aarch64.wasm
+1608ddd5db7d8c0015ac499e14bc7d8a181f837dbd9e43fa1d7eb2fe428d1f4b  template_app-aarch64.wasm
```
Buckram123, line 2: It seems like you renamed package `app` to `template-app`, want to push it? Don't think you should have two wasms in artifacts

Buckram123 (approved):
Finally! LET'S MERGE IT!

PR #57, opened 2023-08-31
Support multi module installation

Buckram123:
To reduce gas consumption this PR aims to install multiple modules in single execution of module-factory

- Added query method on module factory to check that. 
- Any unused funds for modules monetization during `create_account` will be transferred to the proxy

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-core/src/native/module_factory.rs
```diff
@@ -57,18 +59,25 @@ pub enum ExecuteMsg {
...
     },
 }
 
+// TODO: do we want to move to that to have more user-friendly api?
+// #[cosmwasm_schema::cw_serde]
+// pub struct InstallModule {
+//     pub module: ModuleInfo,
+//     pub init_msg: Option<Binary>
+// }
+
```
Buckram123, line 80: @CyberHoward what do you think? Or array of tuples is good enough?

CyberHoward: I'd like this yes, then we can easily add a `coins: Vec<Coin>` in there as well for funding the module on instantiation.

You could also make it non-exhaustive

framework/CHANGELOG.md
```diff
@@ -10,10 +10,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Install modules on account or Sub-account creation
 - Manager stores his sub-accounts and sub-accounts can register or unregister in case of ownership change
+- Query on module factory to see how much funds needs to be attached for installing modules
```
CyberHoward, line 13: This is sooo nice...

framework/contracts/account/manager/tests/upgrades.rs
```diff
@@ -650,6 +656,238 @@ fn create_sub_account_with_installed_module() -> AResult {
...
+                ),
+            ],
+        },
+        GovernanceDetails::Monarchy {
+            monarch: sender.to_string(),
+        },
+        // we attach 1 less coin1
+        Some(&[coin(9, "coin1"), coin(10, "coin2")]),
+    );
+    // Mock doesn't implement debug so we can't .unwrap_err, LOL
```
CyberHoward, line 878: Which mock? If there's a type that's missing the `#[derive(Debug)]` we should create an issue for it.

framework/contracts/native/module-factory/src/commands.rs
```diff
@@ -14,112 +16,153 @@ use abstract_sdk::{
...
-        abstract_core::objects::module::Monetization::None => {}
-        // The monetization must be known to the factory for a module to be installed
-        _ => return Err(ModuleFactoryError::ModuleNotInstallable {}),
-    };
+    let mut sum_of_monetization = Coins::default();
+
+    // install messages
+    let mut sub_messages = Vec::with_capacity(modules_responses.len());
+    // list of modules to register after instantiation
+    let mut installed_modules = VecDeque::with_capacity(modules_responses.len());
```
CyberHoward, line 58: I like it!

framework/contracts/native/module-factory/src/commands.rs
```diff
@@ -231,15 +269,19 @@ pub fn update_factory_binaries(
...
-        deps.storage,
-        &Context {
-            account_base: None,
-            module: None,
-        },
-    )
+fn update_context(deps: DepsMut, context: &Context) -> Result<(), StdError> {
+    // Update context for after init
+    if context.modules.is_empty() {
+        CONTEXT.save(
```
CyberHoward, line 275: Instead of saving an empty context we could also just clear it.

framework/contracts/native/module-factory/src/contract.rs
```diff
@@ -34,7 +43,7 @@ pub fn instantiate(
         deps.storage,
         &Context {
             account_base: None,
-            module: None,
+            modules: VecDeque::new(),
```
CyberHoward, line 46: Same here, we don't really have to save context before we do any execution.

framework/contracts/account/manager/tests/upgrades.rs
Buckram123: AbstractAccount<Mock>

framework/packages/abstract-core/src/native/module_factory.rs
```diff
@@ -95,7 +108,12 @@ pub struct ConfigResponse {
 #[cosmwasm_schema::cw_serde]
 pub struct ContextResponse {
     pub account_base: Option<AccountBase>,
-    pub module: Option<Module>,
+    pub modules: Vec<Module>,
+}
+
+#[cosmwasm_schema::cw_serde]
+pub struct SimulateInstallModulesResponse {
+    pub required_funds: Vec<Coin>,
```
CyberHoward, line 116: see my comment where this is returned

framework/packages/abstract-interface/src/account/manager.rs
```diff
@@ -61,6 +61,19 @@ impl<Chain: CwEnv> Manager<Chain> {
...
+    pub fn install_modules<TInitMsg: Serialize>(
+        &self,
+        modules: Vec<(ModuleInfo, Option<&TInitMsg>)>,
+        funds: Option<&[Coin]>,
+    ) -> Result<Chain::Response, crate::AbstractInterfaceError> {
+        let modules = modules
+            .into_iter()
+            .map(|(module, init_msg)| (module, init_msg.map(|msg| to_binary(msg).unwrap())))
+            .collect();
+        self.execute(&ExecuteMsg::InstallModules { modules }, funds)
```
CyberHoward, line 73: Maybe if Options = None it would query the factory for the cost and use that by default?

framework/packages/abstract-sdk/src/apis/version_registry.rs
```diff
@@ -74,22 +75,31 @@ impl<'a, T: ModuleRegistryInterface> ModuleRegistry<'a, T> {
 
     /// Smart query for a module
     pub fn query_module(&self, module_info: ModuleInfo) -> AbstractSdkResult<Module> {
-        Ok(self.query_all_module_config(module_info)?.module)
+        Ok(self
+            .query_all_module_config(vec![module_info])?
```
CyberHoward, line 79: This `all` is weird, could we come up with a better name (and fix the doc-comment for it?)

`query_module_configs`?

framework/contracts/native/module-factory/src/contract.rs
```diff
@@ -104,14 +116,37 @@ pub fn query_config(deps: Deps) -> StdResult<ConfigResponse> {
...
+        .query_all_module_config(modules)
+        .map_err(|e| cosmwasm_std::StdError::generic_err(e.to_string()))?;
+    let mut coins = Coins::default();
+    for module in module_responses {
+        if let Monetization::InstallFee(fee) = module.config.monetization {
+            coins.add(fee.fee())?;
+        }
+    }
+    let resp = SimulateInstallModulesResponse {
+        required_funds: coins.into_vec(),
```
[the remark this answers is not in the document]
Buckram123: > I think we should split this up in `total_required_funds` and `required_funds: Vec<(ModuleId, Vec<Coin>)>`. That way the front-end can easily show the breakdown of the costs!

Yeah, that's cool! So we can later have something like "total_required_funds, install_funds, minimal_post_install_funds"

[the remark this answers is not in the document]
Buckram123: Are you ok with `install_funds` instead of `required_funds` naming? @CyberHoward

framework/packages/abstract-interface/src/account/manager.rs
CyberHoward: maybe not lol

Buckram123: We can make something like
```rust
enum InstallModulesFunds {
  Manual(Vec<Coin>),
  Auto,
}
```
and default (not sure which tbh)

CyberHoward: maybe a separate method is easier

CyberHoward (approved):

PR #63, opened 2023-09-04
Add version control to abstract app in instantiation

Buckram123:
### Checklist

- [x] CI is green.
- [x] Changelog updated.


modules/contracts/adapters/cw-staking/src/handlers/query.rs
```diff
@@ -30,7 +30,14 @@ pub fn query_handler(
                 // the query can be executed on the local chain
                 let mut provider = resolver::resolve_local_provider(&local_provider_name)
                     .map_err(|e| StdError::generic_err(e.to_string()))?;
-                provider.fetch_data(deps, env, None, ans_host, abstract_registry, staking_token)?;
+                provider.fetch_data(
+                    deps,
+                    env,
+                    None,
+                    ans_host,
+                    &abstract_registry,
```
CyberHoward, line 38: Is there any implementation that actually uses the version control address provided here? (should probably rename this "registry" to version_control_contract).

Buckram123: Osmosis uses it to get proxy addr

PR #65, opened 2023-09-06
Allow funds to be transferred to module on instantiation

Buckram123:
See the title
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/contracts/native/version-control/src/contract.rs
```diff
@@ -89,6 +89,11 @@ pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg) -> V
             namespace,
             monetization,
         } => set_module_monetization(deps, info, module_name, namespace, monetization),
+        ExecuteMsg::SetModuleInstantiationFunds {
```
CyberHoward, line 92: I was thinking to merge the metadata and this method into a single endpoint to configure the module

Buckram123: > I was thinking to merge the metadata and this method into a single endpoint to configure the module

And monetization?

CyberHoward: Yeah if possible! And have it be a `non_exhaustive`

framework/packages/abstract-core/src/objects/validation/verifiers.rs
```diff
@@ -19,7 +19,7 @@ fn is_valid_url(link: &str) -> bool {
     link.starts_with("http://") || link.starts_with("https://") || link.starts_with("ipfs://")
 }
 
-pub fn validate_link(link: &Option<String>) -> Result<(), ValidationError> {
+pub fn validate_link(link: Option<&str>) -> Result<(), ValidationError> {
```
CyberHoward, line 22: Rust Pro

framework/packages/abstract-core/src/native/version_control.rs
```diff
@@ -286,14 +304,15 @@ impl ModuleConfiguration {
 
     pub fn from_storage(storage: &dyn Storage, module: &ModuleInfo) -> Self {
```
CyberHoward, line 305: These are at least 3 load operations for a single module. Can we combine their information into a single data store so we only have to load once?

framework/contracts/native/module-factory/src/commands.rs
```diff
@@ -121,7 +127,7 @@ pub fn execute_create_modules(
                         module_addr: addr.to_string(),
                         module: new_module,
                     },
-                    vec![],
+                    new_module_init_funds,
```
CyberHoward, line 130: We send funds with the `register_module` action? Imo init funds should only be possible on apps and standalone, not adapters (because they are already instantiated) thoughts?

Buckram123: Thank you! Fixed

framework/packages/abstract-core/src/native/version_control.rs
Buckram123: Yep, changed state to have distinct storage for "default" items for module config as well!

framework/contracts/native/version-control/src/commands.rs
```diff
@@ -228,63 +230,85 @@ pub fn set_module_monetization(
...
+            }
+
+            // Update monetization
+            if let Some(monetization) = monetization {
+                current_cfg.monetization = monetization;
+            }
+
+            // Update init funds
+            if let Some(init_funds) = instantiation_funds {
+                if matches!(
```
CyberHoward, line 289: sweet!

PR #69, opened 2023-09-11
Add ability to query version of no-cw2 modules

Buckram123:
To accomplish querying of no-cw2 modules infos we have two few options
1. Return `None` for contract version
2. Return `abstract-core` type `ModuleVersion` of installed module, in case cw2 query failed

This PR implements second option.

Left to discuss:
- Right now it will try to return `ModuleVersion` of any no-cw2 contract, we can make it returning `ModuleVersion` for every Standalone, since they will have non-abstract format for cw2 or no-cw2 at all  
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/contracts/account/manager/src/queries.rs
```diff
@@ -94,12 +100,34 @@ pub fn handle_sub_accounts_query(
...
+pub fn query_module_version(
+    deps: Deps,
+    module_addr: Addr,
+    version_control: &VersionControlContract,
+) -> StdResult<ContractVersion> {
+    let module_registry = version_control.module_registry(deps);
+
+    // Currently manager doesn't save ModuleReference,
+    // so we need to check for standalone first
+    let code_id = deps
```
CyberHoward, line 112: Can we change this logic to only perform the code-id + standalone query if the cw2 query fails OR if the query succeeds but the cw2 name doesn't adhere to our formatting standards? Should be more efficient

framework/contracts/account/manager/src/queries.rs
```diff
@@ -94,12 +100,34 @@ pub fn handle_sub_accounts_query(
...
+        .code_id;
+    let module_info = module_registry
+        .query_standalone_info(code_id)
+        .map_err(|e| StdError::generic_err(e.to_string()))?;
+
+    if let Some(standalone_info) = module_info {
+        Ok(ContractVersion::try_from(standalone_info)
+            .map_err(|e| StdError::generic_err(e.to_string()))?)
+    } else {
+        let req = QueryRequest::Wasm(WasmQuery::Raw {
```
CyberHoward, line 124: I think there's a helper for this in the cw2 crate!

framework/contracts/account/manager/src/queries.rs
Buckram123: Implemented, thanks

CyberHoward (approved):

PR #83, opened 2023-09-14
Add missing docs warning and adapter cleanup.

CyberHoward:
- Adds a `#![warn(missing_docs)]` to the `msg.rs` and adapter package `type.rs` modules. 
- Adds a `adapter_msg_types` similar to `app_msg_types` and applied it to the adapters. 
- Moved the adapter messages to a `msg.rs` file in the adapter's contract repo instead of a package. 
- Moved the adapter-related packages to a `standards` folder for consistency. We can add other standards there in the future.

### Checklist

- [x] CI is green.
- [x] Changelog updated.


Buckram123 (changes requested):
Looking good! Now that I thought about it more, I'm not sure about moving messages from package, it forces users to export the contract as dependency every time they want to build a message, making the package somewhat useless? 

Anyway, only small path updates on integrations missing, I guess!

CyberHoward:
> Looking good! Now that I thought about it more, I'm not sure about moving messages from package, it forces users to export the contract as dependency every time they want to build a message, making the package somewhat useless?
> 
> Anyway, only small path updates on integrations missing, I guess!

Yes true. The reason for it as that the top-level `Msg` types are defined with the macro, which can only implement traits for types that are within the same crate (orphan rule). 

So either all the messages will be defined in the package (including the top-level messages) or the messages are defined with the contract and the macro invocation happens there. 

I guess leaving them in the package is not that bad...

Buckram123 (approved):
LGTM

PR #86, opened 2023-09-19
Update/Must use on abstract sdk methods

Buckram123:
This adds warnings on unused `AccountAction` and unused `CosmosMsg` from the `Executor`

- Helpfully `add_message` and `add_messages` takes `impl Into<CosmosMsg<T>>`, so it shouldn't affect users.
- Unfortunately `add_submessage` takes raw `SubMsg<T>`.  

So we still have an issue: we can't implement message wrapper with unused warning for unused SubMsg, because it will require users to cast it using "into" every time

### Checklist

- [x] CI is green.
- [x] Changelog updated.

### DIFF
--- a/framework/packages/abstract-sdk/src/account_action.rs
+++ b/framework/packages/abstract-sdk/src/account_action.rs
@@ -5,6 +5,7 @@ use cosmwasm_std::CosmosMsg;
 ///
 /// If required you can create an AccountAction from a CosmosMsg using `AccountAction::from(msg)`.
 #[derive(Debug, PartialEq, Clone, Default)]
+#[must_use = "Pass AccountAction to the Executor"]
 pub struct AccountAction(Vec<CosmosMsg>);
 
 impl AccountAction {
--- a/framework/packages/abstract-sdk/src/account_action.rs
+++ b/framework/packages/abstract-sdk/src/account_action.rs
@@ -1,11 +1,11 @@
 use cosmwasm_std::CosmosMsg;
 
 /// Encapsulates an action on the account.
-/// When a method returns an AccountAction, this means this message needs to be dispatched to the account using the [`Execution`] api.
+/// When a method returns an AccountAction, this means this message needs to be dispatched to the account using the [`crate::Execution`] api.
 ///
 /// If required you can create an AccountAction from a CosmosMsg using `AccountAction::from(msg)`.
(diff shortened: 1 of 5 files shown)


Kayanski:
It's already something that the sdk provides helpers for simple message mechanisms. 
For messages with replies, I think we can assume that developers are more advanced and will not forget to use the message result !

Kayanski:
Shouldn't we have a single way of executing all those actions ? 
We could pass the DepositMsgs and AccountAction all to the executor and it would decide what to do with it ?

Buckram123:
> Shouldn't we have a single way of executing all those actions ? We could pass the DepositMsgs and AccountAction all to the executor and it would decide what to do with it ?  So that the sdk would always create something you have to pass to the executor and then to the `Response::add_messages`function ?

Right now executor executes actions only from the proxy, idk if we want to expand that. `DepositMsgs` sends funds from this contract to the proxy, so it's not an action from the proxy

framework/packages/abstract-sdk/src/account_action.rs
```diff
@@ -1,11 +1,11 @@
 use cosmwasm_std::CosmosMsg;
 
 /// Encapsulates an action on the account.
-/// When a method returns an AccountAction, this means this message needs to be dispatched to the account using the [`Execution`] api.
+/// When a method returns an AccountAction, this means this message needs to be dispatched to the account using the [`crate::Execution`] api.
```
CyberHoward, line 4:
```suggestion
/// When a method returns an AccountAction, this means this message needs to be dispatched to the account using the [`Execution`](crate::Execution) api.
```

CyberHoward:
> > Shouldn't we have a single way of executing all those actions ? We could pass the DepositMsgs and AccountAction all to the executor and it would decide what to do with it ?  So that the sdk would always create something you have to pass to the executor and then to the `Response::add_messages`function ?
> 
> Right now executor executes actions only from the proxy, idk if we want to expand that. `DepositMsgs` sends funds from this contract to the proxy, so it's not an action from the proxy

I think Nicolas brings up a good point. The executor API is kind of confusing but IMO we should stick with it for now. We can look into removing the need for Response construction entirely by having a call stack inside the module struct itself and executing that stack at the end of a contract's execution.

CyberHoward (approved):

PR #87, opened 2023-09-20
Update template to a new version

Buckram123:
### Checklist

- [x] CI is green.
- [x] Changelog updated.


app-template/example.env
```diff
@@ -6,4 +6,4 @@ LOCAL_MNEMONIC=""
 TEST_MNEMONIC=""
 MAIN_MNEMONIC=""
 
-STATE_FILE="state.json"
+STATE_FILE="./state.json"
```
CyberHoward, line 9:
```suggestion
# Location where state is stored, defaults to ~/cw-orchestrator/state.json (Optional)
# STATE_FILE="./state.json"
```

app-template/examples/test-local.rs
```diff
@@ -0,0 +1,65 @@
+//! Deploys Abstract and the App module to a local Junod instance. See how to spin up a local chain here: https://docs.junonetwork.io/developer-guides/junod-local-dev-setup
```
CyberHoward, line 1:
```suggestion
//! Deploys Abstract and the App module to a local Junod instance. See how to spin up a local chain here: https://docs.junonetwork.io/developer-guides/junod-local-dev-setup
//! You can also start a juno container by running `just juno-local`.
```

app-template/justfile
```diff
@@ -53,20 +67,12 @@ wasm:
...
     --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
-    ${image}:0.12.13
+    ${image}:0.14.0
 
 # Generate the schemas for the app contract
 schema:
   cargo schema
 
-# Generate the typescript client for the app contract
-ts-codegen: schema
```
CyberHoward, line 63: Why was this removed?

Buckram123: typescript does not exist

app-template/examples/test-local.rs
Buckram123: Thanks!

app-template/justfile
CyberHoward: Okey let's keep it just Rust for now.

CyberHoward (approved):

PR #92, opened 2023-09-27
Migrate subscription module

Buckram123:
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,351 @@
...
+    };
+    if let Some(mut active_sub) = maybe_subscriber {
+        // Subscriber is active, update balance
+        active_sub.expiration_timestamp = active_sub.expiration_timestamp.plus_days(paid_for_days);
+        SUBSCRIBERS.save(deps.storage, &subscriber_addr, &active_sub)?;
+    } else {
+        // Subscriber is (re)activating his subscription.
+        if asset.amount.u128() < required_payment.u128() {
+            return Err(SubscriptionError::InsufficientPayment(
+                required_payment.u128() as u64,
```
AndrewPochapsky, line 105: Is this a safe cast? I guess it doesn't matter too much as it is just used in an error message but still curious.

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,351 @@
...
+        SubscriptionExecuteMsg::RefreshTWA {} => {
+            INCOME_TWA.try_update_value(&env, deps.storage)?;
+            Ok(Response::new())
+        }
+    }
+}
+
+/// Called when either paying with a native token or through the receive_cw20 endpoint when paying
+/// with a CW20.
+pub fn try_pay(
```
AndrewPochapsky, line 62: Might just be me but I think this function could benefit from refactoring some of the inner logic into their own functions to improve readability.

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,351 @@
...
+    ))
+}
+
+pub fn unsubscribe(
+    deps: DepsMut,
+    env: Env,
+    app: SubscriptionApp,
+    unsubscribe_addrs: Vec<String>,
+) -> SubscriptionResult {
+    let unsubscribe_addrs: Vec<Addr> = unsubscribe_addrs
```
AndrewPochapsky, line 154: Should we have some sort of limit on the amount of addresses passed in given that we loop over them?

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,351 @@
...
+
+    // update income
+    INCOME_TWA.accumulate(
+        &env,
+        deps.storage,
+        Decimal::from_atomics(Uint128::from(subscription_state.active_subs), 0)?
+            * subscription_config.subscription_cost_per_week,
+    )?;
+
+    for addr in unsubscribe_addrs.iter() {
```
AndrewPochapsky, line 171: Could we instead use `into_iter` here? I think this would make each `addr.to_string` avoid cloning since the underlying type will be owned.

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,351 @@
...
+    app: SubscriptionApp,
+    unsubscribe_addrs: Vec<String>,
+) -> SubscriptionResult {
+    let unsubscribe_addrs: Vec<Addr> = unsubscribe_addrs
+        .iter()
+        .map(|human| deps.api.addr_validate(human))
+        .collect::<StdResult<_>>()?;
+    let mut subscription_state = SUBSCRIPTION_STATE.load(deps.storage)?;
+    let subscription_config = SUBSCRIPTION_CONFIG.load(deps.storage)?;
+    let mut unsubscribed_addrs: Vec<String> = vec![];
```
AndrewPochapsky, line 160: Hm, perhaps there is a better way to name the `unsubscribe_addrs` and `unsubscribed_addrs` variables, they are kind of confusing.

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    // update subscription count
+    SUBSCRIPTION_STATE.save(deps.storage, &subscription_state)?;
+
+    let mut response = app.tag_response(Response::new().add_messages(claim_msgs), "unsubscribe");
+
+    if let Some(hook) = subscription_config.unsubscription_hook_addr {
+        let msg = UnsubscribedHookMsg {
+            unsubscribed: unsubscribed_addrs.clone(),
```
AndrewPochapsky, line 215: Do we need to clone here? I might be missing something but I don't see it used after this line :)

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,351 @@
...
+          //     let contributors_addr = app.modules(deps.as_ref()).module_address(CONTRIBUTORS_ID)?;
+          //     let contributor_config =
+          //         contr_state::CONTRIBUTION_CONFIG.query(&deps.querier, contributors_addr.clone())?;
+          //     let contributor_state =
+          //         contr_state::CONTRIBUTION_STATE.query(&deps.querier, contributors_addr)?;
+
+          //     let amount = (contributor_state.emissions * contributor_config.emission_user_share)
+          //         / Uint128::from(subscription_state.active_subs);
+          //     Asset::new(token, amount * Uint128::from(1u64))
+          // }
```
AndrewPochapsky, line 270: If we aren't using the code we should probably remove it.

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,351 @@
...
+    };
+
+    if !asset.amount.is_zero() {
+        // Update only if there was claim
+        subscriber.last_emission_claim_timestamp = env.block.time;
+        // SUBSCRIBERS.save(deps.storage, &subscriber_addr, &subscriber)?;
```
AndrewPochapsky, line 276: Same as above, let's remove this code if it is not being used.

modules/contracts/apps/subscription/src/queries.rs
```diff
@@ -0,0 +1,5 @@
+// pub fn handle_value_query(deps: Deps, query: ValueQueryMsg) -> StdResult<Binary> {
+//     query_assets_from_mem(deps, ans_host_addr, '')
+//     if query.asset_info.equal(asset)
+
+// }
```
AndrewPochapsky, line 5: Can we remove this file?

AndrewPochapsky (changes requested):
Apologies for the unsolicited feedback but I couldn't resist taking a look! Some of the suggestions could very well be wrong so I apologize again if that is the case :)

modules/contracts/apps/subscription/src/queries.rs
Buckram123: I need to check my vision, haven't even seen this file

modules/contracts/apps/subscription/src/handlers/execute.rs
Buckram123: Did some basic refactor to this method! Do you think it's enough for now?

modules/contracts/apps/subscription/src/handlers/execute.rs
Buckram123: Yep that's unsafe

modules/contracts/apps/subscription/src/handlers/execute.rs
Buckram123: Agree! Custom error is much better than out of gas

modules/contracts/apps/subscription/src/handlers/execute.rs
Buckram123: Yeah, `addr.into_string` on owned address will avoid clone

modules/contracts/apps/subscription/src/handlers/execute.rs
Buckram123: `canceled_subs` for addrs that are indeed "canceled" is better?

modules/contracts/apps/subscription/src/handlers/execute.rs
Buckram123: Yep, it was vector of hookmsgs before, clippy didn't notice that, LMAO
@AndrewPochapsky >>>> Clippy

modules/contracts/apps/subscription/examples/schema.rs
```diff
@@ -0,0 +1,34 @@
+use abstract_subscription::contract::SubscriptionApp;
+use abstract_subscription::msg::{StateResponse, SubscriberResponse, SubscriptionFeeResponse};
+use abstract_subscription::state::SubscriptionConfig;
+use cosmwasm_schema::{export_schema, export_schema_with_title, remove_schemas, schema_for};
+use cw_asset::{AssetInfo, AssetInfoUnchecked};
+use std::env::current_dir;
+use std::fs::create_dir_all;
+
+fn main() {
+    let mut out_dir = current_dir().unwrap();
```
CyberHoward, line 10: Are these required like this?
I.e. do we need those extra state and config exports?

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,316 @@
...
+    }
+}
+
+/// Called when either paying with a native token or through the receive_cw20 endpoint when paying
+/// with a CW20.
+pub fn try_pay(
+    app: SubscriptionApp,
+    deps: DepsMut,
+    env: Env,
+    _msg_info: MessageInfo,
```
CyberHoward, line 68: remove this if unused

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,316 @@
...
+    ))
+}
+
+pub fn unsubscribe(
+    deps: DepsMut,
+    env: Env,
+    app: SubscriptionApp,
+    unsubscribe_addrs: Vec<String>,
+) -> SubscriptionResult {
+    if unsubscribe_addrs.len() > MAX_UNSUBS {
```
CyberHoward, line 134: can we make this like 10 or 15? 5 seems so little.

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -0,0 +1,316 @@
...
+            Asset::new(token, amount)
+        }
+    };
+
+    if !asset.amount.is_zero() {
+        // Update only if there was claim
+        subscriber.last_emission_claim_timestamp = env.block.time;
+
+        let send_msg = app.bank(deps).transfer(vec![asset], subscriber_addr)?;
+        Ok(Some(app.executor(deps).execute(vec![send_msg])?))
```
CyberHoward, line 248: this should return AccountAction so you can combine multiple of them into one CosmosMsg. Saves a lot of gas!

modules/contracts/apps/subscription/src/handlers/instantiate.rs
```diff
@@ -0,0 +1,34 @@
...
+    deps: DepsMut,
+    env: Env,
+    _info: MessageInfo,
+    _app: SubscriptionApp,
+    msg: SubscriptionInstantiateMsg,
+) -> SubscriptionResult {
+    let subscription_config: SubscriptionConfig = SubscriptionConfig {
+        payment_asset: msg.payment_asset.check(deps.api, None)?,
+        subscription_cost_per_week: msg.subscription_cost_per_week,
+        subscription_per_week_emissions: msg.subscription_per_week_emissions.check(deps.api)?,
```
CyberHoward, line 21: Is there a reason why we opted for a week? It seems very arbitrary. 

I remember making the cost/emissions Decimal so that the app could support short timespans. I.e. have the cost/emissions be per second or per minute, the cost would then be 0.000037 or something like that.

modules/contracts/apps/subscription/src/handlers/instantiate.rs
```diff
@@ -0,0 +1,34 @@
...
+        unsubscription_hook_addr: msg
+            .unsubscription_hook_addr
+            .map(|human| deps.api.addr_validate(&human))
+            .transpose()?,
+    };
+
+    let subscription_state: SubscriptionState = SubscriptionState { active_subs: 0 };
+    SUBSCRIPTION_CONFIG.save(deps.storage, &subscription_config)?;
+    SUBSCRIPTION_STATE.save(deps.storage, &subscription_state)?;
+    INCOME_TWA.instantiate(deps.storage, &env, None, msg.income_averaging_period.u64())?;
```
CyberHoward, line 31: We should check here that the averaging period is more than the timebasis for the subscription. So in the current case a week!

CyberHoward (changes requested):
Great! Just some question about the time range in which everything is denominated (currently weeks). 

Was wondering if this could be updated to something shorter (seconds or minutes)? 

Also we have to keep in mind that some chains (like archway) have 18(!) decimal places for their coins. So I think we should probably move to `Decimal256` and `UInt256` for the TWA.

modules/contracts/apps/subscription/src/handlers/instantiate.rs
Buckram123: Haha, just tried to move from blocks, I guess went a bit too far. Yeah, we can opt for seconds instead :)

AndrewPochapsky (approved):
Thanks for the changes :)

modules/contracts/apps/subscription/src/handlers/instantiate.rs
Buckram123: Updated to seconds, now it feels a less user/test friendly, but more flexible!

modules/contracts/apps/subscription/src/handlers/query.rs
```diff
@@ -26,7 +26,8 @@ pub fn query_handler(
         }
         SubscriptionQueryMsg::Fee {} => {
             let config = SUBSCRIPTION_CONFIG.load(deps.storage)?;
-            let minimal_cost = Uint128::from(DURATION_IN_WEEKS) * config.subscription_cost_per_week;
+            let minimal_cost =
```
CyberHoward, line 29: Here as well, do we need the minimal cost?

CyberHoward (changes requested):
I think the minimal payment requirement was added because of the contribution integration but I think we can remove it now.

modules/contracts/apps/subscription/src/handlers/execute.rs
```diff
@@ -81,14 +80,15 @@ pub fn try_pay(
     }
     // Minimum of one period worth to (re)-subscribe.
     // prevents un- and re-subscribing all the time.
-    let required_payment = Uint128::from(DURATION_IN_WEEKS) * config.subscription_cost_per_week;
-    let paid_for_weeks = asset
+    let required_payment = Uint128::from(DURATION_IN_WEEKS * WEEK_IN_SECONDS)
```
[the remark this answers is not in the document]
Buckram123: Ah, I thought "~monthly" subscription is a requirement for it!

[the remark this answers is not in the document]
Buckram123: But don't we still want some minimal period, knowing unsubscription will be managed by other service and they can subscribe for 1ns which will make their subscription "active" until unsubscribe happen...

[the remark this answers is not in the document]
CyberHoward: Ah yeah. Maybe we could use the `averaging` period of the TWA?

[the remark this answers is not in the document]
CyberHoward: `income_averaging_period`

CyberHoward (approved):

PR #88, opened 2023-09-20
Update optimizer

CyberHoward:
Change the adapter feature complitation to ignore file name and only account for build name.

PR #93, opened 2023-09-27
Deploy/modules

CyberHoward:
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/contracts/account/manager/tests/adapters.rs
```diff
@@ -379,7 +387,7 @@ fn account_install_adapter() -> AResult {
         .claim_namespace(TEST_ACCOUNT_ID, "tester".to_owned())?;
 
     let adapter = BootMockAdapter1V1::new_test(chain);
-    adapter.deploy(V1.parse().unwrap(), MockInitMsg)?;
+    adapter.deploy(V1.parse().unwrap(), MockInitMsg, DeployStrategy::Try)?;
```
Buckram123, line 390: Hm, it feels a bit annoying. What if we put `DeployStrategy` into App/Adapter structs with default strategy "Try".
And in case we want to customize it make a trait, something like:
```
adapter.deploy_strategy(DeployStrategy::Force).deploy(..)
```

framework/packages/abstract-interface/src/native/version_control.rs
```diff
@@ -193,6 +193,16 @@ where
 
         Ok(module.reference.unwrap_app()?)
     }
+    /// Does not page over the apps
+    pub fn remove_all_apps(&self) -> Result<(), crate::AbstractInterfaceError> {
+        let modules = self.module_list(None, None, None)?.modules;
```
Buckram123, line 198: I guess it would be used mainly for tests, but `module_list()` uses `DEFAULT_LIMIT` of 10. We can do multiple queries to load all modules and remove them afterwards. Or let's specify what does "Does not page over the apps" means

scripts/wasm-all.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Detect the architecture #
-if [[ $(arch) == "arm64" ]]; then
+if [[ $(arch) == "arm64a" ]]; then
```
Buckram123, line 4: Was it typo, or now it is? :D

modules/contracts/adapters/dex/Cargo.toml
```diff
@@ -31,7 +31,7 @@ juno = [
   "dep:cw20_junoswap",
   "abstract-wyndex-adapter/full_integration",
 ]
-terra = ["dep:terraswap", "abstract-astroport-adapter/full_integration"]
+terra2 = ["dep:terraswap", "abstract-astroport-adapter/full_integration"]
 neutron = ["abstract-astroport-adapter/full_integration"]
 osmosis = ["abstract-osmosis-adapter/full_integration"]
```
Buckram123, line 36: package.metadata.optimizer still has only "terra"

modules/contracts/adapters/dex/src/lib.rs
```diff
@@ -49,7 +50,10 @@ pub mod interface {
         }
         fn wasm(&self) -> WasmPath {
             artifacts_dir_from_workspace!()
-                .find_wasm_path("abstract_dex_adapter")
+                .find_wasm_path_with_build_postfix(
+                    "abstract_dex_adapter",
+                    BuildPostfix::<Chain>::ChainName(self.get_chain()),
+                )
```
Buckram123, line 56: VERY COOL NEW FEATURE!

modules/contracts/adapters/cw-staking/Cargo.toml
```diff
@@ -38,7 +38,7 @@ interface = [
 ]
 juno = ["abstract-wyndex-adapter/full_integration", "dep:cw-controllers"]
 osmosis = ["abstract-osmosis-adapter/full_integration"]
-terra = ["abstract-astroport-adapter/full_integration"]
+terra2 = ["abstract-astroport-adapter/full_integration"]
```
Buckram123, line 41: package.metadata has terra(not 2)

Buckram123 (commented):
Sick update!

framework/packages/abstract-interface/src/native/version_control.rs
CyberHoward: I just added this because I screwed up the deployment and needed to re-deploy the apps.

scripts/wasm-all.sh
CyberHoward: whoeps, tried compiling with x86 on arm but took way too long. good spot.

modules/contracts/adapters/dex/Cargo.toml
CyberHoward: Can you handle this in #88 ?

Buckram123: Sure

framework/packages/abstract-interface/src/native/version_control.rs
CyberHoward: just removed it

framework/contracts/account/manager/tests/adapters.rs
CyberHoward: I think default should actually be `Error`. Besides that adding a field to the interface struct requires way more code (because it needs to be added to the structs manually). 

I think this approach is pretty good. I had three different functions before but this way the dev really has to decide what they expect when the module is already registered.

modules/contracts/adapters/dex/Cargo.toml
CyberHoward: nvm, fixed it.

PR #95, opened 2023-09-29
Re-add typescript logic to template

CyberHoward:
title

### DIFF
--- a/app-template/template-setup.sh
+++ b/app-template/template-setup.sh
@@ -1,10 +1,5 @@
 #!/usr/bin/env bash
 
-# Add the CI to this repo
-git remote add ci https://github.com/AbstractSDK/rust-ci
-git fetch ci
-git merge ci/main --allow-unrelated-histories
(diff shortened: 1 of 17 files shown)


app-template/template-setup.sh
```diff
@@ -1,10 +1,5 @@
 #!/usr/bin/env bash
 
-# Add the CI to this repo
-git remote add ci https://github.com/AbstractSDK/rust-ci
-git fetch ci
-git merge ci/main --allow-unrelated-histories
-
```
Buckram123, line 7: It gives an error on your machine? What if we update `rust-ci` repo, just merge it everywhere manually?

CyberHoward: We don't really need this because the CI is already included in the template dir. We use the ci repo for ease of pushing out changes to the different rust repos that need CI.

Buckram123 (approved):
LGTM

PR #96, opened 2023-09-30
Fix challange app

CyberHoward:
Fixes the challange app by accepting an expiration data for the end of the challange and updating the internal logic occardingly


modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -458,15 +456,13 @@ fn charge_penalty(deps: DepsMut, app: &ChallengeApp, challenge_id: u64) -> AppRe
     }
 
     let compute_amount_per_friend = || -> Result<u128, AppError> {
-        if challenge.total_check_ins == 0 || num_friends == 0 {
+        if num_friends == 0 {
             return Err(AppError::Std(StdError::generic_err(format!(
-                "Cannot compute amount per friend. total_check_ins: {}, num_friends: {}",
-                challenge.total_check_ins, num_friends
+                "Cannot compute amount per friend. num_friends: {}",
+                num_friends
```
Buckram123, line 462: Cute how we use variable here, when we know the exact string here :rofl:

Buckram123 (commented):
Using `cw_utils::Expiration` seems much more flexible!

modules/contracts/apps/challenge/src/msg.rs
```diff
@@ -122,29 +123,15 @@ pub struct ChallengeRequest {
     /// Desciption of the challenge
     pub description: String,
     /// In what period challenge should end
-    pub end: DurationChoice,
+    pub end: Expiration,
+    /// How much should be paid out when challenge fails.
+    pub strike_amount: u128,
```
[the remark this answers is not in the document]
CyberHoward: I think collateral is the total amount of funds that the user is depositing for the challange while strike_amount is the amount that is charged for each failure.

modules/contracts/apps/challenge/src/state.rs
```diff
@@ -48,54 +50,12 @@ impl ChallengeEntry {
             name: request.name,
             collateral: request.collateral,
             description: request.description,
-            end: Self::to_timestamp(request.end),
-            total_check_ins: 0,
+            end: request.end,
+            strike_amount: 10,
```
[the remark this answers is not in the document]
CyberHoward: I just used some amount, can be whatever!

modules/contracts/apps/challenge/src/msg.rs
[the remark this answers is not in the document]
Buckram123: > I think collateral is the total amount of funds that the user is depositing for the challange while strike_amount is the amount that is charged for each failure.

Okay, so we can fail multiple times and every failure would get charged. Still would prefer to have only one "amount" for managing those

modules/contracts/apps/challenge/src/handlers/execute.rs
CyberHoward: Yeah, also the use of StdErrors here is bad. All of them should be replaced.

framework/packages/abstract-core/src/objects/voting.rs
```diff
@@ -0,0 +1,203 @@
...
+        store: &dyn Storage,
+        vote_id: VoteId,
+        voter: &Addr,
+    ) -> AbstractResult<Option<Vote>> {
+        self.votes.load(store, (vote_id, voter)).map_err(Into::into)
+    }
+
+    // TODO: do we want pagination on this?
+    // I guess no, because if we can't load all votes in one tx means we can't count it
+    // so maybe another TODO: Figure out max len for it
```
CyberHoward, line 239: Or keep a state with 
- total votes for
- total votes agains

Buckram123:
@CyberHoward

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -152,349 +163,228 @@ fn update_friends_for_challenge(
...
         }
-        UpdateFriendsOpKind::Remove => {
-            remove_friends_from_challenge(deps, info, app, challenge_id, friends)
+        UpdateFriendsOpKind::Remove {} => {
+            CHALLENGE_FRIENDS.update(deps.storage, challenge_id, |current_friends| {
+                let mut current_friends = current_friends.ok_or(AppError::ZeroFriends {})?;
+                for rem_friend in friends.iter() {
+                    current_friends.retain(|friend| friend != rem_friend);
+                }
+                AppResult::Ok(current_friends)
```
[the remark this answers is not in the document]
Buckram123: Yeah, it's not possible to reach voting "threshold" in case of 0 friends, so it just will be counted as successful challenge.
But it's just a obviously bad usage of an app, so idk if we should care about it :)

modules/contracts/apps/challenge/src/state.rs
```diff
@@ -16,104 +19,64 @@ pub struct Config {
...
 impl ChallengeEntry {
     /// Creates a new challenge entry with the default status of Uninitialized and no admin strikes.
-    pub fn new(request: ChallengeRequest) -> Self {
+    pub fn new(request: ChallengeRequest, end: Expiration, vote_id: ProposalId) -> Self {
         ChallengeEntry {
             name: request.name,
-            collateral: request.collateral,
+            strike_asset: request.strike_asset,
+            strike_strategy: request.strike_strategy,
             description: request.description,
```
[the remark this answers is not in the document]
Buckram123: Added that and made description optional!

modules/contracts/apps/challenge/src/handlers/execute.rs
```diff
@@ -152,349 +163,228 @@ fn update_friends_for_challenge(
...
+        .cloned()
+        .map(|human| human.check(deps.as_ref(), app))
+        .collect::<AbstractSdkResult<_>>()?;
+
+    let (voters_addrs, friends): (Vec<Addr>, Vec<Friend<Addr>>) =
+        friends_validated.into_iter().unzip();
+
+    let proposal_id = challenge.current_proposal_id;
+
+    SIMPLE_VOTING
```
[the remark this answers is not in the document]
CyberHoward: Because changing the friends during the voting process can mess up the state of the proposal? 

It's like asking 3 people for their opinion but before hearing all of them you swap out someone who doesn't share your view. 

So I think the set of friends should be fixes as long as there are any proposals open. Does that make sense?

framework/packages/abstract-core/src/objects/voting.rs
```diff
@@ -126,7 +127,7 @@ impl<'a> SimpleVoting<'a> {
     pub fn new_proposal(
         &self,
         store: &mut dyn Storage,
-        end: Expiration,
+        end: Timestamp,
```
CyberHoward, line 130: What's the reasoning for doing only timestamps?

Buckram123: Since we now have implicit transition to veto period, it will be simply easier to maintain one type of durations.
For example: we have veto period as 5 blocks and vote duration 50 seconds, when do we transition then? Giving a user to chose one option and stick with it would be also annoying, because some challenge contract instances would be only block-based, other timestamp based

framework/packages/abstract-core/src/objects/voting.rs
```diff
@@ -31,7 +31,13 @@
 //! * List of items queries allowed by `query_` prefix
 //!
 //! ## Details
-//! All methods that modify proposal will return [`ProposalInfo`] to allow logging or checking current status of proposal
+//! All methods that modify proposal will return [`ProposalInfo`] to allow logging or checking current status of proposal.
+//! 
+//! Each proposal goes through the following stages:
+//! 1. Active: proposal is active and can be voted on
+//! 2. WaitingForCount: voting period is finished and awaiting counting
+//! 3. VetoPeriod (optional): voting is counted and veto period is active
```
Buckram123, line 39: Swap 2 and 3 right?

framework/packages/abstract-core/src/objects/voting.rs
```diff
@@ -31,7 +31,13 @@
...
+//! 4. Finished: proposal is finished and count is done
```
Buckram123, line 40: And it's 
A) Finished by count
B) Finished by veto
C) Finished by cancel during voting period

Buckram123: Also 
1. can be voted or canceled on

Buckram123: Thanks for adding proposal lifecycle!

framework/packages/abstract-core/src/objects/voting.rs
Buckram123: And you said that we want to have count after the veto period

CyberHoward: Yup! Was still reading through the code haha

modules/contracts/apps/challenge/tests/integrations.rs
```diff
@@ -201,402 +178,555 @@ fn test_should_successful_install() -> anyhow::Result<()> {
...
-    );
-    assert_eq!(
-        created.challenge.as_ref().unwrap().description,
-        challenge_req.description
-    );
-    assert_eq!(
-        created.challenge.as_ref().unwrap().status,
-        ChallengeStatus::Active
-    );
+        .query::<ChallengeResponse>(&challenge_query)?
```
CyberHoward, line 191: Is the macro not working?

framework/packages/abstract-core/src/objects/voting.rs
CyberHoward: don't worry haha!!

modules/contracts/apps/challenge/tests/integrations.rs
CyberHoward: guess it was just something left from Richard

framework/packages/abstract-core/src/objects/voting.rs
CyberHoward: what do you mean with `canceled` on?

PR #97, opened 2023-10-07
Hackmos Betting App

adairrr:
This PR creates a new betting app for use on the dorahacks platform and chain. It allows Account owners to create rounds, which hold teams, that can be bet upon to win tokens.


framework/packages/abstract-sdk/src/apis/verify.rs
```diff
@@ -105,7 +105,7 @@ impl<'a, T: AccountVerification> AccountRegistry<'a, T> {
         }
     }
 
-    fn account_id(&self, maybe_core_contract_addr: &Addr) -> AbstractSdkResult<AccountId> {
+    pub fn account_id(&self, maybe_core_contract_addr: &Addr) -> AbstractSdkResult<AccountId> {
```
Buckram123, line 108: Not sure why it was not exposed. @CyberHoward is there any downside exposing this?
I will add doc-comment if we are okay with that

adairrr:
Closing this in favor of https://github.com/wagemos/wagemos

framework/packages/abstract-sdk/src/apis/verify.rs
CyberHoward: Nope should be good!

PR #102, opened 2023-10-18
Configure codecov

AndrewPochapsky:
Configures codecov following cw-orch as a guide as well as the [codecov docs for how to have separate components](https://docs.codecov.com/docs/github-4b-getting-aggregate-coverage). Not sure if this will work for this package but they are both cargo workspaces so it should. Will try again after merging if it doesn't end up working. We already have a codecov action configured for the `framwork` directory which is specified [here](https://github.com/AbstractSDK/abstract/blob/main/.github/workflows/framework-test.yml#L42).

codecov.yml from cw-orch: https://github.com/AbstractSDK/cw-orchestrator/blob/main/.github/codecov.yml

Validated codecov.yml file: 
```
[andrew@Andrews-MBP-2 ~/git/abstract] (codecov) $ curl --data-binary @.github/codecov.yml https://codecov.io/validate

Valid!

{
  "coverage": {
    "range": [
      85.0,
      100.0
    ],
    "round": "down",
    "precision": 1,
    "status": {
      "project": {
        "default": {
          "threshold": 1.0
        }
      }
    }
  },
  "ignore": [
    "^tests.*",
    "(?s:.*/examples.*)\\Z",
    "(?s:.*/schema\\.rs.*)\\Z"
  ],
  "comment": {
    "layout": "files",
    "require_changes": true
  },
  "component_management": {
    "individual_components": [
      {
        "component_id": "framework",
        "name": "framework",
        "paths": [
          "^framework/.*"
        ]
      },
      {
        "component_id": "modules",
        "name": "modules",
        "paths": [
          "^modules/.*"
        ]
      }
    ]
  }
}
```

### Checklist

- [x] CI is green.
- [ ] Changelog updated.


CyberHoward:
So the thing here is that I think you need to tell codecov explicitly for which part of the monorepo the coverage was compiled. 

If you look into the `framework-test` and `modules-test` actions you can see that they upload the coverage there as well (for modules the coverage is actually generated on circle-ci). 

The reason why we do it this way is so we only generate coverage for the workspace where we did changes which reduces build times. 

I'd suggest reading the codecov documentation a bit and figure out how to upload a report for a specific part of the repo.

CyberHoward (approved):

.github/workflows/modules-test.yml
```diff
@@ -38,6 +38,34 @@ jobs:
       - name: cargo test --locked
         working-directory: ./modules
         run: cargo test --locked --all-features --all-targets
+  coverage:
```
CyberHoward, line 41: We can't run this here. The github runners are too light. We have a circleCI job running [here](https://github.com/AbstractSDK/abstract/blob/main/.circleci/config.yml#L119C7-L119C24) that does the coverage for the modules. 

So when uploading the file you don't have to specify anything related to the component to tell codecov which compontent you're uploading for? 

Specifically looking in this action: https://github.com/codecov/codecov-action

Don't know, just wondering

AndrewPochapsky: I don't think we need to specify anything in particular to tell it which component to look at, the components are just used for the structure of the report I think. As in, when we run the coverage it will see which coverage comes from which folders and format accordingly. I could be wrong though, first time using this tool :)  Good catch on the circleci stuff, I will update this to just use that.

CyberHoward: Ahh okey! Thanks for looking into it and explaining :)

AndrewPochapsky: Latest commit uncomments the commented out action which triggers circleci to run the codecov job. I am not sure why this was commented out but I think this should work.

CyberHoward: Because the coverage takes a long time to compute so it burns through our plan.

CyberHoward (approved):

PR #105, opened 2023-10-23
Merge release 0.19 changes

CyberHoward:
Merges the release-19 changes into main. 

Also fixes the optimizer builds caused by the inclusion of cw-orch-interchain in the framework workspace by moving the tests and scripts to their own workspace.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


app-template/Cargo.toml
```diff
@@ -37,23 +37,23 @@ cosmwasm-std = { version = "1.2" }
 cosmwasm-schema = { version = "1.2" }
 cw-controllers = { version = "1.0.1" }
 cw-storage-plus = "1.1.0"
-thiserror = { version = "<=1.0.48" }
+thiserror = { version = "1.0.49" }
```
Buckram123, line 40: Glad to see `thiserror` finally fixed the bug!
It would use the `1.0.50` version
```suggestion
thiserror = { version = "1.0.50" }
```

framework/.cargo/config.toml
```diff
@@ -9,8 +9,5 @@ store = "run-script store"
 instantiate = "run-script instantiate"
 migrate = "run-script migrate"
 
-[net]
-git-fetch-with-cli = true
-
```
Buckram123, line 14: :rocket:

framework/Cargo.toml
```diff
@@ -48,7 +46,7 @@ cw-address-like = { version = "1.0" }
 
 schemars = "0.8"
 serde = { version = "1.0", default-features = false, features = ["derive"] }
-thiserror = { version = "<=1.0.48" }
+thiserror = { version = "1.0.49" }
```
Buckram123, line 49:
```suggestion
thiserror = { version = "1.0.50" }
```

framework/packages/abstract-sdk/src/lib.rs
```diff
@@ -7,7 +7,7 @@
 pub type AbstractSdkResult<T> = Result<T, crate::error::AbstractSdkError>;
 
 /// The Abstract Core crate which contains the state and message objects for the native contracts. Also contains helper objects.
-pub extern crate abstract_core as core;
+pub use abstract_core as core;
```
Buckram123, line 10: We should consider rename re-export it as `core` because of rust core library

interchain/Cargo.toml
Buckram123, line 1: Nice

modules/Cargo.toml
```diff
@@ -31,38 +31,41 @@ cw-storage-plus = "~1.1"
...
+cw20 = { git = "https://github.com/AbstractSDK/cw-plus.git", version = "1.0.1", tag = "v1.1.1" }
 
 cw-asset = { version = "3.0" }
 cw-ownable = { version = "0.5" }
 cw-address-like = { version = "1.0" }
 
 schemars = "0.8"
 serde = { version = "1.0", default-features = false, features = ["derive"] }
-thiserror = { version = "<=1.0.48" }
+thiserror = { version = "1.0.49" }
```
Buckram123, line 45:
```suggestion
thiserror = { version = "1.0.50" }
```

Buckram123 (approved):

app-template/artifacts/app.wasm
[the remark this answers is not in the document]
CyberHoward: let ci include it

framework/packages/abstract-sdk/src/lib.rs
CyberHoward: Yeah, that's why we don't re-export anymore...

Buckram123: re-export could fix the "same type with different version" errors. For example users will only use `abstract-app` for the "App" and re-use `abstract-interface`, `abstract-sdk`, `abstract-core`, from it without dealing with weird version errors

CyberHoward: I'll create a separate task for this!

top level crates are abstract-app and abstract-adapter so you're saying they should both export abstract-sdk AND abstract-core?

Buckram123: > AND cw-orchestrator

But this one could be probably re-exported by interface?

PR #109, opened 2023-10-24
Ibc client unit tests for execute

AndrewPochapsky:
This PR adds unit tests for the ibc client execute functions. I also made a few small changes in the contract code itself which I consider improvements, but willing to discuss! These are: 
1. Returning an explicit `Unauthorized` error when a callback is not called from a note contract rather than a `DoesNotExist` one which might be confusing. 
2. Remove a clone on the critical path in turn for a clone on the error path. In my view this would be more optimal unless we expect to process errors more frequently.

### Checklist

- [x] CI is green. - Nope, CI hates me it seems.
- [ ] Changelog updated.


CyberHoward:
CI looking good! Idk what's up with the docs. Going to give it a couple more runs before worrying about it.

CyberHoward (approved):
nice, LGTM!

PR #112, opened 2023-10-26
Ibc host tests

AndrewPochapsky:
Most of the testing was already done for this contract in `tests/integration.rs` so I just added a single test to verify that `send_all_back` works. Though I wasn't too sure how to verify that funds were actually sent, that may need to wait until the end-to-end test. 

There was also a single unit test for testing a query which I removed in favour of including it in the existing integration tests.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/contracts/native/ibc-host/tests/integration.rs
```diff
@@ -189,3 +214,66 @@ fn execute_action_with_account_creation() -> anyhow::Result<()> {
...
+    let abstr = Abstract::load_from(chain.clone())?;
+
+    let account_sequence = 1;
+    let chain = "juno";
+
+    // We need to set the sender as the proxy for juno chain
+    admin_abstr
+        .ibc
+        .host
+        .register_chain_proxy(chain.into(), sender.to_string())?;
```
CyberHoward, line 237: I think it's a bit confusing to use the sender as the proxy in this test. 

Can you add a `polytone_proxy` address and use the `abstr.ibc.host.call_as(polytone_proxy)...` syntax below?

CyberHoward (changes requested):

PR #113, opened 2023-10-30
[Add] Abstract client package

cyberhoward:
The abstract-interface package is currently a mess of different contracts, interfaces and logic. The goal of this task is to create a unified API which should make it easier to script and deploy with abstract.

This initial PR contains the logic to create a new publisher/account, get an existing publisher/account from a namespace, and publish and install apps onto accounts. 

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-client/src/account.rs
```diff
@@ -0,0 +1,162 @@
...
+        Ok(Self::new(abstract_account))
+    }
+
+    pub fn get_account_info(&self) -> AbstractClientResult<AccountInfo<Addr>> {
+        let info_response: InfoResponse = self.abstr_account.manager.info()?;
+        Ok(info_response.info)
+    }
+
+    // Install an application on the account
+    // creates a new sub-account and installs the application on it.
```
AndrewPochapsky, line 139: Does a sub account automatically get created when we install an app on an account?

CyberHoward: No, you need to create the sub-account and then install the app on that sub-account.

framework/packages/abstract-interface/src/deployers.rs
```diff
@@ -7,6 +7,23 @@ use cw_orch::prelude::*;
 use semver::Version;
 use serde::Serialize;
 
+// TODO: Find a way to construct a module M from a contract instance as the `cw_orch::contract::Contract` can be constructed from the ID and chain.
```
CyberHoward, line 10: These comments can be removed

framework/packages/abstract-client/src/account.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    pub fn governance_details(self, governance_details: GovernanceDetails<String>) -> Self {
+        Self {
+            governance_details: Some(governance_details),
+            ..self
+        }
+    }
+
+    pub fn build(self) -> AbstractClientResult<Account<Chain>> {
```
CyberHoward, line 84: Whenever possible it's recommended to make the builder function itself take `&self` because then you can easily create multiple accounts with the same / slightly different settings.

framework/packages/abstract-client/src/account.rs
```diff
@@ -0,0 +1,162 @@
...
+            name: None,
+            description: None,
+            link: None,
+            namespace: None,
+            base_asset: None,
+            governance_details: None,
+        }
+    }
+
+    pub fn name(self, name: impl Into<String>) -> Self {
```
CyberHoward, line 42: In general it's considered best-practice to use `&mut self` for builders: 
https://blog.logrocket.com/build-rust-api-builder-pattern/

framework/packages/abstract-client/tests/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod app;
```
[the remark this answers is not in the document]
AndrewPochapsky: I did not see this, will use it instead :)

framework/packages/abstract-client/src/account.rs
```diff
@@ -0,0 +1,162 @@
...
+};
+use cw_orch::contract::Contract;
+use cw_orch::prelude::*;
+use serde::Serialize;
+
+use crate::{
+    application::Application, client::AbstractClientResult, infrastructure::Infrastructure,
+};
+
+pub struct AccountBuilder<'a, Chain: CwEnv> {
```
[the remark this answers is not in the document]
AndrewPochapsky: Good point, will add.

framework/packages/abstract-client/src/account.rs
AndrewPochapsky: Ah yeah that makes sense, will change.

framework/packages/abstract-client/src/account.rs
AndrewPochapsky: Sure will change, I wasn't really sure which was best in this case.

framework/packages/abstract-client/src/account.rs
AndrewPochapsky: Will do that then, thanks for the clarification :)

framework/packages/abstract-client/tests/lib.rs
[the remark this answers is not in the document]
AndrewPochapsky: A quick attempt to use it runs into the issue that the mock app does not call `abstract_app::cw_orch_interface!` so some extra work would be necessary there. Any reason why we can't refactor the mock app to use this interface? I think it would more closely resemble the customer experience which is what we should aim to do with tests.

framework/packages/abstract-client/src/account.rs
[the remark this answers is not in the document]
AndrewPochapsky: What do you think about adding docs in a later PR once we have dog-fooded the package a bit? I think otherwise we may end up re-writing stuff a lot.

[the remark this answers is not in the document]
CyberHoward: Yup, sounds great!

framework/packages/abstract-client/src/application.rs
```diff
@@ -0,0 +1,36 @@
...
+
+impl<Chain: CwEnv, M> Application<Chain, M> {
+    pub fn new(account: Account<Chain>, module: M) -> Self {
+        Self { account, module }
+    }
+
+    pub fn account(&self) -> &Account<Chain> {
+        &self.account
+    }
+}
```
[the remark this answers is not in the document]
CyberHoward: It will compile but in order to call the function of the module you would need to specify the trait. 

I.e. `ExecuteMsgFns::account(my_app, other_fn_arg)`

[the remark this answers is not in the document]
AndrewPochapsky: Ok nice, that should be fine imo.

framework/packages/abstract-client/src/account.rs
```diff
@@ -0,0 +1,198 @@
...
+        let info_response: InfoResponse = self.abstr_account.manager.info()?;
+        Ok(info_response.info)
+    }
+
+    // Install an application on the account
+    // creates a new sub-account and installs the application on it.
+    pub fn install_app<
+        M: ContractInstance<Chain>
+            + RegisteredModule
+            + InstantiableContract
```
CyberHoward, line 138: should be able to remove this bound for now

framework/packages/abstract-client/src/account.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    let sub_account_id: Option<u32> = wasm_abstract_attributes
+        .into_iter()
+        .find(|a| a.key == "sub_account_added")
+        .map(|a| a.value.parse().unwrap());
+
+    sub_account_id
+}
+
+pub struct InterchainAccount {}
```
CyberHoward, line 198: 👀

framework/packages/abstract-client/src/infrastructure.rs
```diff
@@ -0,0 +1,28 @@
...
+use cw_orch::deploy::Deploy;
+use cw_orch::prelude::*;
+
+use crate::account::{Account, AccountBuilder};
+
+pub(crate) trait Infrastructure<T: CwEnv> {
+    // Get the execution environment
+    fn environment(&self) -> T;
+
+    // Get the infrastructure on the execution environment
```
CyberHoward, line 11: Can make these doc-comments

framework/packages/abstract-client/src/publisher.rs
```diff
@@ -0,0 +1,116 @@
...
+        C: Serialize,
+    >(
+        &self,
+        configuration: &C,
+        funds: &[Coin],
+    ) -> AbstractClientResult<Application<Chain, M>> {
+        self.account.install_app(configuration, funds)
+    }
+
+    pub fn deploy_module<
```
CyberHoward, line 90: I would only expose `deploy_adapter` and `deploy_app` as in the adapters's case the contract needs to be instantiated. 

Also would be nice to re-use the deployer traits here `AppDeployer` and `AdapterDeployer`. You can add them as a trait bond on M in the previously mentioned functions. 

Edit: I see you're already using AppDeployer here, so you definitly need to split the function up.

framework/packages/abstract-interface/src/deployers.rs
```diff
@@ -7,6 +7,11 @@ use cw_orch::prelude::*;
 use semver::Version;
 use serde::Serialize;
 
+pub trait RegisteredModule {
```
CyberHoward, line 10: Doc-comment for this trait would be great to have at this point.

framework/packages/abstract-client/src/account.rs
```diff
@@ -0,0 +1,198 @@
...
+
+impl<Chain: CwEnv> Account<Chain> {
+    pub(crate) fn new(abstract_account: AbstractAccount<Chain>) -> Self {
+        Self {
+            abstr_account: abstract_account,
+        }
+    }
+
+    pub(crate) fn from_namespace(
+        abstr: &Abstract<Chain>,
```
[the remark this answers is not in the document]
AndrewPochapsky: This was meant to be used only internally hence the `crate` modifier and there is a public method on the client that calls this.

framework/packages/abstract-client/src/account.rs
AndrewPochapsky: Haha I think this was some of the starter code you added initially. I haven't though much about how to handle ibc related things here but it should be possible, what do you think?

framework/packages/abstract-client/src/account.rs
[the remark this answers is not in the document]
CyberHoward: whoeps, looked over that. Thanks!

PR #117, opened 2023-10-31
Replace to_binary with to_json_binary and from_binary with from_json

AndrewPochapsky:
`to_binary` is now deprecated in favour of `to_json_binary` which is causing clippy to fail in https://github.com/AbstractSDK/abstract/pull/116 This PR replaces all instances of `to_binary` and `from_binary` to `to_json_binary` and `from_json` respectively. The following commands were used for those curious after realizing this was going to complain for all of the files: 

`find . -type f -name "*.rs" -exec gsed -i 's/from_binary/from_json/g' {} +` 

`find . -type f -name "*.rs" -exec gsed -i 's/to_binary/to_json_binary/g' {} +`

`gsed` is the version of `sed` on GNU which is easier to work with than the sed from BSD, which is what is shipped with mac (how I can't wait to go back to a linux machine lol). 

### Checklist

- [x] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/contracts/account/manager/src/commands.rs
+++ b/framework/contracts/account/manager/src/commands.rs
@@ -40,7 +40,7 @@ use abstract_sdk::{
     ModuleRegistryInterface,
 };
 use cosmwasm_std::{
-    ensure, from_binary, to_binary, wasm_execute, Addr, Attribute, Binary, Coin, CosmosMsg, Deps,
+    ensure, from_json, to_json_binary, wasm_execute, Addr, Attribute, Binary, Coin, CosmosMsg, Deps,
     DepsMut, Empty, Env, MessageInfo, Response, StdError, StdResult, Storage, WasmMsg,
 };
 use cw2::{get_contract_version, ContractVersion};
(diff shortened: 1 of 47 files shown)


framework/contracts/account/manager/src/commands.rs
```diff
@@ -924,7 +924,7 @@ pub fn update_internal_config(
 ) -> ManagerResult {
     // deserialize the config action
     let action: InternalConfigAction =
-        from_binary(&config).map_err(|error| ManagerError::InvalidConfigAction { error })?;
+        from_json(config).map_err(|error| ManagerError::InvalidConfigAction { error })?;
```
Buckram123, line 927: `from_json` only needs value to take `impl AsRef<[u8]>`, was clippy complaining when you were passing it by reference?

Buckram123 (approved):
CI passes, so LGTM! Haven't noticed anything suspicious, ~~except removed references for `from_binary` -> `from_json` transition,~~ but that's even better!

framework/contracts/account/manager/src/commands.rs
Buckram123: Oh yeah, I see "needless borrow" nvm

PR #119, opened 2023-11-01
compiling Cargo.lock on main

Buckram123:
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Buckram123:
Just noticed we have a bit too much gitignores. Do we want to merge them all into one?
image

CyberHoward:
> Just noticed we have a bit too much gitignores. Do we want to merge them all into one? image

https://linear.app/abstract-sdk/issue/ABS-225/merge-duplicate-files-in-monorepo

CyberHoward:
Doesn't seem like anyone is actually blocked on this so moving to draft for now.

CyberHoward:
No problems with compiling so no need for the lock

PR #125, opened 2023-11-06
docs: fix broken links

jvr0x:
This PR fixes broken links in the Abstract documentation.

### Checklist

- [x] CI is green.
- [ ] Changelog updated.


framework/docs/src/3_framework/5_modularity.md
```diff
@@ -11,9 +11,8 @@ crafted by other developers available in the module marketplace. Additionally, y
 custom modules using the [Abstract SDK](https://docs.rs/abstract-sdk/latest/abstract_sdk), tailoring them to specific
 application needs and even contributing back to the community.
 
-```admonish info
```
CyberHoward, line 14: Ouh, you don't like the admonish? Or was it interfering with the url checker? @jvr0x

CyberHoward: image
Because the link seems to render fine on the current docs

CyberHoward: Also curious to why you prefer html tags over the markdown syntax?

jvr0x: I like them but they don't render links properly, they show litelly as `[text](link...)`

jvr0x: In addition, not sure if it has to do with the current theme of the docs or what, but using the admonish info or warning label for example doesn't have any visual difference, just that the user needs to scroll horizontally if the message is too long.

In that sense using md quotes is a better fit.

jvr0x: Oh, i did not see them rendered that way! weird...

jvr0x: I use the html tags only when we link something that is outside of the docs. There's no way to open a link on a new tab with markdown, sadly.

CyberHoward: > Oh, i did not see them rendered that way! weird...

Only when you build locally? You might be missing the admonish pre-processor.

Do you see them in the production docs? like here: https://docs.abstract.money/3_framework/5_modularity.html

CyberHoward: `cargo install mdbook-admonish --vers "1.9.0" --locked`

jvr0x: > > Oh, i did not see them rendered that way! weird...
> 
> Only when you build locally? You might be missing the admonish pre-processor.
> 
> Do you see them in the production docs? like here: https://docs.abstract.money/3_framework/5_modularity.html

Yes i see it here, it was a local thing after all... Alright gonna revert these changes 😅

CyberHoward (approved):

PR #127, opened 2023-11-06
Merge duplicate files in monorepo

Buckram123:
Closes ABS-225
### Checklist

- [x] CI is green.
- [x] Changelog updated.


.gitattributes
```diff
@@ -1 +1,3 @@
-**/artifacts/* binary
\ No newline at end of file
+**/artifacts/* binary
+**/schema/*.json linguist-generated=true
+**/docs/book linguist-generated=true
```
CyberHoward, line 3: this is also in gitignore so probably not needed here?

justfile
```diff
@@ -49,3 +49,9 @@ schema-modules:
     (cd $path; cargo run --example schema --features schema); 
   done
   set +e
+
+schema: schema-modules
```
CyberHoward, line 53: We should have one for framework as well

CyberHoward: there's a cargo-workspace command that lets you run commands on every member of a workspace.

CyberHoward: https://crates.io/crates/cargo-workspaces

CyberHoward: See the justfile in modules!

CyberHoward:
```
# TODO this doesn't work for contracts like the croncat one which is the croncat prefix
publish-schemas:
  SCHEMA_OUT_DIR=$(cd ../../schemas && echo "$PWD") \
    cargo ws exec --no-bail bash -lc 'cargo schema && { jq -r .contract_version schema/module-schema.json > _version.txt; outdir="$SCHEMA_OUT_DIR/abstract/${PWD##*/}/$(cat _version.txt)"; mkdir -p "$outdir"; rm -rf "schema/raw"; cp -a "schema/." "$outdir"; rm _version.txt; }'

```

Buckram123: > ```
> # TODO this doesn't work for contracts like the croncat one which is the croncat prefix
> publish-schemas:
>   SCHEMA_OUT_DIR=$(cd ../../schemas && echo "$PWD") \
>     cargo ws exec --no-bail bash -lc 'cargo schema && { jq -r .contract_version schema/module-schema.json > _version.txt; outdir="$SCHEMA_OUT_DIR/abstract/${PWD##*/}/$(cat _version.txt)"; mkdir -p "$outdir"; rm -rf "schema/raw"; cp -a "schema/." "$outdir"; rm _version.txt; }'
> ```

There's similar script to this one in framework justfile, we can execute that, if we don't plan to remove justfile's

CyberHoward: Yeah, but I don't even know what that command does. 
I think we only need to publish schemas for the modules but not sure. 

@adairrr

adairrr: It runs cargo schema for each of the modules, then copies the contents of the schemas for those modules to the proper directory (name + version) in the schemas folder (which in writing the command was in the `../../schemas`. It's a bit hacky and we should move it to a github action.

CyberHoward: I say we move it to a `.sh` script and call that from the github action. Easier to maintain and we can run it locally as well. 

@Buckram123 Could you do that?

Buckram123: Do we want copy schemas to `../../schemas` in the same script?

framework/justfile
```diff
@@ -54,6 +54,7 @@ migrate +CHAINS:
   just run-script migrate {{CHAINS}}
 
 publish-schemas version:
+  mkdir -p ../../schemas
```
CyberHoward, line 57: We don't want a script to create a dir that's outside of the repo right?

Buckram123: Right

scripts/framework-schema.sh
CyberHoward, line 1: What does this script do?

scripts/modules-schema.sh
CyberHoward, line 1: What does this script do?

justfile
CyberHoward: I don't know, should all the schemas live in to one dir? What's the desired setup here? 
@adairrr

scripts/framework-schema.sh
Buckram123: generate schemas for all framework contracts and copy to the `abstract/../schemas` (if exists)

scripts/modules-schema.sh
Buckram123: generate schemas for all modules and copy to the `abstract/../schemas` (if exists)

schema/execute_msg.json
Buckram123, line 1: Is this file generated by app-template supposed to be here? @CyberHoward

.gitattributes
Buckram123: Yeah, thanks, removed it!

.gitignore
```diff
@@ -1,10 +1,41 @@
...
+**/.DS_Store
+
+# temporary dockerfiles
+**/docker*
+**/.dockerignore
+
+# Schemas
+framework/packages/**/schema
+framework/contracts/**/schema
+./schema
```
Buckram123, line 32:
```suggestion
```
Remove this gitignore line since we decided to move schema inside this repo?

schema/execute_msg.json
CyberHoward: good catch!

PR #128, opened 2023-11-07
Update `abstract-app` and `abstract-sdk` re-exports

cyberhoward:
Update the `abstract-app` crate to re-export `abstract-sdk` under `abstract_app::sdk` and `abstract-core` under `abstract_app::framework`.

If those names are ok I will update `abstract-adapter` as well to follow the same convention. 

This change allows users to exclusively import `abstract-app`.
They will still need to import `abstract-interfaces` and `cw-orch` when compiling with the `interface` flag. 

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-app/src/lib.rs
```diff
@@ -13,6 +13,10 @@ pub use error::AppError;
 pub type AppResult<C = Empty> = Result<Response<C>, AppError>;
 mod interface;
 
+// Re-export core abstract utils and types
+pub use abstract_sdk as sdk;
```
Buckram123, line 17: Not sure about this rename, don't want abstract_sdk compete with some other `sdk`

modules/contracts/apps/croncat/Cargo.toml
```diff
@@ -51,7 +51,7 @@ thiserror = { workspace = true }
 schemars = { workspace = true }
 cw-asset = { workspace = true }
 
-abstract-core = { workspace = true }
+
 abstract-app = { workspace = true }
```
Buckram123, line 55: I think `Cargo.toml` files where abstract-core got removed have one extra newline

Buckram123 (changes requested):
Happy to see re-export abstract packages inside app and adapter! But I feel like `sdk` is a bit too common name and may become an issue one day.
Also it's a bit confusing to have `framework`(abstract-core) package under `framework` workspace
How do you feel about leaving the package names as they are, until we have a cool name for it?

CyberHoward:
> Happy to see re-export abstract packages inside app and adapter! But I feel like `sdk` is a bit too common name and may become an issue one day. Also it's a bit confusing to have `framework`(abstract-core) package under `framework` workspace How do you feel about leaving the package names as they are, until we have a cool name for it?

I feel like that would be a lot of extra characters for importing anything. 

`abstract_app::abstract_core::xyz`
`abstract_app::abstract_sdk::xyz`

Just looks off to me :/ 

@AndrewPochapsky @Kayanski opinions?

framework/packages/abstract-app/src/lib.rs
Buckram123: It could have same issues as we had with `core`, consider this:
1. a little bit annoying: 
```rust
use abstract_app::sdk;
use foo_bar::sdk;

fn aloha() {
    sdk::baz();
}
```
2. a little more annoying
```rust
use abstract_app::sdk;
use foo_bar::sdk;

fn aloha() {
    baz(); // LSP suggests to import sdk::baz()
}
```

Buckram123: > It could have same issues as we had with `core`, consider this:
> 
>     1. a little bit annoying:
> 
> 
> ```rust
> use abstract_app::sdk;
> use foo_bar::sdk;
> 
> fn aloha() {
>     sdk::baz();
> }
> ```
> 
>     2. a little more annoying
> 
> 
> ```rust
> use abstract_app::sdk;
> use foo_bar::sdk;
> 
> fn aloha() {
>     baz(); // LSP suggests to import sdk::baz()
> }
> ```

Ah nevermind, you're right, it won't let you do this part:
> use abstract_app::sdk;
> use foo_bar::sdk;
because it's under abstract_app

PR #130, opened 2023-11-07
Internalize execution stack and Deps

CyberHoward:
Add a runtime constructor for Apps that allows for deps inclusion in the app type as well as an execution stack that can be used to improve the APIs for executing calls.


CyberHoward:
Close in favor of #145

PR #131, opened 2023-11-08
Remove  prefix for builder functions

CyberHoward:
Removes the builder prefix to improve the builder syntax

### DIFF
--- a/app-template/src/contract.rs
+++ b/app-template/src/contract.rs
@@ -20,11 +20,11 @@ pub type AppResult<T = Response> = Result<T, AppError>;
 pub type App = AppContract<AppError, AppInstantiateMsg, AppExecuteMsg, AppQueryMsg, AppMigrateMsg>;
 
 const APP: App = App::new(APP_ID, APP_VERSION, None)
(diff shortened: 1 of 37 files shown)


CyberHoward:
Shoot. If we remove the prefix they conflict with the handler trait functions (`MYAPP.instantiate(deps, env, info, msg)`)

Buckram123:
> Shoot. If we remove the prefix they conflict with the handler trait functions (`MYAPP.instantiate(deps, env, info, msg)`)

One way around it is to use distinct builder type, it was anyway weird that `with_instantiate`, etc. is accessible at any time

CyberHoward:
> > Shoot. If we remove the prefix they conflict with the handler trait functions (`MYAPP.instantiate(deps, env, info, msg)`)
> 
> One way around it is to use distinct builder type, it was anyway weird that `with_instantiate`, etc. is accessible at any time

I think we should work towards this: https://github.com/AbstractSDK/abstract/pull/130

That will separate the runtime VS compile-time types so this issue will disappear.

PR #140, opened 2023-11-12
Fix small issues in payment app

adairrr:
This change:
- check that destination asset exists on init 
- removes max_spread, because both belief price and max_spread must be provided?

Issues:
- swap tests with mock "exceed max spread"

Still necessary:
- check exchange exists on init 
- ability change dest asset
- Ability to configure a "quote" asset that's a string. i.e. "usd", "eur", etc. To be used by the FE.

### Checklist

- [x] CI is green.
- [x] Changelog updated.


CyberHoward:
> * removes max_spread, because both belief price and max_spread must be provided?

It shouldn't. Can check in the wyndex integration if this is broken.

Buckram123:
> > * removes max_spread, because both belief price and max_spread must be provided?
> 
> It shouldn't. Can check in the wyndex integration if this is broken.

All of our swap tests on wyndex uses max_spread 30 and just checked and there is no integration forces to include both on our side. Would be nice to see the error to see if some dex actually forces it

Buckram123:
>  * Ability to configure a "quote" asset that's a string. i.e. "usd", "eur", etc. To be used by the FE.

@CyberHoward can you please give more info on this? Like AssetEntry but without a chain trace?

CyberHoward:
> >  * Ability to configure a "quote" asset that's a string. i.e. "usd", "eur", etc. To be used by the FE.
> 
> @CyberHoward can you please give more info on this? Like AssetEntry but without a chain trace?

No, just a `String`!

modules/contracts/apps/payment/src/handlers/execute.rs
```diff
@@ -79,6 +86,11 @@ pub fn tip(
 
     // Search for trading pairs between the deposited assets and the desired asset
     for pay_asset in asset_entries {
+        // No need to swap if desired asset sent
+        if pay_asset.name == desired_asset {
+            total_amount += pay_asset.amount;
+            continue;
+        }
         // query the pools that contain the desired asset
         let query_msg = ans_host::QueryMsg::PoolList {
```
CyberHoward, line 95: We should have this as a function on `AnsHost` that accepts an optional filter and does the smart-query.

modules/contracts/apps/payment/src/handlers/execute.rs
```diff
@@ -117,7 +126,7 @@ pub fn tip(
     }
 
     // Tip history
-    update_tipper_history(deps.storage, &info.sender, total_amount)?;
+    update_tipper_history(deps.storage, &info.sender, &desired_asset, total_amount)?;
```
CyberHoward, line 129: The logic for this function is incomplete imo. 

the function should accept `Vec<AnsAsset>` where 

1. Any asset that could be swapped to the desired asset is swapped and added to the desired asset's amount
2. Any assets that could NOT be swapped are added to the `Vec`. That way even payments that weren't swappable will still show up. 

Maybe a snapshot map could also be useful for this to simplify the FE integration (i.e. show different state over different time periods without needing an Indexer)

modules/contracts/apps/payment/src/msg.rs
```diff
@@ -31,6 +32,8 @@ pub enum AppExecuteMsg {
     #[cfg_attr(feature = "interface", payable)]
     Tip {},
     UpdateConfig {
+        desired_asset: Option<AssetEntry>,
```
CyberHoward, line 35: There's no way to set desired_asset to None in the contract

CyberHoward (changes requested):
Some initial comments

modules/contracts/apps/payment/src/handlers/execute.rs
Buckram123: >     2. Any assets that could NOT be swapped are added to the `Vec`. That way even payments that weren't swappable will still show up.

Hm, do we want to have some validation of that asset? Like query ans if that asset at least exist? Not sure how annoying spam assets in cosmos yet

CyberHoward: Ah but the assets are reversed queried so if the asset isn't in ANS is won't work anyway

modules/contracts/apps/payment/src/msg.rs
CyberHoward: Will be adressed by this issue: https://linear.app/abstract-sdk/issue/ABS-269/add-clearable-type-in-abstract-sdk

modules/contracts/apps/payment/src/handlers/execute.rs
Buckram123: > We should have this as a function on `AnsHost` that accepts an optional filter and does the smart-query.

We have `pool_list()` method already, updated to use that instead

CyberHoward (approved):

Buckram123 (approved):

PR #143, opened 2023-11-13
Remove requirement for base message in app instantiate

Buckram123:
This PR aims to remove need for providing `BaseInstantiateMsg` by user, and moving it to module factory. Also some API refactors related to this change 
### Checklist

- [ ] CI is green.
- [x] Changelog updated.


Buckram123:
Not sure if we should listen beta clippy

CyberHoward:
> Not sure if we should listen beta clippy

Can do in another PR

framework/contracts/native/module-factory/src/commands.rs
```diff
@@ -29,14 +29,21 @@ pub fn execute_create_modules(
...
-    let version_registry = binding.module_registry(deps.as_ref());
-    let account_registry = binding.account_registry(deps.as_ref());
+    let version_registry = version_control.module_registry(deps.as_ref());
+    let account_registry = version_control.account_registry(deps.as_ref());
 
     // assert that sender is manager
     let account_base = account_registry.assert_manager(&info.sender)?;
 
+    // App base message
+    let app_base_msg = abstract_core::app::BaseInstantiateMsg {
```
CyberHoward, line 41: We only need to construct this when instantiating an app so this can be moved down.

AndrewPochapsky (approved):
Huge win for developer experience! Can you please confirm that module tests pass locally? The CI doesn't work for that and there are quite a few changes there.

framework/packages/abstract-app/src/endpoints/instantiate.rs
```diff
@@ -35,7 +31,7 @@ impl<
         SudoMsg,
     >
 {
-    type InstantiateMsg = InstantiateMsg<Self::CustomInitMsg>;
+    type InstantiateMsg = InstantiateMsg<Binary>;
```
CyberHoward, line 34: Why do we need to lose this type information?

framework/contracts/native/module-factory/src/commands.rs
```diff
@@ -89,12 +89,25 @@ pub fn execute_create_modules(
...
             ModuleReference::App(code_id) => {
+                let init_msg = owner_init_msg.unwrap();
+                // App base message
+                let app_base_msg = abstract_core::app::BaseInstantiateMsg {
+                    ans_host_address: config.ans_host_address.to_string(),
+                    version_control_address: version_control.address.to_string(),
+                    account_base: account_base.clone(),
+                };
+
+                // TODO: App will have to do one extra deserialize, is it avoidable?
```
Buckram123, line 100: @CyberHoward left this TODO, in case we find better solution

CyberHoward: You can use `Value`! https://crates.io/crates/serde-cw-value/0.7.0

CyberHoward: I'd prefer keeping as much type information as possible and working with Value where required.

Buckram123: Yeah we still do one extra deserialize, but in module factory, so it's cleaner, thanks!

CyberHoward: You mean we deserialize the custom message into `Value` to then serialize it again within the bigger struct?

PR #144, opened 2023-11-13
CI/Modules tests without all features

Buckram123:
Don't have actions run permissions to check, so creating PR
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Buckram123:
Not enough to pass CI :sob:

PR #146, opened 2023-11-14
Remove caching of modules to install for freshly created accounts

Buckram123:
This PR changes manager and proxy instantiation to use instantiate2 approach, which helps with
- Removing caching of the modules to install
- Removing admin period for account factory
### Checklist

- [ ] CI is green.
- [x] Changelog updated.


framework/contracts/account/proxy/src/commands.rs
```diff
@@ -216,7 +218,8 @@ mod test {
             assert_that(&res).is_ok();
 
             let actual_modules = load_modules(&deps.storage);
-            assert_that(&actual_modules).has_length(1);
+            // Plus manager
```
CyberHoward, line 221: Plus proxy I presume?

framework/contracts/account/proxy/src/contract.rs
```diff
@@ -36,23 +37,38 @@ pub type ProxyResult<T = Response> = Result<T, ProxyError>;
...
-    let admin_addr = Some(info.sender);
-    ADMIN.set(deps, admin_addr)?;
+    let ans_host = AnsHost {
+        address: deps.api.addr_validate(&msg.ans_host_address)?,
+    };
+    ANS_HOST.save(deps.storage, &ans_host)?;
+    let admin_addr = Some(manager_addr);
+    ADMIN.set(deps.branch(), admin_addr)?;
+
+    if let Some(base_asset) = msg.base_asset {
```
CyberHoward, line 63: Wow, were we just ignoring this until now?

framework/contracts/native/account-factory/src/commands.rs
```diff
@@ -107,123 +105,124 @@ pub fn execute_create_account(
...
+    };
 
-/// instantiates the Treasury contract of the newly created DAO
-pub fn after_proxy_create_manager(
-    deps: DepsMut,
-    env: Env,
-    result: SubMsgResult,
-) -> AccountFactoryResult {
-    let config = CONFIG.load(deps.storage)?;
+    // TODO: should be possible to parse checksums from artifacts/checksums.txt
```
CyberHoward, line 127: This code is only ran when an account is created so not that important to optimize.

framework/contracts/native/account-factory/src/commands.rs
```diff
@@ -107,123 +105,124 @@ pub fn execute_create_account(
...
-                label: format!("Proxy of Account: {}", account_id),
-                msg: to_json_binary(&ProxyInstantiateMsg {
-                    account_id,
-                    ans_host_address: config.ans_host_contract.to_string(),
-                })?,
-            }
-            .into(),
-            reply_on: ReplyOn::Success,
-        }))
+    let salt = generate_salt(env.block.height, &account_id);
```
CyberHoward, line 109: For this application there shouldn't be a need to add the block height to the salt because for every account you can only create once and there's no re-install possible. Removing the height would make it easy to figure out what the proxy address for any account (also interchain accounts) will be. 

I.e. we can send funds to an address that we know will be assigned to the proxy of an account on chain B while the original account is on chain A. 

Also in the `generate_salt` function we take a part of the account's trace to create the salt. I think we should hash the trace and use that instead. That allows the trace to be of any length.

framework/contracts/native/account-factory/src/commands.rs
```diff
@@ -245,164 +244,39 @@ fn query_module(
 /// adds proxy contract address to Manager
```
CyberHoward, line 244: nit: update doc-comment

framework/contracts/account/proxy/src/contract.rs
Buckram123: No, I moved that field from account-factory context :sunglasses:

framework/contracts/native/ibc-host/tests/integration.rs
CyberHoward, line 1: Can remove these comments now?

framework/contracts/native/version-control/src/commands.rs
```diff
@@ -357,12 +370,40 @@ pub fn claim_namespace(
         });
     }
 
+    let fee_msg = claim_namespace_internal(
```
CyberHoward, line 373: line 366 the account factory should no longer have permissions to claim namespaces through this function.

framework/contracts/native/account-factory/src/commands.rs
Buckram123: > This code is only ran when an account is created so not that important to optimize.

Pretty sure it will be a cool feature for `cosmwasm-std`!
Agree, it's not that important, but that would be a nice gas optimization, considering checksum will be queried for every account (should we save it at instantiation?)

CyberHoward: No, it's fine :)

framework/contracts/native/version-control/src/commands.rs
Buckram123: Great nit, thanks!

framework/contracts/native/account-factory/src/commands.rs
Buckram123: Or we can just pass zero as block height here, wdyt?

CyberHoward: sha256 is pretty quick and easy

CyberHoward: > Or we can just pass zero as block height here, wdyt?

I'd prefer having different options (fns). I feel like we might want more options in the future.

Buckram123: Using hash now and added different fn for account address salt

cyberhoward (approved):

PR #149, opened 2023-11-16
Dogfood AbstractClient in calendar module tests

AndrewPochapsky:
Replaces the test setup using the internal types in the calendar module tests with the new abstract-client package as a way of dogfooding. It was a very easy change (almost drop-in) and revealed a few missing parts in the client, notably missing `DerefMut` for `Application` and a way of accessing `AnsHost` from client. It also revealed a way of prettying up the syntax by creating an `AbstractClientBuilder` to be used to initializing balances and `AnsHost` things, so that was also created here. 

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/packages/abstract-client/Cargo.toml
+++ b/framework/packages/abstract-client/Cargo.toml
@@ -12,6 +12,7 @@ keywords.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
+default = ["test-utils"]
 test-utils = ["cw-asset", "abstract-testing"]
 
 [dependencies]
--- a/framework/packages/abstract-client/Cargo.toml
+++ b/framework/packages/abstract-client/Cargo.toml
@@ -11,6 +11,9 @@ keywords.workspace = true
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+test-utils = ["cw-asset", "abstract-testing"]
+
 [dependencies]
 cosmwasm-std.workspace = true
 abstract-interface.workspace = true
@@ -20,7 +23,12 @@ serde.workspace = true
 semver.workspace = true
 thiserror.workspace = true
 
+# Used for test-utils feature
+cw-asset = { workspace = true, optional = true }
+abstract-testing = {workspace = true, optional = true }
+
 [dev-dependencies]
+abstract-client = { path = ".", features = ["test-utils"] }
 cw-asset.workspace = true
 cw-controllers.workspace = true
 abstract-app = { workspace = true, features = ["test-utils"] }
--- a/framework/packages/abstract-client/src/client.rs
+++ b/framework/packages/abstract-client/src/client.rs
@@ -19,7 +19,7 @@ impl<Chain: CwEnv> AbstractClient<Chain> {
         Ok(Self { abstr })
     }
 
-    pub fn ans_host(&self) -> &AnsHost<Chain> {
+    pub fn name_service(&self) -> &AnsHost<Chain> {
         &self.abstr.ans_host
     }
 
--- a/framework/packages/abstract-client/src/client.rs
+++ b/framework/packages/abstract-client/src/client.rs
@@ -1,14 +1,16 @@
-use abstract_interface::{Abstract, AnsHost};
+use abstract_interface::{Abstract, AnsHost, VersionControl};
+use cosmwasm_std::BlockInfo;
 use cw_orch::{deploy::Deploy, prelude::CwEnv};
 
 use crate::{
     account::{Account, AccountBuilder},
     error::AbstractClientError,
+    infrastructure::Infrastructure,
     publisher::{Publisher, PublisherBuilder},
 };
 
 pub struct AbstractClient<Chain: CwEnv> {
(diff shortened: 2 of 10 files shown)


framework/packages/abstract-client/src/client.rs
```diff
@@ -19,6 +19,10 @@ impl<Chain: CwEnv> AbstractClient<Chain> {
         Ok(Self { abstr })
     }
 
+    pub fn ans_host(&self) -> &AnsHost<Chain> {
```
CyberHoward, line 22: Call this `name_service`?
```suggestion
    pub fn name_service(&self) -> &AnsHost<Chain> {
```

modules/contracts/apps/calendar/tests/integration.rs
```diff
@@ -100,58 +95,47 @@ fn setup_with_time(
...
+            monarch: ADMIN.to_owned(),
+        })
+        .namespace("my-namespace")
+        .build()?;
 
-    // claim the namespace so app can be deployed
-    abstr_deployment
-        .version_control
-        .claim_namespace(AccountId::local(1), "my-namespace".to_string())?;
+    publisher.deploy_app::<AppInterface<Mock>>()?;
```
CyberHoward, line 120: Since it's a publisher, should `deploy` -> `publish`?
```suggestion
    publisher.publish_app::<AppInterface<Mock>>()?;
```

modules/contracts/apps/calendar/tests/integration.rs
```diff
@@ -84,12 +84,7 @@ fn request_meeting(
 fn setup_with_time(
```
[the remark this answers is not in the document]
CyberHoward: Or an `AbstractTestingClient`?

modules/contracts/apps/calendar/tests/integration.rs
AndrewPochapsky: Should we also change `deploy_adapter` to `publish_adapter`?

framework/packages/abstract-client/src/client.rs
```diff
@@ -19,6 +21,20 @@ impl<Chain: CwEnv> AbstractClient<Chain> {
...
+    }
+
+    pub fn version_control(&self) -> &VersionControl<Chain> {
+        &self.abstr.version_control
+    }
+
+    // Is there a way for us to return `AbstractClientResult<BlockInfo>` here instead? I think we
+    // would need to add <T: CwEnv> to `AbstractClientError` which I am not sure we want to do.
+    pub fn block_info(&self) -> Result<BlockInfo, <Chain as cw_orch::prelude::TxHandler>::Error> {
+        self.environment().block_info()
```
CyberHoward, line 35: This error type supports `Into<CwEnvError>` which is then convertable to the client error

framework/packages/abstract-client/Cargo.toml
```diff
@@ -11,6 +11,9 @@ keywords.workspace = true
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+test-utils = ["cw-asset", "abstract-testing"]
```
CyberHoward, line 15: enable this by default? 

```suggestion
default = ["test-utils"]
test-utils = ["cw-asset", "abstract-testing"]
```

modules/contracts/apps/calendar/tests/integration.rs
```diff
@@ -84,74 +83,42 @@ fn request_meeting(
...
-    // claim the namespace so app can be deployed
-    abstr_deployment
-        .version_control
-        .claim_namespace(AccountId::local(1), "my-namespace".to_string())?;
-
-    app.deploy(APP_VERSION.parse()?, DeployStrategy::Try)?;
-
-    account.install_app(
-        &app,
+) -> anyhow::Result<(Application<Mock, AppInterface<Mock>>, AbstractClient<Mock>)> {
```
CyberHoward, line 86: We should rename `AppInterface` to `CalendarApp` or `CalendarAppInterface` or something.

CyberHoward (commented):
For the publisher creation we should add a check for the namespace. If it is already claimed then we should load the addresses for the account that claimed it and create the publisher from that instead of creating a new one.

CyberHoward (commented):
Another issue are apps that have dependencies on other apps. I.e. the balancer app has a dependency on the ETF app. In that scenario the `install_app` will fail because the account doesn't have the ETF app installed on it. 

I think we should create a seperate trait that needs to be implemented on the app interface where that configuration can live. I'll post a code snippet on the ADR in Notion

CyberHoward:
Also we should be able to install the app in the same action as creating the sub-account (in the `install_app` fn)

```rust
let sub_account_response = self.abstr_account.manager.create_sub_account(
            vec![], // <---- Here we can provide the app's install config
            "Sub Account".to_owned(),
            None,
            None,
            None,
            None,
            &[],
        )?;
```

modules/contracts/apps/calendar/tests/integration.rs
CyberHoward: Not sure. It both instantiates and registers them. Could that cause confusion? Should `publish` only do the publishing part and not instantiate it?

PR #150, opened 2023-11-16
SDK AuthZ API

Buckram123:
~~MERGE #147 FIRST~~
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-sdk/src/apis/authz.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+impl GenericAuthorization {
+    /// Create new generic authorization
+    pub fn new(msg: String) -> Self {
+        Self { msg }
+    }
+}
+
+trait AuthZAuthrization {
```
CyberHoward, line 135: What's the goal/point of this trait?

CyberHoward: Would be nice to have something that also accepts raw proto messages (`Vec<u8>`)

framework/packages/abstract-sdk/src/apis/authz.rs
```diff
@@ -0,0 +1,246 @@
...
+    /// Grants generic authorization to a **grantee**.
+    ///
+    /// # Arguments
+    ///
+    /// * `grantee` - The address of the grantee.
+    /// * `msg` - Allowed message type url.
+    pub fn grant_generic(
+        &self,
+        grantee: &Addr,
+        msg: String,
```
CyberHoward, line 95:
```suggestion
        msg_type_url: String,
```

framework/packages/abstract-sdk/src/apis/authz.rs
CyberHoward: If there are other grant types (besides GenericAuthorization) then we should also include those in here. And imo that should be in a nested mod

```rust
mod authorizations {
    // trait

    // different impls
}
```

Buckram123: > What's the goal/point of this trait?

It helps structuring different types of grants

> If there are other grant types (besides GenericAuthorization) then we should also include those in here. And imo that should be in a nested mod
I will add other types of grants. Yeah, nested mod is fine

CyberHoward (approved):
LGTM!

PR #151, opened 2023-11-16
Fix namespace registration fee for account creation

Buckram123:
### Checklist

- [x] CI is green.
- [x] Changelog updated.


CyberHoward (changes requested):
I think interchain account creation should be free. Once you have an initial account somewhere you should be able to create interchain accounts for that account without any payment of sorts.

Buckram123:
> I think interchain account creation should be free. Once you have an initial account somewhere you should be able to create interchain accounts for that account without any payment of sorts.

This PR only impacts namespace claiming fee, not account creation

CyberHoward (approved):
Awesome catch and fix!! :)

framework/contracts/native/account-factory/tests/create_account.rs
```diff
@@ -325,6 +332,76 @@ fn create_one_account_with_namespace() -> AResult {
...
+        Some(namespace_to_claim.to_string()),
+        // Account creation fee not covered
+        &[],
+    );
+    assert!(err
+        .unwrap_err()
+        // Error type is inside contract, not the package
+        .root()
+        .to_string()
+        .contains("Invalid fee payment sent."));
```
Buckram123, line 390: @CyberHoward wdyt do we want to move error types to packages as well?

CyberHoward: Error types of the contract? I'd prefer them to stay in the contract.

PR #159, opened 2023-11-22
Assorted abstract client improvements

AndrewPochapsky:
A collection of abstract client improvements: 
1. Adds `type InitMsg` to `RegisteredModule` to bind module install function to take a specific instantiate message
2. Adds logic to `AccountBuilder::build` to return an account with the namespace if it is already claimed. I decided to put this behind a flag to avoid confusion by having it be default logic. The situation I envisioned was a new user trying to create an account with a namespace that already existed and getting back an `Account` for which they do not have the keys. They would then need to debug an `AccessDenied` exception without realizing that it was because they don't own the account itself. I think that this should be possible given that these are all queries that do not check that the caller actually has access to the account (please correct me if I am wrong!).
3. Updates install app logic to install the app in the sub account creation action. This required adding some more event parsing logic to get the address of the module.
4.  Updates return type of `fn block_info` to return an `AbstractClientResult`.

### Checklist

- [x] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-client/src/account.rs
```diff
@@ -66,6 +70,13 @@ impl<'a, Chain: CwEnv> AccountBuilder<'a, Chain> {
         self
     }
 
+    // Would it bet better to make this a "switch" by removing `value` and always setting the flag
```
CyberHoward, line 73: Let's keep it with the bool exposed.

framework/packages/abstract-client/src/account.rs
```diff
@@ -176,17 +207,30 @@ impl<Chain: CwEnv> Account<Chain> {
     }
 }
 
-fn get_sub_account_id_from_events(events: Vec<Event>) -> Option<u32> {
+struct ExtractedEventInfo {
```
CyberHoward, line 210: What "extracted info"? This type could use a better name

framework/packages/abstract-client/src/account.rs
```diff
@@ -176,17 +207,30 @@ impl<Chain: CwEnv> Account<Chain> {
...
+    sub_account_id: Option<u32>,
```
CyberHoward, line 211: Both these fields get unwrapped so these should probably not be `Option`

cyberhoward (changes requested):

framework/packages/abstract-client/src/account.rs
AndrewPochapsky: Haha yeah, I am not too sure how else to call it. The two things we extract are not really related as far as I can tell so this was the best I could come up with. Any ideas?

AndrewPochapsky: Ok, how is `CreationAttributeValues`? Since all of the values are things that were "created".

CyberHoward: `ParsedAccountCreationEvent`?

AndrewPochapsky: How about `ParsedAccountCreationResponse`? It made the related function easier to name as well.

CyberHoward: Sure!

cyberhoward (approved):

PR #156, opened 2023-11-22
[Diff] View diff since last release

CyberHoward:
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

PR #163, opened 2023-11-23
Add migration tests

CyberHoward:
Adds tests for migrations that uses cw-orch clone testing feature. 

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


CyberHoward:
hey @Buckram123 could you give a high-level review on this? 
Tasks I created while working on this: 
https://linear.app/abstract-sdk/issue/ORC-53/add-bank-queries-support-to-cwenv
https://linear.app/abstract-sdk/issue/ABS-260/clean-up-abstract-testing-addresses-mod

framework/contracts/account/manager/Cargo.toml
```diff
@@ -54,6 +54,7 @@ abstract-adapter = { workspace = true, features = ["test-utils"] }
 abstract-app = { workspace = true, features = ["test-utils"] }
 abstract-testing = { workspace = true }
 abstract-macros = { workspace = true }
+abstract-integration-tests = { workspace = true }
```
Buckram123, line 58: Was thinking what if we mocktest migration against previous abstract version? 
```toml
previous-abstract-manager = { package = "abstract-manager", git = "https://github.com/AbstractSDK/abstract/", tag = "v0.19.2" } #TODO don't forget to update it before after every release
```

framework/packages/abstract-integration-tests/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "abstract-integration-tests"
+version = "0.20.0-rc.1"
+edition = { workspace = true }
+license = { workspace = true }
+description = "Package with environment-agnonstic tests"
+
+[lib]
+
```
Buckram123, line 9: Remove if not needed?

framework/packages/abstract-integration-tests/src/mock_modules.rs
```diff
@@ -13,6 +13,7 @@ use cw_orch::prelude::*;
 pub type MockAdapterContract = AdapterContract<AdapterMockError, Empty, Empty, Empty, Empty, Empty>;
 pub type MockAppContract = AppContract<AppMockError, Empty, Empty, Empty, Empty, Empty, Empty>;
 
+#[allow(ambiguous_glob_reexports)]
```
Buckram123, line 16: I would prefer to not allow it

interchain/tests/tests/migrate.rs
```diff
@@ -0,0 +1,57 @@
...
+    abstract_integration_tests::manager::account_install_app(app.clone(), sender)
+}
+
+fn create_sub_account_after_migrate() -> anyhow::Result<()> {
+    let (abstr_deployment, sender, app) = setup()?;
+    abstract_integration_tests::manager::create_sub_account_with_modules_installed(
+        app.clone(),
+        sender,
+    )
+}
```
Buckram123, line 57: What do you think about testing state changes? It could be hard to reach every state item without reading the diff couple times: #156

framework/packages/abstract-integration-tests/src/mock_modules.rs
```diff
@@ -129,57 +130,102 @@ pub mod app_1 {
...
+    macro_rules! gen_standalone_mock {
+        ($name:ident, $id:expr, $version:expr, $cw2:expr) => {
+            use ::abstract_core::app;
+            use ::abstract_sdk::base::Handler;
+            use ::abstract_sdk::features::AccountIdentification;
+            use ::abstract_sdk::{Execution, TransferInterface};
+            use ::cw_orch::prelude::*;
+
+            #[cosmwasm_schema::cw_serde]
+            pub struct MockMsg;
```
Buckram123, line 146: nice

interchain/tests/tests/migrate.rs
```diff
@@ -0,0 +1,57 @@
...
+    Ok((abstr_deployment, sender, app))
+}
+
+#[test]
+fn migrate_infra_success() -> anyhow::Result<()> {
+    let (abstr_deployment, sender, _) = setup()?;
+
+    assert_eq!(abstr_deployment.version_control.code_id()?, 3692);
+    abstr_deployment.migrate_if_needed()?;
+    assert_eq!(abstr_deployment.version_control.code_id()?, 5000003);
```
Buckram123, line 42: don't think code_ids should be consts here

Buckram123 (commented):
Nice progress and tests clean up!

framework/packages/abstract-integration-tests/Cargo.toml
Buckram123: What a quiet library 🥁

framework/contracts/account/manager/Cargo.toml
CyberHoward: We don't have to specify old versions for migration tests. We just need to specify an account that has the old version on-chain and load it! So we need to have a clean account that's on the old version that is already instantiated.

Still need to write the logic to retrieve those.

framework/packages/abstract-integration-tests/Cargo.toml
CyberHoward: I think it was required lol, can remove

framework/packages/abstract-integration-tests/src/mock_modules.rs
CyberHoward: Everything is re-exported that's in there so there shouldn't be any reason why you'd need to index into it (which is the only way this could be an issue) 

If you still not like it I can see about changing the mod names or even removing the internal mods.

interchain/tests/tests/migrate.rs
CyberHoward: State changes are catched by our snapshot tests! #157

framework/contracts/account/manager/Cargo.toml
Buckram123: Ah, I see, this would allow 2+ versions migration testing as well, nice!

interchain/tests/tests/migrate.rs
CyberHoward: Then where should I retrieve them? They won't change (unless we change the network we test against) 

We'll encounter the same issue for tests related to balances...

CyberHoward: Ohhh but I see what you're saying. this code-id will change when we do our next migration...

framework/packages/abstract-integration-tests/src/mock_modules.rs
CyberHoward: I might be mis-understanding it. Investigating...

interchain/tests/tests/migrate.rs
Buckram123: > Ohhh but I see what you're saying. this code-id will change when we do our next migration...

It will change on the next commit actually :rofl:
Uh I just realized that we can't check previous version code_id from Abstract object, right? I guess we can parse it from github/last_version_tag/state.json, but that could be easier to do manually

framework/packages/abstract-integration-tests/src/mock_modules.rs
Buckram123: TY for fixing!

interchain/tests/tests/migrate.rs
Buckram123: > It will change on the next commit actually 🤣

Nvm I didn't know `ForkMock` starts with hardcoded code_id

CyberHoward: Yup! It's because cw-multi-test does that as well afaik

CyberHoward:
@Buckram123 Does this work for you locally? CI seems to be unhappy

CyberHoward:
> > @Buckram123 Does this work for you locally? CI seems to be unhappy
> 
> From PR Description:
> 
> > Note: before running migration tests copy state.json of previous version
> 
> I didn't want to accidentally push `state.json` to 0.20 release. :P

Can you just push state with this? I'm going to filter the state anyway during migrations.

framework/Cargo.toml
```diff
@@ -90,11 +90,17 @@ abstract-adapter-utils = { version = "0.20.0", path = "packages/standards/utils"
...
+abstract-integration-tests = { path = "packages/abstract-integration-tests" }
+
 ## Testing
 rstest = "0.17.0"
 speculoos = "0.11.0"
 anyhow = "1"
 
+[patch.crates-io]
+cw-orch-core = { git = "https://github.com/AbstractSDK/cw-orchestrator", branch = "update/add-readonly-flag-daemon-state", version = "0.18.0" }
+cw-orch = { git = "https://github.com/AbstractSDK/cw-orchestrator", branch = "update/add-readonly-flag-daemon-state", version = "0.18.0" }
```
Buckram123, line 102: TODO: remove patch here and in `interchain` after release

PR #165, opened 2023-11-27
Remove Adapter `Remove` endpoint

CyberHoward:
Removes the broken `Remove` endpoint on the adapter.

The remove endpoint was used to un-register an adapter from its dependencies.

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-adapter/src/endpoints/execute.rs
```diff
@@ -18,6 +18,7 @@ use cosmwasm_std::{
     wasm_execute, Addr, CosmosMsg, Deps, DepsMut, Env, MessageInfo, QuerierWrapper, Response,
     StdError, StdResult,
 };
+use cosmwasm_std::{Addr, Deps, DepsMut, Env, MessageInfo, Response};
```
Buckram123, line 21: Oopsie 2 Addrs

Buckram123 (approved):
abstract and modules tests don't fail. Schemas a little bit out-dated but IMO we should add it to the end of WASM job so it does 2 commits 1 for wasm artifacts and 1 for schemas

PR #166, opened 2023-11-27
[ABS-263] Replace `cw-controllers` with `cw-ownable` for auth handling in ibc_cilent

AndrewPochapsky:
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


AndrewPochapsky:
Snapshot test failures seem to be unrelated I think

CyberHoward:
> Snapshot test failures seem to be unrelated I think

It's not unrelated. The item storage object was changed, which means that both the content of that item (the state) and the namespace under which it is stored (defined by the const) were changed. 

If we were to have deployed this contract in a previous release then we would have been required to add migration logic for this change! Does that make sense? It's an important concept to grasp.

CyberHoward (approved):

PR #167, opened 2023-11-28
Update app interface names

CyberHoward:
Update the App interface names to be different from the app's type.


Buckram123 (approved):
clippy told me I should approve

PR #169, opened 2023-11-29
Fix/account state logic

CyberHoward:
Updates `Abstract` to make `Account` private. 

Also updates the factory's account creation logic to return an account with it's contracts having unique IDs. This prevents the account's state to be overwritten. 

TODO:

- [x] Ensure an `Account` that uses the generic `abstract:manager` and `abstract:proxy` id's can not be created.
- [x] Fix tests that are broken by this change 

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-interface/src/native/account_factory.rs
```diff
@@ -77,17 +77,30 @@ impl<Chain: CwEnv> AccountFactory<Chain> {
...
+        // set addresses
         let manager_address = &result.event_attr_value(ABSTRACT_EVENT_TYPE, "manager_address")?;
         self.get_chain()
             .state()
-            .set_address(MANAGER, &Addr::unchecked(manager_address));
+            .set_address(&manager_id, &Addr::unchecked(manager_address));
         let proxy_address = &result.event_attr_value(ABSTRACT_EVENT_TYPE, "proxy_address")?;
         self.get_chain()
             .state()
             .set_address(PROXY, &Addr::unchecked(proxy_address));
```
Buckram123, line 99: That is really nice! set_address for proxy should be &proxy_id as well, I suppose?

framework/packages/abstract-core/src/objects/account/account_trace.rs
```diff
@@ -181,6 +181,22 @@ impl AccountTrace {
...
 
+impl TryFrom<&str> for AccountTrace {
+    type Error = AbstractError;
+
+    fn try_from(trace: &str) -> Result<Self, Self::Error> {
+        if trace == LOCAL {
+            Ok(Self::Local)
+        } else {
+            let chain_trace: Vec<ChainName> = trace
+                .split(CHAIN_DELIMITER)
```
Buckram123, line 192: nit: do we want to rename it to
`TRACE_DELIMETER = CHAIN_DELIMETER`?

framework/packages/abstract-interface/src/account/proxy.rs
```diff
@@ -1,16 +1,19 @@
...
+use abstract_core::{objects::AccountId, proxy::*, PROXY};
 
 use cw_orch::{interface, prelude::*};
 
 #[interface(InstantiateMsg, ExecuteMsg, QueryMsg, MigrateMsg)]
 pub struct Proxy<Chain>;
 
+// TODO: would be nice to hide default `new`
+impl<Chain: CwEnv> Proxy<Chain> {
+    pub fn new_from_id(account_id: &AccountId, chain: Chain) -> Self {
```
CyberHoward, line 11: I'm wondering if the presumption is going to be that this account already has its state stored locally? So when you create this object and you interact with it you're not sure it has the state it requires (its address) as that's only available in the version control. 

In that case you need to load it from versions control directly or?

interchain/tests/src/bin/setup_funds.rs
```diff
@@ -82,7 +82,7 @@ pub fn test_send_funds() -> AnyResult<()> {
     // )?;
 
     // Get the ibc client address
-    let remote_account = AbstractAccount::new(&abstr_stargaze, Some(remote_account_id.clone()));
+    let remote_account = AbstractAccount::new(&abstr_stargaze, remote_account_id.clone());
     let client = remote_account.proxy.config()?;
```
CyberHoward, line 86: For example I think this query will fail because the remote account doesn't have an address associated with it. It's purely constructed from its ID but not fetched from the version control. I think it makes sense to only allow `AbstractAccount` creation (outside this crate) from:

`AccountFactory::create_account()`
`VersionControl::get_account()`

Otherwise an account object can be created without it having valid state.

Buckram123: > For example I think this query will fail because the remote account doesn't have an address associated with it. It's purely constructed from its ID but not fetched from the version control.

`AbstractAccount::new` currently fetches account addresses from the version control

> I think it makes sense to only allow `AbstractAccount` creation (outside this crate) from:
> 
> `AccountFactory::create_account()` `VersionControl::get_account()`
> 
> Otherwise an account object can be created without it having valid state.

Proxy/Manager - yes, account - not really

PR #170, opened 2023-11-30
Change account to sub-account

Buckram123:
Did a small test and it turns out transfers was supported!
So I decided to add recursive acceptance of the ownership for better user experience
### Checklist

- [x] CI is green.
- [x] Changelog updated.


CyberHoward:
And it registers the sub-account on the main account after the transfer is accepted? damn!

framework/contracts/account/manager/src/contract.rs
```diff
@@ -206,6 +206,16 @@ pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg) -> M
                         cw_ownable::Action::AcceptOwnership => update_governance(deps.storage)?,
                         _ => vec![],
```
CyberHoward, line 207: We should clear the cache state (proposed governance) when the `RenounceOwnership` variant is called.

framework/contracts/account/manager/src/contract.rs
```diff
@@ -206,6 +206,16 @@ pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg) -> M
...
                     };
+                    let mut info = info;
+                    // if new governance is sub-account trick cw-ownable to think it was called by proxy
```
CyberHoward, line 210: I am very confused by this, could you elaborate on why this is needed?

CyberHoward: Ahhhh I get it!! Okey haha, sorry

framework/contracts/account/manager/src/contract.rs
```diff
@@ -197,7 +197,9 @@ pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg) -> M
                     handle_sub_account_action(deps, info, action)
                 }
                 ExecuteMsg::Callback(CallbackMsg {}) => handle_callback(deps, env, info),
+                // Used to claim or renounce an ownership change.
                 ExecuteMsg::UpdateOwnership(action) => {
+                    // Disallow the user from using the TransferOwnership action.
                     let msgs = match action {
                         // Disallow the user from using the TransferOwnership action
                         cw_ownable::Action::TransferOwnership { .. } => {
```
AndrewPochapsky, line 205: Can't comment directly on the line but should we rename the `ShouldUseSetOwner` error variant now?

framework/contracts/account/manager/src/error.rs
```diff
@@ -88,7 +88,7 @@ pub enum ManagerError {
     #[error("invalid configuration action, {}", error)]
     InvalidConfigAction { error: StdError },
 
-    #[error("Must use SetOwner to change owner")]
+    #[error("Must use ProposeOwner to change owner")]
     MustUseSetOwner {},
```
AndrewPochapsky, line 92: Should we also rename the variant itself?

framework/contracts/account/manager/src/contract.rs
Buckram123: > I am very confused by this, could you elaborate on why this is needed?

Our `execute_update_ownership` macro checks new owner against `info.sender`, and to support top-level admin claims we need to edit sender when it's called by the top-level admin :)

framework/contracts/account/manager/src/error.rs
Buckram123: Sure, updated!

framework/contracts/account/manager/src/contract.rs
Buckram123: moved it into `update_governance` function as well, so pending owner can't renounce himself(which is okay but still minor bug)

CyberHoward (approved):

PR #171, opened 2023-12-01
Update VC settings and registration procedure

CyberHoward:
Updates the VC instantiate message to dis-allow direct module registration. Also update the deployment logic to self-approve abstract modules on deployments


framework/packages/abstract-interface/src/deployment.rs
```diff
@@ -93,6 +93,9 @@ impl<Chain: CwEnv> Deploy<Chain> for Abstract<Chain> {
             .version_control
             .register_natives(deployment.contracts())?;
 
+        // Approve abstract contracts if needed
+        deployment.version_control.approve_any_abstract_modules()?;
+
```
Buckram123, line 98: Shouldn't we do it on a contract side?
Auto-accept if it's coming from Account(0) owner which has ABSTRACT_NAMESPACE namespace

Buckram123:
> @Buckram123 I feature-flagged the boolean because if we enable the approval for tests then we need to change A LOT of tests.

Nice! In the future we can add this boolean as part of the `Abstract::DeployData` in case we want to have it true for some chains(testnets) for example

framework/packages/abstract-interface/src/native/version_control.rs
```diff
@@ -141,6 +140,33 @@ where
...
+
+        if proposed_abstract_modules.modules.is_empty() {
+            return Ok(());
+        }
+
+        self.approve_or_reject_modules(
+            proposed_abstract_modules
+                .modules
+                .iter()
+                .map(|m| m.module.info.clone())
```
AndrewPochapsky, line 163: Can we make this `.into_iter().map(|m| m.module.info)`?

AndrewPochapsky (approved):

PR #176, opened 2023-12-07
Release 0.20

CyberHoward:
Tracking branch for `v0.20`

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/Cargo.toml
Buckram123, line 1: Don't forget to include `Cargo.lock` files so wasms will remain reproducible

Buckram123: @CyberHoward so what do you think? Want it remain compilable in the future

CyberHoward: Yeah you're right, should include lock file

Buckram123 (approved):
LGTM

PR #177, opened 2023-12-07
Internalize dependencies

Kayanski:
This PR aims at internalizing the dependencies inside abstract apps.
This change DOESN'T include Sylvia. This is purely changing the way abstract deals with the deps object

Closes ABS-220

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Kayanski:
> Clone #145 in favor if this?

I would like to keep the other one as well, because the Sylvia route has not been abandonned yet

PR #178, opened 2023-12-07
Add support to client for installing an app with dependencies

AndrewPochapsky:
This PR adds a new trait `DependencyCreation` which allows a module to specify how to install its dependencies. This is added as a trait bound to a new `install_app_with_dependencies` function on the abstract client. The `dca` app was used as an example for how this would work in a real module example.

### Checklist

- [x] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-interface/src/deployers.rs
CyberHoward, line 1: I'd suggest moving the non-deployer traits to a different file (module_traits.rs / client_traits.rs / ?)

framework/packages/abstract-client/src/account.rs
```diff
@@ -160,18 +160,73 @@ impl<Chain: CwEnv> Account<Chain> {
...
+    ) -> AbstractClientResult<Application<Chain, M>> {
+        let mut install_configs: Vec<ModuleInstallConfig> =
+            M::dependency_install_configs(dependencies_config)?;
+        install_configs.push(M::install_config(module_configuration)?);
+
+        self.install_app_internal(install_configs, funds)
+    }
+
+    pub fn admin(&self) -> AbstractClientResult<Addr> {
+        self.abstr_account.manager.address().map_err(Into::into)
```
CyberHoward, line 192: The Admin of the account is not the manager, the **direct** admin of a module is the manager, the admin of the account / indirect admin of a module is the owner of the account. 

`.owner` or `.admin` should return the address that controls the account. For sub-accounts that might get tricky (do we return the proxy of the owning account or the top-level owner of all the accounts?) (related, do we want a `SubAccount` struct or not?)

I think an application should have a `.direct_admin` that returns the manager and `.account_admin` or `.account_owner` that returns the obvious

framework/packages/abstract-client/src/account.rs
```diff
@@ -160,18 +160,73 @@ impl<Chain: CwEnv> Account<Chain> {
...
+    }
+
+    pub fn module_infos(
```
CyberHoward, line 195: start_after and limit are really constraints that exist on the smart-contract level. As a developer they are a pain to work with so it'd be great if we could just get all the information and do the paging in this function. 

It'll be a recurring operation so could be worth to write a small macro to reduce the boilerplate for paging the queries.

framework/packages/abstract-client/src/account.rs
```diff
@@ -221,15 +268,25 @@ impl<Chain: CwEnv> Account<Chain> {
             .find(|a| a.key == "sub_account_added")
             .map(|a| a.value.parse().unwrap());
 
-        let module_address: Option<String> = wasm_abstract_attributes
+        let module_addresses: Option<String> = wasm_abstract_attributes
             .iter()
             .find(|a| a.key == "new_modules")
             .map(|a| a.value.parse().unwrap());
 
+        // When there are multiple modules registered the addresses are returned in a common
```
CyberHoward, line 276: This comment is not entirely correct because the instantiation logic actually doesn't assert any order in the install. However, if you can ensure that the dependency logic always puts the top-level module last, then this is correct. But the comment should probably reflect that nuance

framework/packages/abstract-client/src/publisher.rs
```diff
@@ -120,4 +142,19 @@ impl<Chain: CwEnv> Publisher<Chain> {
     pub fn proxy(&self) -> AbstractClientResult<Addr> {
         self.account.proxy()
     }
+
+    pub fn module_infos(
+        &self,
+        start_after: Option<String>,
```
CyberHoward, line 148: Same comment on paging as before

framework/packages/abstract-interface/src/deployers.rs
```diff
@@ -17,6 +19,39 @@ pub trait RegisteredModule {
...
+
+/// Trait to make it easier to construct `ModuleInfo` and `ModuleInstallConfig` for a
+/// `RegisteredModule`.
+pub trait InstallConfig: RegisteredModule {
+    /// Constructs the `ModuleInfo` by using information from `RegisteredModule`.
+    fn module_info() -> Result<ModuleInfo, crate::AbstractInterfaceError> {
+        ModuleInfo::from_id(Self::module_id(), Self::module_version().into()).map_err(Into::into)
+    }
+
+    /// Constructs the ModuleInstallConfig for an App Interface
```
CyberHoward, line 44: Rust is awesome and it support linking to types directly, and the links actually work in an IDE or when reading the docs on docs.rs


```suggestion
    /// Constructs the [`ModuleInstallConfig`] for an App Interface
```

So whenever referring to types in doc-comments, try to link to the type! 
https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

It also supports MD and code-blocks and other cool things.

CyberHoward (commented):
Left a couple comments and some information. I think we can merge this if you fix the `.admin` function on the manager the nit doc-comment and address the paging on a different PR

framework/packages/abstract-app/src/interface.rs
```diff
@@ -154,6 +154,9 @@ macro_rules! cw_orch_interface {
                 }
 			}
 
+            // use default implementations
+			impl<Chain: ::cw_orch::prelude::CwEnv> ::abstract_interface::InstallConfig for $interface_name<Chain> { }
```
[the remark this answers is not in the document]
AndrewPochapsky: This is exactly what I was hoping existed :)

framework/packages/abstract-client/src/account.rs
AndrewPochapsky: Good idea! I'll implement the macro later if necessary since for now we have just a single instance of this.

framework/packages/abstract-client/src/account.rs
AndrewPochapsky: Yes good idea

framework/packages/abstract-interface/src/deployers.rs
AndrewPochapsky: That is awesome, thanks for the tip!

framework/packages/abstract-client/src/account.rs
AndrewPochapsky: Do we not want to pass along the `manager`/`proxy` naming through the client? I think as a consumer `direct_admin` and `account_admin` might be more confusing.

framework/packages/abstract-interface/src/deployers.rs
AndrewPochapsky: Will do that in a separate PR if that is alright.

framework/packages/abstract-client/src/account.rs
```diff
@@ -160,18 +160,84 @@ impl<Chain: CwEnv> Account<Chain> {
...
+        install_configs.push(M::install_config(module_configuration)?);
+
+        self.install_app_internal(install_configs, funds)
+    }
+
+    pub fn direct_admin(&self) -> AbstractClientResult<Addr> {
+        self.abstr_account.manager.address().map_err(Into::into)
+    }
+
+    pub fn account_owner(&self) -> AbstractClientResult<cw_ownable::Ownership<String>> {
```
AndrewPochapsky, line 195: We return `String` and not `Addr` from the manager since we use the [cw_ownable_query](https://docs.rs/cw-ownable/latest/cw_ownable/attr.cw_ownable_query.html) macro. Weird to me that they defaulted to a `String` and not `Addr` though.

framework/packages/abstract-client/src/account.rs
AndrewPochapsky: I already added a `proxy` query earlier it seems. I think it would be nice to keep `manager` and `proxy` for consistency with our docs.

framework/packages/abstract-client/src/account.rs
CyberHoward: Yeah IMO it should return Addr. However, we can change the query to let it return `Ownership<Addr>` if we want because `struct Addr(String)` is serialized/de-serialized just like a String.

framework/packages/abstract-client/src/account.rs
CyberHoward: Well, with the sub-accounts `direct_admin` is not an account's admin. The direct admin of any account is either a wallet/multisig/whatever or another account.

CyberHoward: So it returning the manager address doesn't make any sense imo

AndrewPochapsky: Is how it is now incorrect (see new commits)? I thought you said `direct_admin` was manager and `account_owner` was the owner: 

>I think an application should have a .direct_admin that returns the manager and .account_admin or .account_owner that returns the obvious

CyberHoward: For an `Application`, this code is for an `Account`

PR #184, opened 2023-12-11
Fix Version Control Config smart query

Buckram123:
VC Config smart query was returning factory address instead of VC Config. After change it returns 
```rust
    pub allow_direct_module_registration_and_updates: bool,
    pub namespace_registration_fee: cosmwasm_std::Coin,
```
as expected

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-core/src/native/version_control.rs
```diff
@@ -23,13 +23,12 @@ pub mod state {
...
-        account::AccountId, common_namespace::ADMIN_NAMESPACE, module::ModuleInfo,
-        module_reference::ModuleReference, namespace::Namespace,
+        account::AccountId, module::ModuleInfo, module_reference::ModuleReference,
+        namespace::Namespace,
     };
 
     use super::{AccountBase, Config, ModuleConfiguration, ModuleDefaultConfiguration};
 
-    pub const ADMIN: Admin = Admin::new(ADMIN_NAMESPACE);
     pub const FACTORY: Admin = Admin::new("fac");
```
Buckram123, line 32: Noticed that we had this unused `ADMIN` state item. And not sure why FACTORY is `Admin`, instead of `Item<Addr>`

framework/packages/abstract-core/src/native/version_control.rs
```diff
@@ -319,7 +318,8 @@ pub struct NamespaceListResponse {
 
 #[cosmwasm_schema::cw_serde]
 pub struct ConfigResponse {
-    pub factory: Addr,
+    pub allow_direct_module_registration_and_updates: bool,
```
CyberHoward, line 321: Imo it should still include the factory address

framework/packages/abstract-core/src/native/version_control.rs
CyberHoward: Because we do an assertion for the address in the contract and Admin has nice helpers for it. 

But I guess that could be confusing? Open to including it in Config as well.

CyberHoward (changes requested):
Nice spot on the unused Admin!

framework/packages/abstract-core/src/native/version_control.rs
Buckram123: Even though it has nice helpers, imo it will be clearer to have simple `Item<Addr>` and compare to that. Also we still want to have custom error from that, instead of "not admin" on the contract that has `cw-ownable` owner

CyberHoward (approved):

CyberHoward:
Add migration test in #163

CyberHoward:
> > Add migration test in #163
> 
> It will get tested as soon as they will be in the same branch. Are we planning to merge main into 0.20 release?

Yes

PR #185, opened 2023-12-12
Add cw-clearable to abstract

Buckram123:
Sometimes we want to clear state instead of setting it, so we created a small package for it: https://docs.rs/cw-clearable/0.1.2/cw_clearable/enum.Clearable.html

Added `Clearable` type where it is applicable

Merged #184 and #140 here to get access to new messages

Fixes ABS-269, ABS-270
### Checklist

- [x] CI is green.
- [x] Changelog updated.


Buckram123:
Draft until #184 and #140 merged

modules/contracts/apps/payment/src/contract.rs
```diff
@@ -26,10 +26,8 @@ pub type PaymentApp = AppContract<
     cw20::Cw20ReceiveMsg,
 >;
 
-const DEX_DEPENDENCY: StaticDependency = StaticDependency::new(
-    abstract_dex_adapter::DEX_ADAPTER_ID,
-    &[abstract_dex_adapter::contract::CONTRACT_VERSION],
-);
+const DEX_DEPENDENCY: StaticDependency =
+    StaticDependency::new(abstract_dex_adapter::DEX_ADAPTER_ID, &["^0.19.0"]);
```
CyberHoward, line 30: Why use a hard-coded value here?

framework/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Top level account owner now has admin privileges on the apps and adapters
 - Multiple `AbstractAccount`s now don't overlap
 - Top level account owner can now claim pending sub-accounts directly
+- `Clearable` helper type was added to the messages where clearing optional state could be useful
```
AndrewPochapsky, line 32: Why can't we just use `Option`?

CyberHoward: Because `Option<Option<T>>` is kind of an anti-pattern imo. It doesn't give any meaningful type information.

`Some(None)` would clear the value vs `Some(Clear)` with `Clearable`

Buckram123: How json looks for `Some(None)`?

CyberHoward: honestly no idea...

CyberHoward: https://github.com/serde-rs/serde/issues/1042

I don't think it's possible to serialize into a `Some(None)` value actually.

PR #188, opened 2023-12-13
Remove floats from wasm V2 (without removing flatten)

Buckram123:
Implements deserialize without float logic
Keeps flatten opposed to #187

Both tested by this test: https://github.com/AbstractSDK/abstract/blob/477f92d9ccb3a907c8c6a79ef786aa8c66b113db/framework/contracts/account/manager/tests/adapters.rs#L645-L675
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/packages/abstract-core/Cargo.toml
+++ b/framework/packages/abstract-core/Cargo.toml
@@ -32,6 +32,7 @@ cw-orch = { workspace = true, optional = true }
 cw-ownable = { workspace = true }
 polytone = { workspace = true }
 sha256 = { version = "1", default-features = false }
+serde-cw-value = "0.7.0"
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


CyberHoward:
So with this custom deser implementation we can remove the flatten attribute?

` #[serde(flatten)]`

Buckram123:
> So with this custom deser implementation we can remove the flatten attribute? Because that's what was inserting the floating point operation right?
> 
> ` #[serde(flatten)]`

We still need to `Serialize` flatten, so manager passes correct message to the Adapter

Buckram123:
For some reason it fails to optimize out f64 visitors from `FlatMapSerializer`, which obviously should get optimized, so closed in favor of #187

PR #189, opened 2023-12-14
Add abstract client to subscription module tests

AndrewPochapsky:
This PR adds the abstract client to the subscription module tests. A large part of this change involved adding functionality to the client that allows it to deploy cw20 contracts using a builder interface. This makes it more intuitive for developers and limits the number of dependencies that they need to take on (cw20 logic is distributed among 3 different packages!). This logic is only present for testing.

### Checklist

- [x] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-client/Cargo.toml
```diff
@@ -13,7 +13,7 @@ keywords.workspace = true
 
 [features]
 default = ["test-utils"]
-test-utils = ["cw-asset"]
+test-utils = ["cw-asset", "cw20", "cw20-base", "cw-plus-interface"]
```
CyberHoward, line 16: I thought optional deps require a `deps:` prefix? so `"deps:cw20", "deps:cw20-base", ...` 
@Buckram123 do you know? 

I think adding them this way also exposes them as their own features. 

So `features = ["cw20"]` is something you can do when importing `abstract-client`, but again could be wrong!

framework/packages/abstract-client/tests/integration.rs
```diff
@@ -359,3 +363,110 @@ fn can_install_module_with_dependencies() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn can_build_cw20_with_all_options() -> anyhow::Result<()> {
```
CyberHoward, line 368: very nice!

framework/packages/abstract-client/src/client.rs
```diff
@@ -57,10 +57,18 @@ impl<Chain: CwEnv> AbstractClient<Chain> {
     ) -> AbstractClientResult<Account<Chain>> {
         Account::from_namespace(&self.abstr, namespace)
     }
+
+    pub fn chain(&self) -> Chain {
```
CyberHoward, line 61: We originally also had `chain` as the function name instead of `environment` but rust-std exports a trait with a `chain` function so it caused weird conflicts in some places, which is why I renamed it to `environment`. 

I think it's important to be consistent tho, so id prefer keeping this exposed in the `Infrastructure` trait exclusively.

We could rename the `environment` function to `env` or something else?

CyberHoward (approved):

framework/packages/abstract-client/src/client.rs
```diff
@@ -57,10 +57,18 @@ impl<Chain: CwEnv> AbstractClient<Chain> {
...
+    }
+
+    pub fn sender(&self) -> Addr {
+        self.environment().sender()
+    }
 }
 
 #[cfg(feature = "test-utils")]
-mod test_utils {
+pub mod test_utils {
```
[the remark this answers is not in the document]
CyberHoward: I'm used to viewing big PR changes. I almost never use the web-ui but pull the branch into vs-code and view the diff/files there. Makes it way easier to review :)

framework/packages/abstract-client/src/client.rs
```diff
@@ -97,6 +120,94 @@ mod test_utils {
...
+        use cw_orch::prelude::{CwOrchInstantiate, CwOrchUpload, Mock};
+        use cw_plus_interface::cw20_base::InstantiateMsg;
+
+        use crate::client::AbstractClientResult;
+
+        pub struct Cw20Builder {
+            chain: Mock,
+            name: String,
+            symbol: String,
+            decimals: u8,
```
Buckram123, line 140: I guess decimals could have default value as well

framework/packages/abstract-client/src/client.rs
```diff
@@ -97,6 +120,94 @@ mod test_utils {
...
+                self.admin = Some(Addr::unchecked(admin.into()));
+                self
+            }
+
+            pub fn instantiate_with_id(&self, id: &str) -> AbstractClientResult<Cw20Base<Mock>> {
+                let cw20 = Cw20Base::new(id, self.chain.clone());
+
+                // TODO: Consider adding error if the code-id is already uploaded. This would
+                // imply that the user is trying to instantiate twice using the same id which would
+                // overwrite the state.
```
Buckram123, line 192: What do you think about removing id as a field and generating it as "cw20-{symbol}" instead?

modules/contracts/apps/subscription/tests/integration.rs
```diff
@@ -29,129 +28,79 @@ const DENOM: &str = "abstr";
...
-        Some(&sender),
-        None,
-    )?;
-    Ok(emission_cw20)
+#[allow(dead_code)]
+struct Cw20Subscription {
+    client: AbstractClient<Mock>,
+    subscription_app: Application<Mock, SubscriptionInterface<Mock>>,
+    payment_asset: AssetInfo,
+}
```
Buckram123, line 43: Nice!

modules/contracts/apps/subscription/src/msg.rs
```diff
@@ -90,8 +90,8 @@ pub enum SubscriptionExecuteMsg {
         subscription_cost_per_second: Option<Decimal>,
         /// Subscription emissions per second
         subscription_per_second_emissions: Option<EmissionType<String>>,
-        /// New unsubscription hook addr
-        unsubscription_hook_addr: Option<Clearable<String>>,
+        /// New unsubscribe_hook_addr hook addr
+        unsubscribe_hook_addr: Option<Clearable<String>>,
```
Buckram123, line 94: I think comment a bit off now

Buckram123 (approved):
LGTM

framework/packages/abstract-client/src/client.rs
Buckram123: So it will simply become `instantiate` with consuming self(not sure why it's reference rn)

framework/packages/abstract-client/src/client.rs
```diff
@@ -97,6 +120,94 @@ mod test_utils {
...
+
+        use cosmwasm_std::Addr;
+
+        use cw_orch::prelude::{CwOrchInstantiate, CwOrchUpload, Mock};
+        use cw_plus_interface::cw20_base::InstantiateMsg;
+
+        use crate::client::AbstractClientResult;
+
+        pub struct Cw20Builder {
+            chain: Mock,
```
Buckram123, line 137:
```suggestion
        pub struct Cw20Builder<Chain: CwEnv> {
            chain: Chain,
```
Can't it take any CwEnv chain?

framework/packages/abstract-client/src/client.rs
AndrewPochapsky: I'd rather have the `Infrastructure` trait remain internal and only expose what we want through the client. Happy to change the name of this function though.

framework/packages/abstract-client/src/client.rs
AndrewPochapsky: I like having it be a ref to allow creating multiple contracts with similar/same configurations. Plus it follows best practice of a builder to use ref :)

framework/packages/abstract-client/src/client.rs
AndrewPochapsky: True, not 100% if we want to expose this in a non-test environment though.

framework/packages/abstract-client/Cargo.toml
AndrewPochapsky: I don't think we should have partial features like that since just a dep on `cw20` would not be sufficient for using the `cw20_builder`. But having a feature for the `cw20_builder` as a whole is a good idea.

framework/packages/abstract-client/src/client.rs
CyberHoward: In that case, where are you using this function? If you want to don't want to expose `Infrastructure` then you probably don't want to expose anything chain-related? Is that feasible?

AndrewPochapsky: The trait also has other stuff like returning an instance of `Abstract` which we don't want to do. I added `chain()` here to be used in the subscription module tests for raw queries. I tried to allow a `QuerierWrapper` to be returned by the client but with the `Rc<RefCell<...>>` stuff it didn't seem possible.

AndrewPochapsky: Maybe we just need to expose `chain` in `Mock` environments then?

CyberHoward: Raw queries is a feature that's been requested but not yet implemented: https://linear.app/abstract-sdk/issue/ORC-61/add-raw-query-support

Maybe we can split the `Infrastructure` trait into `AbstractInfrastructure` and `Environment` or whatever where 

```rust
trait AbstractInfrastructure<Chain> : Environment<Chain> {
   // construct Abstract<Chain>
}

trait Environment<Chain> {
 fn chain(&self) -> Chain
}
```

Because getting the environment is something that I've had to do quite a lot (hence there's a trait that provides it).

AndrewPochapsky: Sure, I can look at adding that in a future change.

framework/packages/abstract-client/src/client.rs
AndrewPochapsky: Sure, I don't mind making it available at the `CwEnv` level if there is consensus, the one thing I am not sure about is usually chains already have a cw20 code-id present right? Would it be a unusual to `upload` a new one as we do here?

framework/packages/abstract-client/src/client.rs
AndrewPochapsky: Which default would make sense do you think?

CyberHoward: 6 is the default value for Cosmos coins (and cw20s afaik)

framework/packages/abstract-client/src/client.rs
CyberHoward: It could be. It could use the abstract version control to check if it's registered and use that code-id. But i'm not sure that's something that makes sense to include here...

AndrewPochapsky:
Going to merge to make it easier to put up another PR, @Buckram123 please read the questions in the comments you posed and connect with me offline :)

PR #190, opened 2023-12-16
Add MutCwEnv to client + other improvements

AndrewPochapsky:
- Splits out `environment` from `Infrastructure` trait into its own public trait `Environment`. We still want to avoid requiring devs to use this trait by creating proxies through the client but it should be there for more advanced usage if necessary.
- Removes `Environment` impl for `AccountBuilder` to avoid making it public (would be weird imo). We only used this once and I just replaced that line with what the impl used to be.
- Implement the `MutCwEnv` methods on both `AbstractClient` and `Account` to allow setting balances. Also added the query counterparts to allow retrieving balances. Doing this directly on an `Account` object feels great.
- Moves some `Mock` only impls in Client to the general case (I didn't realize at first that this was available for all `CwEnv`!)
- Moves `test_utils` module in the client to its own file for organizational purposes.

### Checklist

- [x] CI is green.
- [x] Changelog updated.


CyberHoward:
This is great! 

We already bumped cw-orch on the release branch (https://github.com/AbstractSDK/abstract/pull/176) as the migration testing branch was depending on this branch's new version and the migration branch depended on some new features that were only stabilized in the new release. 

imo it makes sense to merge the release branch into main right now and then address this PR.

CyberHoward:
@AndrewPochapsky this can be update now

PR #193, opened 2023-12-18
Only allow for incremental migrations of modules.

Buckram123:
### Checklist

- [x] CI is green.
- [x] Changelog updated.


CyberHoward (changes requested):
Can we assert this on manager and proxy as well?

framework/packages/abstract-core/src/objects/module_version.rs
```diff
@@ -107,6 +107,28 @@ pub fn assert_contract_upgrade(
...
+    ensure!(
+        no_skips,
+        AbstractError::CannotSkipVersion {
+            contract,
+            from: from_version,
+            to: to_version,
+        }
+    );
     Ok(())
 }
```
Buckram123, line 133: @CyberHoward this function runs on every native module as well

CyberHoward (approved):

PR #195, opened 2023-12-19
[CI] Move modules tests to circle ci

Buckram123:
Fixes ABS-241
### Checklist

- [x] CI is green.
- [x] Changelog updated.


.circleci/config.yml
```diff
@@ -119,6 +126,33 @@ jobs:
...
+            cd modules
+            cargo generate-lockfile
+      - restore_cache:
+          keys:
+            - cargocache-v2-tests-rust:1.72.0-{{ checksum "modules/Cargo.lock" }}
+      - run:
+          name: Modules tests
+          command: |
+            cd modules
+            cargo test --locked --all-targets -j 2 # Manually enable features if required, osmosis test tube requires golang to function
```
Buckram123, line 147: it will take eternity with 2 compilation jobs or OOM if we pass more

Buckram123: @CyberHoward 11 mins for tests and another 10 mins for saving cache. Is it fine?

CyberHoward: OMM?

Buckram123: > OMM?

Out of memory

Buckram123: > Also can't we use nextest?

We can, but why? We will have to install it first, which most likely will take more time than it will save

Buckram123: Unless it's cached job and ran N(pretty sure 2 digit) times. Tests itself not slow but compilation of modules is

CyberHoward: > > Also can't we use nextest?
> 
> We can, but why? We will have to install it first, which most likely will take more time than it will save

Because nextest is way faster. Can use pre-built bin: https://nexte.st/book/pre-built-binaries#linux-x86_64

scripts/modules-tests.sh
```diff
@@ -0,0 +1,38 @@
+# go into the directory we want to compile
+cd ./modules
+
+# Create lock file if it does not exist
+if [ ! -f Cargo.lock ]; then
+  cargo generate-lockfile
+fi
+
+# Force recompile osmosis test tube so build.rs is runned
+cargo clean -p osmosis-test-tube
```
Buckram123, line 10: Maybe it's possible to instead cache `.so` files generated by `osmosis-test-tube`, but I'm done with CI for now

CyberHoward (approved):

PR #198, opened 2023-12-22
[ABS-250] Add module getter to Application

AndrewPochapsky:
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-client/src/application.rs
```diff
@@ -34,4 +40,22 @@ impl<Chain: CwEnv, M> Application<Chain, M> {
     pub fn account(&self) -> &Account<Chain> {
         &self.account
     }
+
+    /// Attempts to get a module on the application. This would typically be a dependency of the
+    /// module of type `M`.
+    pub fn get_module<T: RegisteredModule + From<Contract<Chain>>>(
+        &self,
+    ) -> AbstractClientResult<T> {
```
Buckram123, line 48: Let's stick to one style and remove `get_` prefix from get_module

framework/packages/abstract-client/src/application.rs
```diff
@@ -34,4 +40,22 @@ impl<Chain: CwEnv, M> Application<Chain, M> {
...
+        let module_id = T::module_id();
+        let module_info: Option<ManagerModuleInfo> =
+            self.account.abstr_account.manager.module_info(module_id)?;
+        if module_info.is_none() {
+            return Err(AbstractClientError::ModuleNotInstalled {});
+        }
+        let contract = Contract::new(module_id.to_owned(), self.account.environment())
+            .with_address(Some(&module_info.unwrap().address));
```
Buckram123, line 56: I would prefer `let else` for module info here instead of double match (is_none + unwrap)

Buckram123 (approved):
Few small nits. Cool feature!

framework/packages/abstract-client/src/application.rs
```diff
@@ -43,19 +43,18 @@ impl<Chain: CwEnv, M> Application<Chain, M> {
...
+            let module: T = contract.into();
+            Ok(module)
+        } else {
+            Err(AbstractClientError::ModuleNotInstalled {})
         }
-        let contract = Contract::new(module_id.to_owned(), self.account.environment())
-            .with_address(Some(&module_info.unwrap().address));
-
-        let module: T = contract.into();
-        Ok(module)
```
Buckram123: Yeah that's better, but I meant for [let-else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html), not `if let`, I thought you went with `is_none` else `unwrap` to avoid extra indentation, with `let-else` this would be you'll basically keep same code, but `module_info.is_none()` -> `let Some(module_info) = module_info else` and remove `unwrap`.
Both is fine, just checking out if I didn't deliver proposal clearly

AndrewPochapsky: Ah ok, I see. I will just leave it as is for simplicity

Buckram123 (approved):

PR #199, opened 2023-12-22
Concentrated liquidity priority

Buckram123:
This PR aims to add priority to prefer concentrated liquidity pools over other pool types.
Also adds more details on error for query_pool_metadata.
### Checklist

- [x] CI is green.
- [ ] Changelog updated.


CyberHoward:
I think this is undefined behavior until we figure out how to expose the preferences to users.

PR #200, opened 2023-12-26
Namespace Unclaim Docs

CyberHoward:
Add docs about what happens when you unclaim a namespace

### DIFF
--- a/framework/docs/src/5_platform/2_version_control.md
+++ b/framework/docs/src/5_platform/2_version_control.md
@@ -2,6 +2,14 @@
 
 The Version Control contract acts as the registry for all modules and accounts within the Abstract platform. Abstract Accounts can use it to claim namespaces and register their modules. The Version Control contract allows modules to be queried by its namespace, name, and version, returning its reference which may be a code id or address.
 
+## Namespaces
+
+An account's namespace is a unique identifier that is used to provide a publishing domain for modules and a human readable name for any Abstract Account.
+
+Namespaces are claimed by an account and can be used to publish modules. Namespaces are unique and can only be claimed once. An account can only claim one namespace.
+
+When a namespace is removed from an account, any modules published under that namespace will be removed from the registry. This is to prevent malicious actors from registering modules under trusted namespaces.
+
 ## Propose Modules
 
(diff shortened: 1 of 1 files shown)


framework/docs/src/5_platform/2_version_control.md
```diff
@@ -2,6 +2,14 @@
 
 The Version Control contract acts as the registry for all modules and accounts within the Abstract platform. Abstract Accounts can use it to claim namespaces and register their modules. The Version Control contract allows modules to be queried by its namespace, name, and version, returning its reference which may be a code id or address.
 
+## Namespaces
+
+An account's namespace is a unique identifier that is used to provide a publishing domain for modules and a human readable name for any Abstract Account.
+
+Namespaces are claimed by an account and can be used to publish modules. Namespaces are unique and can only be claimed once. An account can only claim one namespace.
+
+When a namespace is removed from an account, any modules published under that namespace will be removed from the registry. This is to prevent malicious actors from registering modules under trusted namespaces.
```
Buckram123, line 11: I feel like if we want `to prevent malicious actors from registering modules under trusted namespaces` then we shouldn't allow namespace unclaims at all.

What will prevent another user from occupying the same modules under the same namespace after the namespace becomes free

CyberHoward: That's true but if we disallow unclaiming then we can't force namespace transfers, which might be required at some point.

Should we instead have a transfer function that?

Buckram123: > That's true but if we disallow unclaiming then we can't force namespace transfers, which might be required at some point.
> 
> Should we instead have a transfer function that?

We already have account transfers supported, which is more preferred for this case.

CyberHoward: True, but we got into this situation because Adair told 4T2 to change their main account because they already had 30 sub-accounts from deploying there while debugging. So they unclaimed their namespace and then complained that their module was no longer available. That's why I added these docs. 

I 100% agree that account ownership is better. And subaccounts can be renounced. However, it doesn't look like a sub-account deregisters itself when it's renounced?

PR #203, opened 2023-12-28
Properly handle renouncing accounts

Buckram123:
- Account unclaims its namespace when renounced
- Account unregister itself on it's owner account when it is a sub-account during renounce.
- Having sub accounts will prevent you from renouncing

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-client/src/account.rs
```diff
@@ -221,7 +221,7 @@ impl<Chain: CwEnv> Account<Chain> {
 
     /// Returns the owner address of the account.
     /// If the account is a sub-account, it will return the top-level owner address.
-    pub fn owner(&self) -> AbstractClientResult<Addr> {
+    pub fn owner(&self) -> AbstractClientResult<Option<Addr>> {
```
CyberHoward, line 224: I'd prefer this to err if there's no owner. Making this Option makes it more verbose to call and in 99% of cases you'll expect there to be an owner because otherwise it's useless. If user is working with an env that has renounced accounts they can catch the err.

cyberhoward (approved):

PR #206, opened 2024-01-02
Add documentation to Abstract Client

cyberhoward:
Adds doc-comments to the abstract-client crate


framework/packages/abstract-client/src/publisher.rs
```diff
@@ -131,22 +158,29 @@ impl<Chain: CwEnv> Publisher<Chain> {
...
     pub fn account(&self) -> &Account<Chain> {
         &self.account
     }
+
     // TODO: add `account_admin` fn to get the (Sub-)Account's admin.
+    // TODO: I don't get why it's called admin
+    /// Address of the manager
     pub fn admin(&self) -> AbstractClientResult<Addr> {
         self.account.manager()
     }
```
Buckram123, line 171: Should I rename it to manager?

framework/packages/abstract-client/src/publisher.rs
```diff
@@ -115,6 +140,8 @@ impl<Chain: CwEnv> Publisher<Chain> {
             .map_err(Into::into)
     }
 
+    /// Publish Abstract Adapter
+    // TODO: why it's publish app and deploy adapter, shouldn't we stick to one way?
     pub fn deploy_adapter<
```
Buckram123, line 145: Should it be `publish_adapter` or `publish_app` should become `deploy_app`?

framework/packages/abstract-client/src/client.rs
```diff
@@ -9,59 +41,93 @@ use crate::{
...
+    /// # let sender = cosmwasm_std::Addr::unchecked("sender");
+    /// # let chain = cw_orch::prelude::Mock::new(&sender);
+    /// use abstract_client::client::AbstractClient;
+    ///
+    /// let client = AbstractClient::new(chain)?;
+    /// # Ok::<(), AbstractClientError>(())
+    /// ```
     pub fn new(chain: Chain) -> AbstractClientResult<Self> {
+        // TODO: New error type for not found abstract on this chain?
         let abstr = Abstract::load_from(chain)?;
```
Buckram123, line 67: Loading abstract is completely unchecked, would we prefer to add check on `load_from` or here?

Buckram123 (commented):
@CyberHoward Marked few things to discuss while I was documenting it

Buckram123:
@CyberHoward Ready for a review, decided to remove setup macro, because it was one-liner

framework/packages/abstract-client/src/account.rs
```diff
@@ -176,32 +236,43 @@ impl<Chain: CwEnv> Account<Chain> {
...
+    /// This creates a new sub-account and installs the application on it.
+    pub fn install_app<M: ContractInstance<Chain> + InstallConfig + From<Contract<Chain>>>(
         &self,
         configuration: &M::InitMsg,
         funds: &[Coin],
     ) -> AbstractClientResult<Application<Chain, M>> {
         self.install_app_internal(vec![M::install_config(configuration)?], funds)
     }
 
+    pub fn install_adapter<
```
CyberHoward, line 263: This method doesn't really make sense IMO. 

Adapters should get installed on the current account. Not a new sub-account.

Buckram123: Hm, that's an interesting point! Should we have a way to install apps on current account as well?

CyberHoward: Or, what about this: 

We add a `install_on_sub_account` boolean attribute to the `Account` that can be set in the builder and defaults to `true`. That way it's a global setting and you can do whatever. 

Want to install adapters on sub-accounts? cool. 
Want to install apps on current account? done.
etc

framework/packages/abstract-client/src/application.rs
```diff
@@ -32,11 +38,15 @@ impl<Chain: CwEnv, M> DerefMut for Application<Chain, M> {
     }
 }
 
-impl<Chain: CwEnv, M> Application<Chain, M> {
-    pub fn new(account: Account<Chain>, module: M) -> Self {
-        Self { account, module }
+impl<Chain: CwEnv, M: RegisteredModule> Application<Chain, M> {
```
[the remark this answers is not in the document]
Buckram123: Do you want to revert it!

framework/packages/abstract-client/src/account.rs
Buckram123: We have multiple ways of obtaining account beyond builder, do we want to have ability to set it on `Account` object as well? And I think depending on a context we might want to use different defaults values
- `from_namespace` -> true
-  `install_*` -> if installed on sub account false(so you don't create another sub-account on top of it accidentally), else true
- `get_last_account` - true

CyberHoward: > We have multiple ways of obtaining account beyond builder, do we want to have ability to set it on `Account` object as well? And I think depending on a context we might want to use different defaults values
> 
> * `from_namespace` -> true
> * `install_*` -> if installed on sub account false(so you don't create another sub-account on top of it accidentally), else clone Account object
> * `get_last_account` - true

If we enable fetch_from_namespace by default then we don't need the `from_namespace` anymore.

framework/packages/abstract-client/src/application.rs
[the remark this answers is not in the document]
CyberHoward: I guess not. We'll figure out the standalone case later.

Buckram123 (approved):

PR #207, opened 2024-01-02
Add `get_last_account` to client.

cyberhoward:
Adds a function to the AbstractClient to get the last account that a signer created.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/packages/abstract-client/src/client.rs
+++ b/framework/packages/abstract-client/src/client.rs
@@ -108,6 +108,7 @@ impl<Chain: CwEnv> AbstractClient<Chain> {
 
     // Retrieve the last account created by the client.
     /// Returns `None` if no account has been created yet.
+    /// **Note**: This only returns accounts that were created with the Client. Any accounts created through the web-app will not be returned.
     pub fn get_last_account(&self) -> AbstractClientResult<Option<Account<Chain>>> {
         let addresses = self.environment().state().get_all_addresses()?;
         // Now search for all the keys that start with "abstract:manager-x" and return the one which has the highest x.
--- a/framework/packages/abstract-client/src/client.rs
+++ b/framework/packages/abstract-client/src/client.rs
@@ -169,6 +169,8 @@ pub(crate) fn is_local_manager(id: &str) -> AbstractClientResult<Option<AccountI
 
 #[cfg(test)]
 mod tests {
+    use cw_orch::{mock::Mock};
+
     use super::*;
 
     #[test]
@@ -187,5 +189,24 @@ mod tests {
     fn not_manager() {
         let result = is_local_manager("abstract:proxy-local-9");
         assert!(result.unwrap().is_none());
+    }
+
+    #[test]
+    fn last_owned_abstract_account() {
+        let sender = Addr::unchecked("sender");
+        let chain = Mock::new(&sender);
+        Abstract::deploy_on(chain.clone(), sender.to_string()).unwrap();
+
+        let client = AbstractClient::new(chain).unwrap();
+        let _acc = client.account_builder().build().unwrap();
+        let acc_2 = client.account_builder().build().unwrap();
+
+        let other_owner = Addr::unchecked("other_owner");
+        // create account with sender as sender but other owner
+        client.account_builder().governance_details(abstract_core::objects::gov_type::GovernanceDetails::Monarchy { monarch: other_owner.to_string() }).build().unwrap();
+
+        let last_account = client.get_last_account().unwrap().unwrap();
+
+        assert_eq!(acc_2.id().unwrap(), last_account.id().unwrap());
     }
 }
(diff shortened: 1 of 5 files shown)


framework/packages/abstract-client/src/client.rs
```diff
@@ -119,3 +120,56 @@ impl<Chain: MutCwEnv> AbstractClient<Chain> {
...
+            let addresses = self.environment().state().get_all_addresses()?;
+
+            // Now search for all the keys that start with "abstract:manager-x" and return the one which has the highest x.
+            let mut last_account: Option<(u32, Account<Daemon>)> = None;
+            for (id, _) in addresses {
+                if !id.starts_with(MANAGER) {
+                    continue;
+                }
+
+                let account_id_str = id.rsplitn(2, '-').next().unwrap();
```
Buckram123, line 148: Wouldn't it take only sequence here, for example
```
"abstract:manager-local-5".rsplitn(2, '-').next().unwrap() == "5" 
```
I would prefer to use `split_once` here

framework/packages/abstract-core/src/objects/account/account_id.rs
```diff
@@ -76,6 +76,43 @@ impl AccountId {
     }
 }
 
+impl TryFrom<&str> for AccountId {
+    type Error = AbstractError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        let mut val_iter = value.rsplitn(2, '-');
```
Buckram123, line 83: let's use rsplit_once instead

framework/packages/abstract-client/src/client.rs
```diff
@@ -119,3 +120,56 @@ impl<Chain: MutCwEnv> AbstractClient<Chain> {
...
+        /// Retrieve the last account created by the client.
+        /// Returns `None` if no account has been created yet.
+        pub fn get_last_account(&self) -> AbstractClientResult<Option<Account<Daemon>>> {
+            let addresses = self.environment().state().get_all_addresses()?;
+
+            // Now search for all the keys that start with "abstract:manager-x" and return the one which has the highest x.
+            let mut last_account: Option<(u32, Account<Daemon>)> = None;
+            for (id, _) in addresses {
+                if !id.starts_with(MANAGER) {
+                    continue;
```
Buckram123, line 145: Actually, can you cut it into different method, that takes `HashMap<String, Addr>` so it's unit-testable?

Buckram123 (changes requested):
Looks like a small bug, but maybe I'm wrong if it's tested

framework/packages/abstract-client/src/client.rs
CyberHoward: Ah yeah you're right!

framework/packages/abstract-client/src/client.rs
CyberHoward: yes :)

framework/packages/abstract-core/src/objects/account/account_id.rs
CyberHoward: Here's me thinking I know the `std` crate xd

framework/packages/abstract-client/src/client.rs
CyberHoward: Actually this will be hard to test because we need an actual Daemon object with an owner and we can't easily test all the different edge cases.

Buckram123: > Actually this will be hard to test because we need an actual Daemon object with an owner and we can't easily test all the different edge cases.

Hmm, is it tho? Can't that helper method can take `CwEnv` instead and test against Mock?

framework/packages/abstract-client/src/client.rs
```diff
@@ -102,6 +105,34 @@ impl<Chain: CwEnv> AbstractClient<Chain> {
             .map_err(Into::into)
             .map_err(Into::into)
     }
+
+    // Retrieve the last account created by the client.
+    /// Returns `None` if no account has been created yet.
+    pub fn get_last_account(&self) -> AbstractClientResult<Option<Account<Chain>>> {
```
Buckram123, line 111: Woud be nice to mention that it would only load accounts stored in this system

Buckram123 (approved):
LGTM!
What do you think about method that will fetch all account ids from the chain owned by the sender? Or would you consider it only something for the FE?

CyberHoward:
> LGTM! What do you think about method that will fetch all account ids from the chain owned by the sender? Or would you consider it only something for the FE?

That would be very nice to have! But we can add that in a different PR. :)
https://linear.app/abstract-sdk/issue/ABS-287/implement-an-account-fetcher-for-abstract-client

framework/packages/abstract-client/src/client.rs
CyberHoward: done!

PR #210, opened 2024-01-05
Fix the Version Control namespace query

Buckram123:
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-core/src/native/version_control.rs
```diff
@@ -300,7 +300,24 @@ pub struct ModulesListResponse {
 }
 
 #[cosmwasm_schema::cw_serde]
-pub struct NamespaceResponse {
+pub enum NamespaceResponse {
+    Claimed { info: NamespaceInfo },
```
CyberHoward, line 304: Can claimed just contain the info directly? 

Or have it as a unit struct like 

```suggestion
    Claimed(NamespaceInfo),
```

Buckram123: > Can claimed just contain the info directly?

It's useful to have another struct for "unwrapping" it
 
> Or have it as a unit struct like

Sure, I don't mind

PR #213, opened 2024-01-11
Resolver trait for cw `AnsHost<Chain>` interface

Buckram123:
Implementation of `Resolver` trait for `cw-orch` Name Service interface
### Checklist

- [x] CI is green.
- [x] Changelog updated.

### DIFF
--- a/framework/packages/abstract-interface/src/native/ans_host.rs
+++ b/framework/packages/abstract-interface/src/native/ans_host.rs
@@ -17,7 +17,7 @@ use std::env;
 pub struct AnsHost<Chain>;
 
 impl<Chain: CwEnv> AnsHost<Chain> {
-    pub fn resolve<R: Resolver<Chain>>(&self, item: &R) -> Result<R::Output, CwOrchError> {
+    pub fn resolve<R: ClientResolve<Chain>>(&self, item: &R) -> Result<R::Output, CwOrchError> {
         item.resolve(self)
     }
 }
@@ -50,7 +50,7 @@ where
     }
 }
 
-pub trait Resolver<Chain: CwEnv> {
+pub trait ClientResolve<Chain: CwEnv> {
     /// Result of resolving an entry.
     type Output;
     /// Resolve an entry into its value.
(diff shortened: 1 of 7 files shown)


framework/packages/abstract-sdk/src/ans_resolve.rs
```diff
@@ -86,24 +73,11 @@ impl Resolve for AnsAsset {
     }
 }
 
-impl Resolve for &AnsAsset {
```
CyberHoward, line 89: Why was this removed?

Buckram123: > Why was this removed?

It was not used and it's weird to implement trait for &&T

framework/packages/abstract-interface/src/native/ans_host.rs
```diff
@@ -46,206 +50,137 @@ where
...
-
-#[cfg(feature = "daemon")]
-impl AnsHost<Daemon> {
-    pub fn update_all_local(&self) -> Result<(), crate::AbstractInterfaceError> {
-        self.update_assets_local()?;
-        self.update_contracts_local()?;
-        self.update_pools_local()?;
-        Ok(())
-    }
+pub trait Resolver<Chain: CwEnv> {
```
CyberHoward, line 53: `Resolver` is an active statement. To implement `Resolver` for an `Entry` is like saying the `Entry` is the `Resolver`, which is not the case. 

Maybe we can rename it to `ClientResolve`?

CyberHoward: Also because the `Resolve` and `ClientResolve` are so similar it might be nice to try and merge the traits. 

I'm not entirely sure how you would go about doing that but you could try to move the AnsHost functions to a trait. Then implement that trait for the interface and the abstract-core object and then use the trait as a trait bound. 

```rust
trait <T: Resolver> Resolve<T> {
    type Output;

    fn resolve(&self, resolver: &T) -> Self::Output;
}
```
Then implement it with the resolver trait.

```rust
impl <T: Resolver> Resolve<T> for AssetEntry {
    // ...
    fn resolve(&self, resolver: &T) -> Self::Output {
        resolver.query_asset_entry(self)
    }
}
```

Maybe there's a more elegant way?

Buckram123: If we okay to make breaking change of resolve trait on contracts, because currently it takes two params: `QuerierWrapper` and `AnsHost` address. But for suggested change most suitable would be to use `AbstractNameServiceClient` as `Resolver`?

cyberhoward (approved):

PR #214, opened 2024-01-17
Update Abstract to work with cw-orch-authz

Kayanski:
This PR aims at integrating with cw-orch-authz and update dependencies to build the savings app.

# Changes made: 

1. Remove "~1.1" version constraints for cw-storage-plus
2. Patch cw-orch to work with cw-orch-authz (not necessary in this PR ?)
3. Add features to the abstract client to be able to execute on any module (not just the proxy)
4. ~~Abstract Client builder: register dexes on ANS~~ Moved to #236 
5. ~~The dex standard was changed in order to be able to call the adapter from an app instead of having to go through the proxy everytime.~~ Moved to #235

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-client/src/account.rs
```diff
@@ -382,6 +382,31 @@ impl<Chain: CwEnv> Account<Chain> {
...
+    pub fn execute_on_module<M>(
+        &self,
+        execute_msg: &M::ExecuteMsg,
+        funds: &[Coin],
+    ) -> AbstractClientResult<<Chain as TxHandler>::Response>
+    where
+        M: ContractInstance<Chain>
+            + ExecutableContract
+            + DependencyCreation
+            + InstallConfig
```
CyberHoward, line 395: Most of these trait bounds can be removed afaik

CyberHoward:
Could the description be updated to explain why, for example, the dex standard was changed?

framework/packages/abstract-client/src/account.rs
Kayanski: Yes !

CyberHoward:
For 3 we have this: https://github.com/AbstractSDK/abstract/blob/1bc94bc727edfa9141c5df2d6b0b0cd65341f1c6/framework/packages/abstract-client/src/application.rs#L43

Which we can update to be 

```rust
 pub fn new<T: >(account: Account<Chain>) -> AbstractClientResult<Self> {
        // Sanity check: the module must be installed on the account
        account.module_addresses(vec![M::module_id().to_string()])?;
        Ok(Self { account, module })
    }
```

So you can get the application from an account and easily execute on it. I have a stash ready with the changes but my tests are failing after running cargo update... 

I think 3 is too low-level for the client package.

Kayanski:
> For 3 we have this:
> 
> https://github.com/AbstractSDK/abstract/blob/1bc94bc727edfa9141c5df2d6b0b0cd65341f1c6/framework/packages/abstract-client/src/application.rs#L43
> 
> Which we can update to be
> 
> ```rust
>  pub fn new(account: Account<Chain>) -> AbstractClientResult<Self> {
>         // Sanity check: the module must be installed on the account
>         account.module_addresses(vec![M::module_id().to_string()])?;
>         Ok(Self { account, module })
>     }
> ```
> 
> So you can get the module from an account and easily execute on it. I have a stash ready with the changes but my tests are failing after running cargo update...
> 
> I think 3 is too low-level for the client package.

But using this you won't be executing on the module via the proxy, only directly on the module no ?

CyberHoward:
> > For 3 we have this:
> > https://github.com/AbstractSDK/abstract/blob/1bc94bc727edfa9141c5df2d6b0b0cd65341f1c6/framework/packages/abstract-client/src/application.rs#L43
> > 
> > Which we can update to be
> > ```rust
> >  pub fn new(account: Account<Chain>) -> AbstractClientResult<Self> {
> >         // Sanity check: the module must be installed on the account
> >         account.module_addresses(vec![M::module_id().to_string()])?;
> >         Ok(Self { account, module })
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So you can get the module from an account and easily execute on it. I have a stash ready with the changes but my tests are failing after running cargo update...
> > I think 3 is too low-level for the client package.
> 
> But using this you won't be executing on the module via the proxy, only directly on the module no ?

Why do you need this? We changed the module admin logic to allow for direct calls from the account owner instead of requiring it to go through the manager.

Kayanski:
> > > For 3 we have this:
> > > https://github.com/AbstractSDK/abstract/blob/1bc94bc727edfa9141c5df2d6b0b0cd65341f1c6/framework/packages/abstract-client/src/application.rs#L43
> > > 
> > > Which we can update to be
> > > ```rust
> > >  pub fn new(account: Account<Chain>) -> AbstractClientResult<Self> {
> > >         // Sanity check: the module must be installed on the account
> > >         account.module_addresses(vec![M::module_id().to_string()])?;
> > >         Ok(Self { account, module })
> > >     }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > So you can get the module from an account and easily execute on it. I have a stash ready with the changes but my tests are failing after running cargo update...
> > > I think 3 is too low-level for the client package.
> > 
> > 
> > But using this you won't be executing on the module via the proxy, only directly on the module no ?
> 
> Why do you need this? We changed the module admin logic to allow for direct calls from the account owner instead of requiring it to go through the manager.

My bad, didn't have that in mind

CyberHoward:
Moving this to draft because it needs to be split up.

integrations/osmosis-adapter/src/dex.rs
```diff
@@ -231,6 +231,7 @@ impl DexCommand for Osmosis {
 
         let token_in = Coin::try_from(offer_asset)?.to_string();
 
+        #[allow(deprecated)]
```
CyberHoward, line 234: What's the alternative now? Dangerous to keep this around for new versions.

PR #217, opened 2024-01-25
Dex fee query

Buckram123:
This PR aims to add usage fee query for the Dex Adapter
### Checklist

- [x] CI is green.
- [x] Changelog updated.


CyberHoward:
We should bump dex version for this change

cyberhoward (approved):

CyberHoward:
Let's rename `UsageFee` to `Fees` and have a `swap_fee` field in there. We can add more fee types in the future then.

cyberhoward (changes requested):
Let's rename UsageFee to Fees and have a swap_fee field in there. We can add more fee types in the future then.

Buckram123:
> Let's rename UsageFee to Fees and have a swap_fee field in there. We can add more fee types in the future then.

I'll create `DexFees` instead, `UsageFee` could find some use by other applications

framework/packages/standards/dex/src/msg.rs
```diff
@@ -145,4 +156,23 @@ pub enum DexQueryMsg {
...
+pub struct DexFees {
+    /// Fee for using swap action
+    pub swap_fee: Fee,
+    /// Address where all fees will go
+    pub recipient: Addr,
+}
+
+impl DexFees {
+    /// Usage fee for swap
+    pub fn swap_usage_fees(&self) -> AbstractResult<UsageFee> {
```
CyberHoward, line 175:
```suggestion
    pub fn swap_fees(&self) -> AbstractResult<UsageFee> {
```

cyberhoward (approved):

modules/contracts/adapters/dex/src/handlers/execute.rs
```diff
@@ -54,7 +54,7 @@ pub fn execute_handler(
             // unwrap namespace, since it's unlikely to have unclaimed abstract namespace
             let namespace_info = namespace.unwrap();
             ensure_eq!(
-                namespace_info.account_base.proxy,
+                namespace_info.account_base.manager,
```
CyberHoward, line 57: This would be weird. The manager can only call it if it has the module installed.

Can we change this to use the accounts's admin as the accepted sender? Like we do for apps?

CyberHoward:
Another safeguard we could add is to add a `MAX_FEE` of 5% or something. That way it can never be completely abused.

modules/contracts/adapters/dex/src/handlers/execute.rs
Buckram123: Sure, fixed!

Buckram123:
> Another safeguard we could add is to add a `MAX_FEE` of 5% or something. That way it can never be completely abused.

In the dex or for the `Fee` object?

CyberHoward:
> > Another safeguard we could add is to add a `MAX_FEE` of 5% or something. That way it can never be completely abused.
> 
> In the dex or for the `Fee` object?

In the dex

framework/CHANGELOG.md
```diff
@@ -9,8 +9,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Usage fee query for the dex adapter
```
CyberHoward, line 12: I think we should separate the module changelog from the framework changelog because we can version them differently going forward.

PR #218, opened 2024-01-25
Update import formatting

cyberhoward:
Runs rustfmt with `group_imports = "StdExternalCrate"` and `import_granularity = "Crate"` to make imports smaller and more consistent. 

### Checklist

- [x] CI is green.
- [x] Changelog updated.


rustfmt.toml
```diff
@@ -0,0 +1,10 @@
+# stable
+newline_style = "Unix"
+hard_tabs = false
+tab_spaces = 4
+reorder_imports = true
+reorder_modules = true
+
+# These are nightly
+# group_imports = "StdExternalCrate"
+# imports_granularity = "Crate"
```
Buckram123, line 10: Do you have to uncomment them to format?

Buckram123 (approved):
LGTM! Nice, IMO it makes them easier to process with eyes as well

rustfmt.toml
CyberHoward: Yes, because they are `nightly` only

PR #220, opened 2024-01-26
Move DependencyCreation to the abstract-client

Buckram123:
### Checklist

- [ ] CI is green.
- [x] Changelog updated.

### DIFF
--- a/framework/packages/abstract-app/Cargo.toml
+++ b/framework/packages/abstract-app/Cargo.toml
@@ -11,7 +11,7 @@ crate-type = ["cdylib", "rlib"]
 
 
 [features]
-interface-macro = ["dep:abstract-interface", "dep:cw-orch"]
+interface-macro = ["dep:abstract-interface", "dep:cw-orch", "abstract-client"]
 test-utils = [
   "dep:abstract-testing",
   "interface-macro",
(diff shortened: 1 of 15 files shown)


framework/packages/abstract-app/Cargo.toml
```diff
@@ -11,7 +11,7 @@ crate-type = ["cdylib", "rlib"]
 
 
 [features]
-interface-macro = ["dep:abstract-interface", "dep:cw-orch"]
+interface-macro = ["dep:abstract-interface", "dep:cw-orch", "abstract-client"]
```
CyberHoward, line 14: Why does this need abstract-client?

Buckram123: > Why does this need abstract-client?

Because of `DependencyCreation` trait

Buckram123: Ah yeah, should be only for `test-utils` right

CyberHoward: I think it's only needed in `test-utils`

Buckram123:
`abstract-interface` already exposed in `abstract-app` with `interface-macro` feature

PR #221, opened 2024-01-29
Added dex api for generating swap messages

Kayanski:
This PR aims at adding a simple route on the dex api for generating swap messages.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/modules/contracts/adapters/dex/src/api.rs
+++ b/modules/contracts/adapters/dex/src/api.rs
@@ -132,6 +132,30 @@ impl<'a, T: DexInterface> Dex<'a, T> {
         })?;
         Ok(response)
     }
+
+    /// Generate the raw messages that are need to run a swap
+    pub fn generate_swap_messages(
+        &self,
+        offer_asset: OfferAsset,
+        ask_asset: AssetEntry,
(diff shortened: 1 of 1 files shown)


modules/contracts/adapters/dex/src/api.rs
```diff
@@ -132,6 +132,30 @@ impl<'a, T: DexInterface> Dex<'a, T> {
...
+            message: DexExecuteMsg::Action {
+                dex: self.dex_name(),
+                action: DexAction::Swap {
+                    offer_asset,
+                    ask_asset,
+                    max_spread,
+                    belief_price,
+                },
+            },
+            proxy_addr: sender_receiver.into(),
```
Buckram123, line 155: This address is not restricted to the proxy, do we want to migrate this change from #214 here?

Kayanski: Sure but that's the interface that is provided here. We will migrate that once the change is done. Here the api indicates `sender_receiver` which corresponds to that change

Buckram123 (approved):
LGTM

PR #222, opened 2024-01-29
Update template to v0.20

CyberHoward:
Updates the Abstract Version and removes the typescript code from the template. 

The FE component of the template will be worked out independently.


Buckram123 (approved):
LGTM

PR #223, opened 2024-01-29
Authz msg wrapping to authz api in the abstract sdk

Kayanski:
This PR aims at adding a new `execute` endpoint in the authz api, that allows to execute messages using authz grants. 

By mistake, it depends on https://github.com/AbstractSDK/abstract/pull/221
### Checklist

- [x] CI is green.
- [x] Changelog updated.


CyberHoward:
@Kayanski there's a feature problem, see CI. 

I think some feature on cosmwasm-std needs to be enabled when stargate feature is enabled in sdk.

Kayanski:
> @Kayanski there's a feature problem, see CI.
> 
> I think some feature on cosmwasm-std needs to be enabled when stargate feature is enabled in sdk.

Corrected

cyberhoward (approved):

PR #226, opened 2024-01-31
Publish 0.20 + IBC changes

Kayanski:
This PR aims at merging abstract 0.20 with the changes and additional files related to IBC and updates

## Changes to the client

I added some changes to the AbstractClient --> Account that I think are useful to export here : 
  - `upgrade` to upgrade an account to the new proxy and manager versions
  - `activate_ibc` to allow the account to use IBC
  - `create_ibc_account` is self explanatory

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-client/src/account.rs
```diff
@@ -327,6 +335,22 @@ impl<Chain: CwEnv> Account<Chain> {
...
+        self.abstr_account.manager.upgrade(vec![
+            (
+                ModuleInfo::from_id("abstract:manager", ModuleVersion::Latest)?,
+                Some(to_json_binary(&abstract_core::manager::MigrateMsg {})?),
+            ),
+            (
+                ModuleInfo::from_id("abstract:proxy", ModuleVersion::Latest)?,
+                Some(to_json_binary(&abstract_core::proxy::MigrateMsg {})?),
+            ),
+        ])?;
```
Buckram123, line 350: Let's reuse names from core registry

```suggestion
            (
                ModuleInfo::from_id(abstract_core::registry::MANAGER, ModuleVersion::Latest)?,
                Some(to_json_binary(&abstract_core::manager::MigrateMsg {})?),
            ),
            (
                ModuleInfo::from_id(abstract_core::registry::PROXY, ModuleVersion::Latest)?,
                Some(to_json_binary(&abstract_core::proxy::MigrateMsg {})?),
            ),
        ])?;
```

framework/packages/abstract-client/Cargo.toml
```diff
@@ -38,6 +38,8 @@ cw20 = { package = "abstract-cw20", version = "1.2.2", optional = true, features
 cw20-base = { package = "abstract-cw20-base", version = "1.2.2", optional = true, features = [
   "interface",
 ] }
+abstract-manager = { version = "0.20.0", path = "../../contracts/account/manager" }
+abstract-proxy = { version = "0.20.0", path = "../../contracts/account/proxy" }
```
Buckram123, line 42: Pretty sure it ended up not being used
```suggestion
```

interchain/scripts/src/bin/create_remote_account.rs
```diff
@@ -0,0 +1,119 @@
...
+                    // Neutron and Juno don't have an IBC-polytone connection :/
+                    if !(src_chain.0.chain_id == NEUTRON_1.chain_id
+                        && dst_chain.0.chain_id == JUNO_1.chain_id
+                        || src_chain.0.chain_id == JUNO_1.chain_id
+                            && dst_chain.0.chain_id == NEUTRON_1.chain_id)
+                    {
+                        connect(src_chain.clone(), dst_chain.clone(), runtime.handle())?;
+                    }
+                }
+            }
```
Buckram123, line 52: 1) Can we have a single `if` here instead?
2) For neutron-juno I would like to see a separate method which will have a list of unsupported(or supported) polytone connections and do something like match-loop instead

Buckram123 (commented):
Cool new client functionalities and reusable scripts! Few small nits

interchain/scripts/src/bin/create_remote_account.rs
Kayanski: I added a helper to know which chains are connected via Polytone. Is that better ?

interchain/scripts/src/abstract_ibc.rs
```diff
@@ -48,3 +50,30 @@ pub fn abstract_ibc_connection_with<Chain: IbcQueryHandler, IBC: InterchainEnv<C
...
+
+pub fn get_polytone_deployment_id(src_chain: &ChainInfo, dst_chain: &ChainInfo) -> String {
+    format!("{}-->{}", src_chain.chain_id, dst_chain.chain_id)
+}
+pub fn has_polytone_connection(
+    src_chain: ChainInfo,
+    dst_chain: ChainInfo,
+    rt: &Handle,
+) -> anyhow::Result<bool> {
+    // We just need to verify if the polytone deployment crate has the contracts in it
```
Buckram123, line 63: What if there is no ibc channel tho?

Kayanski: What do you mean ?

Buckram123: Like Polytone can be deployed on both chains, but no connection between them, or is it not possible? Or is it covered by deployment id?

interchain/scripts/src/bin/create_remote_account.rs
Buckram123: Yep

interchain/scripts/src/abstract_ibc.rs
CyberHoward: In that case they wouldn't be in here afaik

interchain/scripts/state.json
CyberHoward, line 1: Kind of annoying that the order of these values changed. Is there a way we can keep them sorted?

PR #230, opened 2024-02-02
User defined modules salt

Buckram123:
This PR aims to add control of instantiate2 salt for the modules
### Checklist

- [x] CI is green.
- [x] Changelog updated.

### DIFF
--- a/framework/packages/abstract-core/src/native/account_factory.rs
+++ b/framework/packages/abstract-core/src/native/account_factory.rs
@@ -44,7 +44,7 @@ pub mod state {
 }
 
 use cosmwasm_schema::QueryResponses;
-use cosmwasm_std::Addr;
+use cosmwasm_std::{Addr, Binary};
 
 use crate::{
     manager::ModuleInstallConfig,
@@ -104,6 +104,8 @@ pub enum ExecuteMsg {
         namespace: Option<String>,
         // Provide list of module to install after account creation
         install_modules: Vec<ModuleInstallConfig>,
+        // Provide module salt, that will be added to instantiate2 salt of installed modules
+        module_salt: Option<Binary>,
     },
 }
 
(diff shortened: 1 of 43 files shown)


framework/packages/abstract-core/src/native/account_factory.rs
```diff
@@ -104,6 +104,8 @@ pub enum ExecuteMsg {
         namespace: Option<String>,
         // Provide list of module to install after account creation
         install_modules: Vec<ModuleInstallConfig>,
+        // Provide module salt, that will be added to instantiate2 salt of installed modules
+        module_salt: Option<Binary>,
```
CyberHoward, line 108: If also used as salt for account it should be renamed to `instantiate_salt` or similar. Preferably also renamed in the manager API.

framework/packages/abstract-interface/src/account/manager.rs
```diff
@@ -65,22 +65,31 @@ impl<Chain: CwEnv> Manager<Chain> {
     pub fn replace_api(
         &self,
         module_id: &str,
+        // this field is required if you reinstall same version
+        module_salt: Option<Binary>,
```
CyberHoward, line 69: replace_api is actually replace_adapter and adapters never need salts (as they are already instantiated)

framework/packages/abstract-interface/src/native/account_factory.rs
```diff
@@ -62,6 +63,7 @@ impl<Chain: CwEnv> AccountFactory<Chain> {
             namespace,
             base_asset,
             install_modules,
+            module_salt,
```
CyberHoward, line 66: Let's just replace all "module_salt" with "instantiate_salt"

Buckram123:
Closed in favor of #243

PR #231, opened 2024-02-05
Implement account fetcher method on AccountClient

CyberHoward:
Implements an unified method for fetching an account. Also adds Display methods to the Account structures for err_unwrapping purposes.

`client.account_from( ... )` which accepts `Namespace`, `AccountId` and `Addr` of an App.

### DIFF
--- a/framework/packages/abstract-client/src/account.rs
+++ b/framework/packages/abstract-client/src/account.rs
@@ -19,6 +19,8 @@
 //! assert_eq!(alice_account.owner()?, client.sender());
 //! # Ok::<(), AbstractClientError>(())
 //! ```
+use std::fmt::{Debug, Display};
+
 use abstract_core::{
     manager::{
         state::AccountInfo, InfoResponse, ManagerModuleInfo, ModuleAddressesResponse,
@@ -151,7 +153,7 @@ impl<'a, Chain: CwEnv> AccountBuilder<'a, Chain> {
             // Check if namespace already claimed
             if let Some(ref namespace) = self.namespace {
                 let account_from_namespace_result: Option<Account<Chain>> =
-                    Account::from_namespace(
+                    Account::maybe_from_namespace(
                         self.abstr,
                         namespace.clone(),
                         self.install_on_sub_account,
@@ -228,7 +230,7 @@ impl<Chain: CwEnv> Account<Chain> {
         }
     }
 
-    pub(crate) fn from_namespace(
+    pub(crate) fn maybe_from_namespace(
         abstr: &Abstract<Chain>,
         namespace: Namespace,
         install_on_sub_account: bool,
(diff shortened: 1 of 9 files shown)


framework/packages/abstract-client/src/account.rs
```diff
@@ -145,26 +140,87 @@ impl<'a, Chain: CwEnv> AccountBuilder<'a, Chain> {
...
+        self
+    }
+
+    /// Retrieve the account from a registered [`Namespace`].
+    ///
+    /// Is similar to the `namespace` method, but does not claim the namespace if it is not already claimed.
+    pub fn from_namespace(&mut self, namespace: Namespace) -> &mut Self {
+        self.fetch_from = Some(AccountSource::Namespace(namespace));
+        self
+    }
```
Buckram123, line 163: It feels odd to have fetch inside builder, since you won't be able to change any other information about account. I would prefer those method on `Account` itself with `TryFrom<AccountSource>`. What do you think?

Buckram123: > since you won't be able to change any other information about account.

*I mean we can change it, but we don't want to. Do we want to have `update_account` on the account itself?

CyberHoward: I think tryfrom accountsource could be a bit weird discover. 

I think we can add it but have the `from_xyz` methods on the `Account` itself instead of the builder, like you suggest.

CyberHoward: And yeah we can create an update_account that returns a new "builder" based on the old account and just changes the settings but that's for another day imo.

framework/packages/abstract-client/src/source.rs
```diff
@@ -0,0 +1,30 @@
...
+    fn from(account_id: AccountId) -> Self {
+        AccountSource::AccountId(account_id)
+    }
+}
+
+impl From<Addr> for AccountSource {
+    fn from(addr: Addr) -> Self {
+        AccountSource::App(addr)
+    }
+}
```
Buckram123, line 30: Hm, this one is debatable, because it's not clear what address is that. Should we remove `From<Addr>`, so it's used explicitly `AccountSource::App`?

Buckram123 (approved):
Nice!! Pretty sure we want to change `From<Addr>`, because I would consider it as address of the base contract and not the `App` by default.

We can keep it if we could make it work for base contracts as well

framework/packages/abstract-client/src/source.rs
CyberHoward: Yeah that makes sense! Will remove it.

CyberHoward: But then we have to expose the type

PR #233, opened 2024-02-05
Started accepting raw assets in dex adapter'

Kayanski:
This PR aims at providing a simple interface that bypasses ANS for abstract-sdk users. 

This needs to be merged after https://github.com/AbstractSDK/abstract/pull/245

### TODO
- [x] Modify the adapter contract to accept assets without ANS
- [x] Test to verify common use cases work 
- [x] Create a functional and simple API for users who have all the addresses already !
   - Use `.ans_dex(..).swap(...)` to swap with ans assets
   - Use `.dex(..).swap(...)` to swap with cw_assets
- [x] Make sure the API is rational and refactor code for simplicity
- [x] Choose a strategy for astrovault adapterand implement it

## Done
- Removed unique_pool_id from fetch data
- Added pool type fetcher for astrovault
- Added possibility to use `PoolAddress::Contract` for Kujira (to have a common api) 

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


CyberHoward:
> To know which dexes are available, we could use a dex enum instead of dex string name

Yes, but then you explicitly require the dev who depends on the contract to also update their code. The idea of it being String was that it could be hot-swapped easily.

framework/packages/standards/dex/src/msg.rs
```diff
@@ -40,14 +101,14 @@ impl adapter::AdapterQueryMsg for DexQueryMsg {}
 #[cosmwasm_schema::cw_serde]
 pub struct SimulateSwapResponse {
     /// The pool on which the swap was simulated
-    pub pool: DexAssetPairing,
+    pub pool: (AskAsset, AskAsset, DexName),
```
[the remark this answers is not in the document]
Kayanski: Ok, but do we return AnsEntry or AssetInfo then in that case ? That's the whole question here

framework/packages/standards/dex/src/msg.rs
```diff
@@ -40,14 +101,14 @@ impl adapter::AdapterQueryMsg for DexQueryMsg {}
...
     /// Amount you would receive when performing the swap.
     pub return_amount: Uint128,
     /// Spread in ask_asset for this swap
     pub spread_amount: Uint128,
     // LP/protocol fees could be withheld from either input or output so commission asset must be included.
     /// Commission charged for the swap
-    pub commission: (AssetEntry, Uint128),
+    pub commission: OfferAsset,
```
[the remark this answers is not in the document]
Kayanski: Same here

CyberHoward:
What I don't like about this approach is that it can easily be mis-used. 

You can combine ANS and Raw in the same API call seemingly without issues but then you'll run into issues when you run it. 
Same with the pool_address that you can optionally provide. Nothing stops a dev from not providing it when using Raw. I think that's a bad design. 

Instead maybe we could do something like: 

```rust
pub enum DexAction {
         Swap(SwapAction<AnsAsset, AssetEntry>), 
         RawSwap(SwapAction<Asset, AssetInfo>),
}
```

And then have specific constructors for each: 

```rust
impl SwapAction<AnsAsset, AssetEntry> {
    fn new(...) -> Self;
}

impl SwapAction<Asset, AssetInfo> {
    fn new(...) -> Self;
}
```

So they can't be wrongly constructed. Thoughts?

Kayanski:
> What I don't like about this approach is that it can easily be mis-used.
> 
> You can combine ANS and Raw in the same API call seemingly without issues but then you'll run into issues when you run it. Same with the pool_address that you can optionally provide. Nothing stops a dev from not providing it when using Raw. I think that's a bad design.
> 
> Instead maybe we could do something like:
> 
> ```rust
> pub enum DexAction {
>          Swap(SwapAction<AnsAsset, AssetEntry>), 
>          RawSwap(SwapAction<Asset, AssetInfo>),
> }
> ```
> 
> And then have specific constructors for each:
> 
> ```rust
> impl SwapAction<AnsAsset, AssetEntry> {
>     fn new(...) -> Self;
> }
> 
> impl SwapAction<Asset, AssetInfo> {
>     fn new(...) -> Self;
> }
> ```
> 
> So they can't be wrongly constructed. Thoughts?

That's the thing, you won't run into issues by mixing ans and raw here !

framework/packages/standards/dex/src/msg.rs
CyberHoward, line 1: Can we split raw and ans into two different .rs files?

modules/contracts/adapters/dex/src/api.rs
CyberHoward, line 1: Can we have two different `mod` in here to keep the impls a bit more seperated?

framework/packages/standards/dex/src/msg.rs
Kayanski: Done --> added raw_action and ans_action files for the dex action definition and implementations

modules/contracts/adapters/dex/src/api.rs
Kayanski: Done

CyberHoward (commented):
Let's Get This Merged ;) 

We can bump versions and I'd like to see some tests that use the raw API both for testing and for dogfooding. If there's anything that we missed it should show up in tests.

PR #234, opened 2024-02-06
Abstract Client Sub account customization

Buckram123:
We have a way to create customized sub-account via AbstractClient , added a test for it.
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/packages/abstract-client/tests/integration.rs
+++ b/framework/packages/abstract-client/tests/integration.rs
@@ -802,7 +802,7 @@ fn doc_example_test() -> anyhow::Result<()> {
 
 #[test]
 fn can_get_abstract_account_from_client_account() -> anyhow::Result<()> {
-    let sender: Addr = Addr::unchecked("sender");
+    let sender: Addr = Addr::unchecked(OWNER);
     let env: Mock = Mock::new(&sender);
 
     // Build the client
@@ -811,5 +811,23 @@ fn can_get_abstract_account_from_client_account() -> anyhow::Result<()> {
     let account = client.account_builder().build()?;
     let abstract_account: &abstract_interface::AbstractAccount<Mock> = account.as_ref();
     assert_eq!(abstract_account.id()?, AccountId::local(1));
+    Ok(())
+}
+
+#[test]
+fn can_customize_sub_account() -> anyhow::Result<()> {
+    let client = AbstractClient::builder(Mock::new(&Addr::unchecked(OWNER))).build()?;
+    let account = client.account_builder().build()?;
+    let sub_account = client
+        .account_builder()
+        .name("foo-bar")
+        .ownership(GovernanceDetails::SubAccount {
+            manager: account.manager()?.to_string(),
+            proxy: account.proxy()?.to_string(),
+        })
+        .build()?;
+    
+    let info = sub_account.info()?;
+    assert_eq!(info.name, "foo-bar");
     Ok(())
 }


framework/packages/abstract-client/tests/integration.rs
```diff
@@ -813,3 +813,21 @@ fn can_get_abstract_account_from_client_account() -> anyhow::Result<()> {
...
+    let client = AbstractClient::builder(Mock::new(&Addr::unchecked(OWNER))).build()?;
+    let account = client.account_builder().build()?;
+    let sub_account = client
+        .account_builder()
+        .name("foo-bar")
+        .ownership(GovernanceDetails::SubAccount {
+            manager: account.manager()?.to_string(),
+            proxy: account.proxy()?.to_string(),
+        })
+        .build()?;
```
Buckram123, line 828: @adairrr Would you consider it good enough way and I should just add documentation on how to do it? Or would you prefer to have customization when installing app from account itself, instead of pre-defined fields?

CyberHoward: can it have `fn sub_account(&mut self, owner_account: &Account) -> &mut Self`?

Kinda weird you need to provide sub-account governance details yourself.

CyberHoward: And IMO the build should throw an err if ownership is set to sub_account is this is an action that needs to be called on the owning account, so attempting to create one manually should fail.

Buckram123: > can it have `fn sub_account(&mut self, owner_account: &Account) -> &mut Self`?
> 
> Kinda weird you need to provide sub-account governance details yourself.

Should we hide `ownership()` method then?

CyberHoward: Ah, interesting. This isn't actually enforced in our code... This means that anyone can create sub-accounts on other people's accounts I think.

CyberHoward: > > can it have `fn sub_account(&mut self, owner_account: &Account) -> &mut Self`?
> > Kinda weird you need to provide sub-account governance details yourself.
> 
> Should we hide `ownership()` method then?

I don't think so. I don't see anyone attempting to create sub-accounts this way, but should be commented that it's not possible.

framework/packages/abstract-client/Cargo.toml
```diff
@@ -46,6 +46,7 @@ cw-asset.workspace = true
 cw-controllers.workspace = true
 abstract-app = { workspace = true, features = ["test-utils"] }
 abstract-adapter = { workspace = true, features = ["test-utils"] }
+account-factory.workspace = true
```
CyberHoward, line 49: Why do we need this direct dep here?

framework/packages/abstract-client/src/account.rs
```diff
@@ -192,18 +188,23 @@ impl<'a, Chain: CwEnv> AccountBuilder<'a, Chain> {
...
-        )?;
+        let account_details = AccountDetails {
+            name,
+            description: self.description.clone(),
+            link: self.link.clone(),
+            namespace: self.namespace.as_ref().map(ToString::to_string),
+            base_asset: self.base_asset.clone(),
+            install_modules: vec![],
+        };
+        let abstract_account = if let Some(owner_account) = self.owner_account {
```
CyberHoward, line 199: 🔥

framework/packages/abstract-interface/src/account/mod.rs
```diff
@@ -190,6 +190,52 @@ impl<Chain: CwEnv> AbstractAccount<Chain> {
...
+                link,
+                base_asset,
+                namespace,
+                install_modules,
+            },
+            funds,
+        )?;
+
+        // Parse data from events
+        let acc_seq = &result.event_attr_value(ABSTRACT_EVENT_TYPE, "account_sequence")?;
```
CyberHoward, line 221: I think this functionality is implemented in some other places as well right? Might be worth pulling it into its own fn

framework/packages/abstract-client/Cargo.toml
Buckram123: It's dev-dep for checking error type :)

framework/packages/abstract-interface/src/account/mod.rs
Buckram123: Pulled it into own fn

framework/packages/abstract-client/src/account.rs
```diff
@@ -182,18 +190,23 @@ impl<'a, Chain: CwEnv> AccountBuilder<'a, Chain> {
...
+            owner_account
+                .abstr_account
+                .create_sub_account(account_details, Some(&[]))?
```
CyberHoward, line 204: We should probably add a `.funds` field in the builder as well. But lets do that in a different PR.

CyberHoward (approved):

PR #235, opened 2024-02-06
Remove manager checks in adapters `fetch_data`

Buckram123:
This PR aims to remove check for manager in `fetch_data` to allow querying `generate_message` non-proxy accounts as well as executing actions by authorized addresses. Originated from #214
### Checklist

- [x] CI is green.
- [x] Changelog updated.

### DIFF
--- a/integrations/Cargo.toml
+++ b/integrations/Cargo.toml
@@ -64,17 +64,18 @@ abstract-dex-standard = { version = "0.20.0" }
 abstract-staking-standard = { version = "0.20.0" }
 
 # TODO: REMOVE As soon as new dex-standard published
-# [patch.crates-io]
-# abstract-dex-standard = { path = "../framework/packages/standards/dex" }
-# abstract-adapter = { path = "../framework/packages/abstract-adapter" }
-# abstract-app = { path = "../framework/packages/abstract-app" }
-# abstract-interface = { path = "../framework/packages/abstract-interface" }
-# abstract-sdk = { path = "../framework/packages/abstract-sdk" }
-# abstract-testing = { path = "../framework/packages/abstract-testing" }
-# abstract-core = { path = "../framework/packages/abstract-core" }
-# abstract-macros = { path = "../framework/packages/abstract-macros" }
-# abstract-adapter-utils = { path = "../framework/packages/standards/utils" }
-# abstract-staking-standard = { path = "../framework/packages/standards/staking" }
+[patch.crates-io]
+cw-orch = { git = "https://github.com/AbstractSDK/cw-orchestrator", branch = "fix/savings-app-fixes" }
+abstract-dex-standard = { path = "../framework/packages/standards/dex" }
+abstract-adapter = { path = "../framework/packages/abstract-adapter" }
+abstract-app = { path = "../framework/packages/abstract-app" }
+abstract-interface = { path = "../framework/packages/abstract-interface" }
+abstract-sdk = { path = "../framework/packages/abstract-sdk" }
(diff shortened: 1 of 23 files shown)


integrations/Cargo.toml
```diff
@@ -64,17 +64,18 @@ abstract-dex-standard = { version = "0.20.0" }
...
-# abstract-app = { path = "../framework/packages/abstract-app" }
-# abstract-interface = { path = "../framework/packages/abstract-interface" }
-# abstract-sdk = { path = "../framework/packages/abstract-sdk" }
-# abstract-testing = { path = "../framework/packages/abstract-testing" }
-# abstract-core = { path = "../framework/packages/abstract-core" }
-# abstract-macros = { path = "../framework/packages/abstract-macros" }
-# abstract-adapter-utils = { path = "../framework/packages/standards/utils" }
-# abstract-staking-standard = { path = "../framework/packages/standards/staking" }
+[patch.crates-io]
+cw-orch = { git = "https://github.com/AbstractSDK/cw-orchestrator", branch = "fix/savings-app-fixes" }
```
CyberHoward, line 68: Do we need this cw-orch patch?

integrations/osmosis-adapter/src/dex.rs
```diff
@@ -7,7 +7,7 @@ use crate::{AVAILABLE_CHAINS, OSMOSIS};
 #[derive(Default)]
 pub struct Osmosis {
     pub version_control_contract: Option<VersionControlContract>,
-    pub local_proxy_addr: Option<Addr>,
+    pub sender: Option<Addr>,
```
CyberHoward, line 10: sender? 

I think this makes the logic confusing, same with having sender in fetch_data function. 

`sender` doesn't matter for adapters. What matters is which address the messages should be constructed for. So maybe we can rename it to `construct_for_addr` or `addr_as_sender` or similar but `sender` itself implies that it is the actual caller of this message, which it almost never is.

CyberHoward (changes requested):
Needs some renaming :)

integrations/osmosis-adapter/src/dex.rs
Buckram123: Agree, this will be a sender of the generated message! It's hard to get not confusing name for this... `action_sender`?

integrations/Cargo.toml
Buckram123: Nope, thanks

integrations/osmosis-adapter/src/dex.rs
Buckram123: renamed it to `addr_as_sender` @CyberHoward

CyberHoward:
https://github.com/AbstractSDK/abstract/pull/235/commits/fa0e58284a8d93c254e8a0dd844e3b5d438937fc#diff-61c31670b5b3bbdc5af1df6f732a39eee1fe08990baaada18c4f97d28b3168ceR198

dex trait still has sender

Buckram123:
> [fa0e582#diff-61c31670b5b3bbdc5af1df6f732a39eee1fe08990baaada18c4f97d28b3168ceR198](https://github.com/AbstractSDK/abstract/commit/fa0e58284a8d93c254e8a0dd844e3b5d438937fc#diff-61c31670b5b3bbdc5af1df6f732a39eee1fe08990baaada18c4f97d28b3168ceR198)
> 
> dex trait still has sender

Nice catch, fixed :)

CyberHoward (approved):

PR #241, opened 2024-02-07
Install applications and adapters in builder of abstractclient account

Buckram123:
This PR aims to add apps and adapters installation from `AbstractClient` builder. It also includes a way to attach funds for this action both manual and automatic(with option to provide own assert function)
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-client/src/account.rs
```diff
@@ -155,6 +165,51 @@ impl<'a, Chain: CwEnv> AccountBuilder<'a, Chain> {
...
+    /// Install an adapter on current account.
+    pub fn install_adapter<M: InstallConfig<InitMsg = Empty>>(
+        &mut self,
+    ) -> AbstractClientResult<&mut Self> {
+        self.install_modules.push(M::install_config(&Empty {})?);
+        Ok(self)
+    }
+
+    /// Install an application on current account.
+    pub fn install_app<M: InstallConfig<InitMsg = Empty>>(
```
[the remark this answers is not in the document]
Buckram123: Yes, bad copy paste from install_adapter, thanks!

framework/packages/abstract-client/src/account.rs
```diff
@@ -591,6 +687,22 @@ impl<Chain: CwEnv> Account<Chain> {
         // We install only one module
         Addr::unchecked(module_address)
     }
+
+    pub(crate) fn module<T: RegisteredModule + From<Contract<Chain>>>(
```
CyberHoward, line 691: So we only want `Application.module` and not `Account.module`?

Buckram123: > So we only want `Application.module` and not `Account.module`?

I thought that when we're using `.module` we are expecting to get dependency of an `Application`. I'm okay to expose it on `Account` as well, if that would make sense. User still can get module from an `Application` with `.module` or `Deref`

Buckram123:
> Could you add a test for the builder app_install method? I only found tests for the adapter fn

Added one!

CyberHoward (approved):

PR #243, opened 2024-02-08
Update module salt, to be more predictable

Buckram123:
This probably works, can't test it yet tho, because instantiate2 generate addresses with comas, and I don't feel like we should change our logic around that! Will open PR after update to MockBech32 of cw-orch

### Checklist

- [ ] CI is green.
- [x] Changelog updated.


Buckram123:
Closes ABS-301, ABS-300

framework/packages/abstract-core/src/objects/module.rs
```diff
@@ -477,17 +477,15 @@ impl Default for Monetization {
...
     let account_id = sha256::digest(account_id.to_string());
     let accound_id_bytes: &mut [u8] = &mut salt[0..32];
     accound_id_bytes.copy_from_slice(&account_id.as_bytes()[0..32]);
 
-    // 0..8 bytes for block height
-    let block_height_bytes = &mut salt[32..40];
-    block_height_bytes.copy_from_slice(&block_height.to_be_bytes());
-
+    // 1 bytes for nounce
+    salt[32] = nounce;
```
CyberHoward, line 488: This is kind of unpredictable as it messes with the generated hash. Maybe we can concat the u8 to the account_id string and hash that?

Buckram123: I don't get how it makes it unpredictable?

Buckram123: It's always 32 bytes for account id and 1 byte for nonce, which is equal to the amount you reinstalled the app

framework/contracts/account/manager/src/commands.rs
```diff
@@ -210,6 +227,37 @@ pub(crate) fn install_modules_internal(
     ))
 }
 
+fn find_nonce(
```
CyberHoward, line 230: Won't it be more efficient/predictable to get the nonce for each module individually instead of getting the minimal nonce for all modules?

CyberHoward: Also this fn generates the salt that you end up re-generating after this fn. Maybe it can return the salt(s) itself?

CyberHoward: Yeah I mean it's a subtle difference...

Maybe we shouldn't even allow for re-installing?

CyberHoward: Because I could see some really cool applications with 100% predictable module addresses. 
As soon as we add this nonce it becomes <100% ...

Buckram123: > Yeah I mean it's a subtle difference...
> 
> Maybe we shouldn't even allow for re-installing?

implemented that!

framework/packages/abstract-core/src/objects/module.rs
```diff
@@ -476,9 +476,13 @@ impl Default for Monetization {
 /// Module Metadata String
 pub type ModuleMetadata = String;
 
+pub const SALT_POSTFIX: &[u8] = b"abstract";
 /// Generate salt helper
 pub fn generate_instantiate_salt(account_id: &AccountId) -> Binary {
-    let hash = <sha2::Sha256 as sha2::Digest>::digest(account_id.to_string());
+    let account_id_hash = <sha2::Sha256 as sha2::Digest>::digest(account_id.to_string());
+    let mut hash = account_id_hash.to_vec();
+    hash.extend(SALT_POSTFIX);
```
CyberHoward, line 484: Could we add a comment on why this is needed?

Buckram123: Depends if we keep it, haha

Buckram123: Don't think adding our organization name in salt needs explanation tho

framework/contracts/native/account-factory/src/commands.rs
```diff
@@ -70,33 +70,35 @@ pub fn execute_create_account(
...
-
-        // load the next account id
-        // if it doesn't exist then it's the first account so set it to 0.
-        let next_sequence = LOCAL_ACCOUNT_SEQUENCE.may_load(deps.storage)?.unwrap_or(0);
-
-        // Check if the caller is the owner when instantiating the abstract account
-        if next_sequence == ABSTRACT_ACCOUNT_ID.seq() {
-            cw_ownable::assert_owner(deps.storage, &info.sender)?;
+    let account_id = match account_id {
+        Some(account_id) if account_id.is_local() => {
```
CyberHoward, line 74: whatttt this syntax is crazy! Didn't know you could do this

framework/contracts/native/account-factory/src/commands.rs
```diff
@@ -136,7 +138,7 @@ pub fn execute_create_account(
         })?;
     }
 
-    let salt = generate_account_salt(&account_id);
+    let salt = module::generate_instantiate_salt(&account_id);
```
CyberHoward, line 141: can we move salt gen to its own file/mod?

weird to call module mod here

Buckram123: > can we move salt gen to its own file/mod?
> 
> weird to call module mod here

Sure, sorry I can't stop think of account contracts as modules.

framework/packages/abstract-core/src/objects/module.rs
CyberHoward: can salt be arbitrary length?

framework/packages/abstract-core/Cargo.toml
```diff
@@ -31,7 +31,7 @@ cw20 = { workspace = true }
 cw-orch = { workspace = true, optional = true }
 cw-ownable = { workspace = true }
 polytone = { workspace = true }
-sha256 = { version = "1", default-features = false }
+sha2 = { version = "0.10.8", default-features = false }
```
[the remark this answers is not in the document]
Buckram123: sha2 is smaller dependency(sha2 is part of sha256) and sha256 only outputs sha256 strings, while sha2 gives `GenericArray`, so we can get `Vec<u8>` from that

framework/packages/abstract-core/src/objects/module.rs
Buckram123: up to 64 bytes, and we're using only 40 right now. 32 for sha256(account_id) and 8 for b"abstract".
We have checks for length:
https://github.com/AbstractSDK/abstract/blob/7ea32f60664d6a56eb4ef509c5a3af6439a37370/framework/packages/abstract-core/src/objects/module.rs#L833-L838

framework/contracts/native/account-factory/src/commands.rs
Buckram123: moved!

CyberHoward (approved):

PR #245, opened 2024-02-09
Bump cw-orch and use `MockBech32`

CyberHoward:
Bumps cw-orch and uses bech32 mock because of instantiate2 use. Replaces #239.

Migrate tests get a fix here : https://github.com/AbstractSDK/abstract/pull/256

## Checklist

- [x] Finish interchain tests

- [x] Replace cw-orch patch by published version when bech32 mock is released : https://github.com/AbstractSDK/cw-orchestrator/pull/325
- In framework
- In modules
- In integrations
- In interchain

- [x] Replace cw-orch-interchain by tagged version when bech32 fix is released : https://github.com/AbstractSDK/cw-orch-interchain/pull/28
- In interchain

PR #250, opened 2024-02-13
Fix/adapter ibc

Kayanski:
This PR allows to receive callbacks on abstract adapters.

First merge: 

- [x] https://github.com/AbstractSDK/abstract/pull/245
- [x] https://github.com/AbstractSDK/cw-orch-interchain/pull/28 (then update deps to use version)

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Buckram123 (approved):
Nice to see ibc mock adapter!

CyberHoward:
Let's merge #245 first and then rebase this to main!

framework/packages/abstract-sdk/src/base/endpoints/ibc_callback.rs
```diff
@@ -1,10 +1,11 @@
...
+use crate::{base::Handler, features::AbstractRegistryAccess, AbstractSdkError, ModuleInterface};
+use abstract_core::objects::module_reference::ModuleReference;
+use abstract_core::{ibc::IbcResponseMsg, objects::module::ModuleInfo, IBC_CLIENT};
 use cosmwasm_std::{DepsMut, Env, MessageInfo, Response};
 
-use crate::{base::Handler, AbstractSdkError, ModuleInterface};
-
 /// Trait for a contract's IBC callback ExecuteMsg variant.
-pub trait IbcCallbackEndpoint: Handler + ModuleInterface {
+pub trait IbcCallbackEndpoint: Handler + ModuleInterface + AbstractRegistryAccess {
```
CyberHoward, line 8: If we use VC then we can remove `ModuleInterface`. Although I would argue using `ModuleInterface` is better as it ensures that the account actually has IBC enabled.

Kayanski: That doesn't work on adapters because there is no associated account (the account is found using the sender and not the contract storage). So what we can do is implement this version on adapters and keep the previous one on apps to keep this condition

PR #251, opened 2024-02-13
Improve modules not found during install error And other API errors

Buckram123:
Even though we improved errors on queries inside apps by adding API object, we still miss information about what method exactly failed. This PR aims to improve it by adding method_name inside error.

### Checklist

- [x] CI is green.
- [ ] Changelog updated.


cyberhoward (approved):
Very cool, didn't know about that crate!!

Buckram123:
> Very cool, didn't know about that crate!!

Do we want to expand errors like that for AnsHost queries as well?

CyberHoward:
> > Very cool, didn't know about that crate!!
> 
> Do we want to expand errors like that for AnsHost queries as well?

Yeah could be nice!

Buckram123:
> Yeah could be nice!

Added!

cyberhoward (approved):

PR #256, opened 2024-02-16
Fix/migration testing

Kayanski:
This PR aims at fixing migration tests. They shouldn't trigger when the version on-chain is the same as the version in the crate.

This needs to be merged after : https://github.com/AbstractSDK/abstract/pull/245

## Done

Upgraded abstract version to allow migration test to run. This is a good practice from now on to update the Abstract version from the on-chain version when merging changes

### Checklist

- [x] CI is green.
- [x] Changelog updated.


Buckram123 (approved):
LGTM

framework/packages/abstract-interface/src/migrate.rs
```diff
@@ -37,4 +43,105 @@ impl<T: CwEnv> Abstract<T> {
...
+
+        // TODO ibc client
+
+        self.version_control.approve_any_abstract_modules()?;
+
+        Ok(has_migrated)
+    }
+}
+
+fn contract_version<Chain: CwEnv, A: ContractInstance<Chain>>(
```
CyberHoward, line 135: Would be great to have this as a query on cw-orch!

Kayanski: You think ? This is cw2 specific, so not sure it belongs in cw-orch. But could be included in the cw2 interface, dunno how re-usable that could be

Buckram123: > You think ? This is cw2 specific, so not sure it belongs in cw-orch. But could be included in the cw2 interface, dunno how re-usable that could be

we can make it as sub-interface of WasmQuerier with `cw-plus` feature enabled, wdyt? And use like `chain.wasm_querier().cw2().contract_info(addr)?`

PR #259, opened 2024-02-16
Integrate module to module IBC messages

Kayanski:
This PR aims at allowing module-to-module ibc message passing via Abstract.

Depends on [fix/callbacks](https://github.com/AbstractSDK/abstract/pull/277) (Removing user callbacks)
Depends on #329 
Depends on #327 

More details here [Abstract IBC Module Interaction ADR](https://www.notion.so/abstract-money/IBC-Module-Interactions-6ef0bc2e75fe40f49e7d90f36d3d5014)

There are tests for a basic use case and for simple security issues

Closes ABS-351

## Api Additions

- Added new IBC Client endpoint for module-to-module interaction 
    ```rust
    ModuleIbcAction {
      host_chain: String,
      source_module: InstalledModuleIdentification,
      target_module: InstalledModuleIdentification,
      msg: Binary,
      callback_info: Option<CallbackInfo>,
    }
    ```

  This allows any `source_module` to send any `msg` to any other `target_module` on a remote `host_chain` and optionally awaiting a `callback` when the IBC message is executed (or failed). 

- Added a new execute msg on the ibc host, only callable by the polytone proxy : 
  ```rust
  /// Allows for remote execution from the Polytone implementation on a local module
      ModuleExecute {
          source_module: InstalledModuleIdentification,
          target_module: InstalledModuleIdentification,
          msg: Binary,
      },
  ```

- Added a `module_ibc` endpoint for modules to be able to receive module-to-module messages. This endpoint is only callable by the IBC-host on the chain. As the ibc-host is never installed on accounts, even apps need to go through version control directly to query the module reference to verify the sender. 

## Other code

- The IBC client is responsible to make sure `source_module` correspond to the message sender [Implementation](https://github.com/AbstractSDK/abstract/pull/259/files#diff-340c2e65b49d00fc81a38be7ad44feffd6166205d2f9f20f3c64773e028dcfd1R267)

- The Ibc Host propagate the message to the `target_module` by resolving it against version control and the eventual account passed

- Added IBC host to the registered Abstract modules inside version control
   (needed for apps to be able to query the ibc host address)

- Use current version instead of latest for client and host addr verification.
- Only allow same-account module interactions when account id is important for module identification (apps, standalone)
 
## Details on `InstalledModuleIdentification`

This structure is needed to be able to indentify apps and standalone (modules that have a contract address that depends on the calling account).
This is not used in the case of an adapter. 
A better impl could be the following: 
```rust
pub enum InstalledModuleIdentification{
  App{
   account_id: AccountId,
   module_info: ModuleInfo
  },
  Standalone{
   account_id: AccountId,
   module_info: ModuleInfo
  },
  Adapter{
    module_info: ModuleInfo
  }
}
```
But this feels weird, when resolving ModuleInfo, one still has to make sure the module type is respected. So not sure what to choose there

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/contracts/native/ibc-client/src/commands.rs
```diff
@@ -206,32 +244,67 @@ pub fn execute_send_packet(
...
+            source_module,
+            target_module,
+        } => {
+            // We make sure the source_module is indeed who they say they are
+            let registry = CONFIG
+                .load(deps.storage)?
+                .version_control
+                .query_module(source_module.clone(), &deps.querier)?;
+
+            if registry.reference.unwrap_addr()? != info.sender {
```
CyberHoward, line 284: So this only works for adapters? Should this be updated to also support code-ids?

framework/contracts/native/ibc-client/src/commands.rs
```diff
@@ -206,32 +244,67 @@ pub fn execute_send_packet(
...
+            send_remote_host_module_action(
+                deps.as_ref(),
+                host_chain,
+                source_module.clone(),
+                target_module.clone(),
+                msg.clone(),
+                callback_request,
+            )?
+
+            // If they are, we send a message to the module on the remote chain
```
CyberHoward, line 301: mis-placed comment

framework/contracts/native/ibc-client/src/commands.rs
```diff
@@ -206,32 +244,67 @@ pub fn execute_send_packet(
...
+                callback_request,
+            )?
+        }
+        HostAction::Module {
+            msg,
+            source_module,
+            target_module,
+        } => {
+            // We make sure the source_module is indeed who they say they are
+            let registry = CONFIG
```
CyberHoward, line 279: This limits IBC module calls to modules that are exactly the same version. Is that the behaviour we want?

CyberHoward (commented):
Could you elaborate on why there should be an extra endpoint for this? 

I thought we could just call the module from the ibc-host without an extra endpoint as the ibc-host is the owner of the IBC account so it has all the privelages it needs. 

However, this endpoint could be unique in that it should enable the developer to trust some computation on a different chain. 

I almost question if this shouldn't be its own module type because for this type of application you would even want the configuration of the instances of the module to be protected from being changed by anyone other than the module itself. 

So the notion of an "admin" would only be valid for the "hub" instantiation of the contract while the spokes are controlled by the "hub" instance.

This, and other logic/state could be useful to have in its own base package. `ibc-app`? 

It also makes the possibility to have duplex communication between these applications very interesting. Essentially this is a first step towards mesh multi-chain applications.

I think my main feedback right now is that, if we do this as a different endpoint, the permission structures that we currently have need to be re-considered. If we don't "nerf" the ability for any module to perform admin-actions on a remote account then having a seperate endpoint isn't that valuable imo. 

If we can somehow ensure that the state of a remote account can only be altered by the owner of the local account, and not by any modules installed on that account, then I think this route becomes more interesting as it allows for siloed communication that can't be tampered with by other modules installed on the account.

Would love to have a call on this to ideate it out more. Ideally with some notion doc :)

CyberHoward:
Discussion here: https://www.notion.so/abstract-money/IBC-Module-Interactions-cf839655ef2a44e58efd1323668fbeb2?pvs=4

framework/contracts/native/ibc-client/src/commands.rs
Kayanski: That's not really what is happenning here. The source and target modules can be completely different and with mismatched versions

framework/contracts/native/ibc-client/src/commands.rs
Kayanski: Updated now !

framework/contracts/native/ibc-client/src/commands.rs
Kayanski: In the current version, modules versions and considered only for address resolution

framework/packages/abstract-sdk/src/base/endpoints/ibc_callback.rs
```diff
@@ -5,7 +5,7 @@ use cosmwasm_std::{Addr, Deps, DepsMut, Env, MessageInfo, Response};
 /// Trait for a contract's IBC callback ExecuteMsg variant.
 pub trait IbcCallbackEndpoint: Handler {
     /// Queries the IBC Client address.
-    fn ibc_client(&self, deps: Deps) -> Result<Addr, Self::Error>;
+    fn ibc_client(&self, deps: Deps) -> Result<Addr, AbstractSdkError>;
```
CyberHoward, line 8: Can't we keep Self::Error here?

Kayanski: No because Self::Error is for the app developper, not the person implementing IBCCallbackEndpoint and Self::Error implements From<AbstractSdkError>, so it's better to keep AbstractSdkError here.

Kayanski: Got mixed up, yes we can !

framework/packages/abstract-sdk/src/base/handler.rs
```diff
@@ -176,6 +176,11 @@ where
         }
         None
     }
+    /// Get an ibc callback handler if it exists.
```
CyberHoward, line 179:
```suggestion
    /// Get an IBC module call handler if it exists.
```

framework/packages/abstract-sdk/src/base/endpoints/ibc_callback.rs
CyberHoward, line 1: This endpoint should be added in the documentation [here](https://docs.abstract.money/4_get_started/3_module_builder.html#handlers).

framework/packages/abstract-sdk/src/base/contract_base.rs
```diff
@@ -38,12 +38,18 @@ pub type IbcCallbackHandlerFn<Module, Error> = fn(
     Env,
     MessageInfo,
     Module,
-    CallbackId,
+    InstalledModuleIdentification,
```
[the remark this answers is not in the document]
Kayanski: Because it's already used to fetch the ibc callback function (just like for reply)

framework/packages/abstract-app/src/endpoints/module_ibc.rs
```diff
@@ -0,0 +1,39 @@
...
+        Error,
+        CustomInitMsg,
+        CustomExecMsg,
+        CustomQueryMsg,
+        CustomMigrateMsg,
+        ReceiveMsg,
+        SudoMsg,
+    >
+{
+    // This can't be used because accounts don't install the ibc_host on their accounts so this is never available
```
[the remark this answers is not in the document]
Kayanski: This is now using current version instead of latest

framework/packages/abstract-adapter/src/state.rs
```diff
@@ -22,11 +22,19 @@ pub const AUTHORIZED_ADDRESSES_NAMESPACE: &str = "authorized_addresses";
 pub const MAXIMUM_AUTHORIZED_ADDRESSES: u32 = 15;
 
 pub trait ContractError:
```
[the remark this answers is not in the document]
Kayanski: For the syntax of the ibc_client and ibc_host functions inside ibc_callbacks and module_ibc to be easier.
This happens to match with the abstract-app behavior

framework/packages/abstract-adapter/src/endpoints/ibc_callback.rs
```diff
@@ -11,22 +11,9 @@ impl<Error: ContractError, CustomInitMsg, CustomExecMsg, CustomQueryMsg, Receive
...
-                    err
-                })?,
-                &deps.querier,
-            )
-            .map_err(|err| {
-                let err: AbstractError = err.into();
-                let err: AbstractSdkError = err.into();
-                err
-            })?;
+            .query_module(ModuleInfo::from_id_latest(IBC_CLIENT)?, &deps.querier)
```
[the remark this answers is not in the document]
Kayanski: This was discussed before, because for adapters, there is no associated account (remeber, the ibc client is calling this endpoint).
For the apps, we need to make sure that the ibc-client is installed so using the query from the account directly

framework/contracts/native/ibc-host/src/endpoints/packet.rs
```diff
@@ -105,3 +110,44 @@ pub fn handle_host_action(
...
+pub fn handle_host_module_action(
+    deps: DepsMut,
+    client_chain: ChainName,
+    source_module: InstalledModuleIdentification,
+    target_module: InstalledModuleIdentification,
+    msg: Binary,
+) -> HostResult {
+    // We resolve the target module
+    let vc = CONFIG.load(deps.storage)?.version_control;
+    let target_module_resolved =
```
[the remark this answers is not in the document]
Kayanski: Added a function to resolve from query

framework/packages/abstract-adapter/src/endpoints/ibc_callback.rs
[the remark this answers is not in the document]
CyberHoward: Yes, we decided to use the specific IBC contract version when resolving so we can update the implementation in the future.

framework/contracts/native/ibc-host/src/endpoints/packet.rs
```diff
@@ -1,18 +1,29 @@
...
+use cosmwasm_std::{wasm_execute, Binary, DepsMut, Empty, Env, Response};
 
 use crate::{
     account_commands::{self, receive_dispatch, receive_register, receive_send_all_back},
     contract::HostResult,
 };
 
+use abstract_core::base::ExecuteMsg as MiddlewareExecMsg;
+
+pub fn remote_account_id(remote_chain: ChainName, account_id: AccountId) -> HostResult<AccountId> {
```
CyberHoward, line 20:
```suggestion
pub fn client_to_host_account_id(remote_chain: ChainName, account_id: AccountId) -> HostResult<AccountId> {
```

framework/contracts/native/ibc-client/src/error.rs
```diff
@@ -36,6 +36,9 @@ pub enum IbcClientError {
     #[error("packages that contain internal calls are not allowed")]
     ForbiddenInternalCall {},
 
+    #[error("A non-module package can execute an ibc module call")]
```
CyberHoward, line 39: What is ment with this error message?

framework/contracts/native/ibc-client/src/ibc.rs
```diff
@@ -88,14 +88,17 @@ pub fn receive_action_callback(
...
             )
         }
-        IbcClientCallback::UserRemoteAction(callback_info) => {
-            // Here we transfer the callback back to the module that requested it
-            let callback = IbcResponseMsg {
+        IbcClientCallback::ModuleRemoteAction {
+            callback_info,
+            sender_module,
+        } => {
+            let callback = IbcCallbackMsg {
```
CyberHoward, line 95: We have access to the initator message in this function. Would it not be nice to have it included in the callback as well? 

`callback.initiator_msg`

Buckram123 (commented):
Look pretty good, few last concerns
- About preventing installed module to execute action from account. Idk if it makes sense, because installed module already can send pretty much any cosmos message from the account
- It feels like it shouldn't be enabled by default for every app and adapter, and will make unnecessary wasm bloat for most of the application, can we somehow generate module ibc handler only when it's used?

Buckram123 (approved):
LGTM. Lets battle-test it to see if we get more ideas on refining it!

Kayanski:
> Look pretty good, few last concerns
> 
>     * About preventing installed module to execute action from account. Idk if it makes sense, because installed module already can send pretty much any cosmos message from the account
> 
>     * It feels like it shouldn't be enabled by default for every app and adapter, and will make unnecessary wasm bloat for most of the application, can we somehow generate module ibc handler only when it's used?

Thanks for the review ! 

On the bloating part, the endpoint has maybe 20 lines in total if it's not used by any handler (just like ib callback). I'm not sure it's what we can call 'bloat' here

framework/contracts/native/ibc-client/src/ibc.rs
Kayanski: So we'll have a callback message as well as an initiator_msg in there, that's maybe too much ?

framework/packages/abstract-core/src/native/ibc_client.rs
```diff
@@ -100,43 +104,119 @@ pub enum ExecuteMsg {
...
+
+impl InstalledModuleIdentification {
+    pub fn addr(
+        &self,
+        deps: Deps,
+        vc: VersionControlContract,
+    ) -> Result<ModuleAddr, AbstractError> {
+        let target_module_resolved = vc.query_module(self.module_info.clone(), &deps.querier)?;
+
+        let no_account_id_error =
```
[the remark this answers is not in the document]
Kayanski: I don't understand this. Like a new variant inside AbstractError ? 
It's inside abstract-core so AbstractError makes sense

PR #262, opened 2024-02-21
Fx the modules cargo hack

Buckram123:
This PR aims to avoid cargo hacking mutually exclusive features (full-integrations) of different dexes which reduces feature combinations from to `5216` to `856`
### Checklist

- [ ] CI is green.
- [x] Changelog updated.

### DIFF
--- a/.github/workflows/modules-check.yml
+++ b/.github/workflows/modules-check.yml
@@ -94,7 +94,7 @@ jobs:
         uses: taiki-e/install-action@cargo-hack
       - name: cargo hack
         working-directory: ./modules
-        run: cargo hack --feature-powerset check --lib --tests
+        run: cargo hack check --feature-powerset --mutually-exclusive-features wynd,osmosis,astroport,bow,terraswap,astrovault --lib --tests
   msrv:
     runs-on: ubuntu-latest
     # we use a matrix here just because env can't be used in job names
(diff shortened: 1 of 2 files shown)


.github/workflows/modules-check.yml
```diff
@@ -94,7 +94,7 @@ jobs:
         uses: taiki-e/install-action@cargo-hack
       - name: cargo hack
         working-directory: ./modules
-        run: cargo hack --feature-powerset check --lib --tests
+        run: cargo hack check --feature-powerset --mutually-exclusive-features wynd,osmosis,astroport,bow,terraswap,astrovault --lib --tests
```
Buckram123, line 97: Hm `testing,osmosis-test,interface,cw-orch` features is mutually exclusive to those as well, or no? It turns it down to `144` combinations

CyberHoward: Can we make our own cargo hack sets? 

I'd say we just need to compile with
- All chains, no cw-orch
- All chains, with cw-orch
- No chains, no cw-orch
- No chains, with cw-orch

where we replace "with cw-orch" with the biggest feature that includes cw-orch (so probably osmosis-test?)

CyberHoward: Yeah that's great!

Buckram123: > Yeah that's great!

Ah it tries to use `group` features even if there's none of these features.
This flag with "include-features" didn't help:
```
--ignore-unknown-features
    Skip passing --features flag to `cargo` if that feature does not exist in the package.

    This flag can only be used together with either --features or --include-features.
```
As it still tries to include group everywhere

I'll just mutually exclude it then as originally planned, don't think it makes sense to create cargo hack command for every existing package

CyberHoward (approved):

PR #269, opened 2024-02-27
osmosis tests and fixes

Buckram123:
### Checklist

- [x] CI is green.
- [x] Changelog updated.
- [x] Providing liquidity (reverse ordering added test and fixed)
- [x] One-sided providing liquidity
- [x] Withdrawing liquidity


CyberHoward (commented):
Hmmm so we don't need to filter `token_in_maxs` in the provide liquidity fn? 

let osmo_msg: CosmosMsg = MsgJoinPool {
            sender: self.addr_as_sender.as_ref().unwrap().to_string(),
            pool_id,
            share_out_amount,
            token_in_maxs, < -- This erred on a 0coin (i think but wasn't able to confirm yet)
        }

Buckram123:
> Hmmm so we don't need to filter `token_in_maxs` in the provide liquidity fn?

You mean in the dex itself instead of integration? We can sort them, but it don't make sense if it's not native coins

Buckram123:
> let osmo_msg: CosmosMsg = MsgJoinPool { sender: self.addr_as_sender.as_ref().unwrap().to_string(), pool_id, share_out_amount, token_in_maxs, < -- This erred on a 0coin (i think but wasn't able to confirm yet) }

Ah, now I see what you mean, we do a swap in another integrations, so added that as well!

CyberHoward:
> > let osmo_msg: CosmosMsg = MsgJoinPool { sender: self.addr_as_sender.as_ref().unwrap().to_string(), pool_id, share_out_amount, token_in_maxs, < -- This erred on a 0coin (i think but wasn't able to confirm yet) }
> 
> Ah, now I see what you mean, we do a swap in another integrations, so added that as well!

We don't actually have to do a swap on it. Will save some gas if we just don't include the 0coin!

Buckram123:
> We don't actually have to do a swap on it. Will save some gas if we just don't include the 0coin!

Does it requires no swap only for osmosis or for every dex? It would be weird if we do swap in one integration and don't for another. If for every dex, I would prefer to fix it outside of this task

CyberHoward:
> > We don't actually have to do a swap on it. Will save some gas if we just don't include the 0coin!
> 
> Does it requires no swap only for osmosis or for every dex? It would be weird if we do swap in one integration and don't for another. If for every dex, I would prefer to fix it outside of this task, since it may require extra testing

Only for some of them. It's essentially the same but if we skip the swap we skip a contract load. So i'd prefer to skip when we can. 

Dev doesn't need to know about the difference.

Buckram123:
> Only for some of them. It's essentially the same but if we skip the swap we skip a contract load. So i'd prefer to skip when we can.

We can't skip swap for osmosis in current implementation. Both of these functions wouldn't work if one of deposits are 0:
https://github.com/AbstractSDK/abstract/blob/af357da0c900f07e286b9f3b8aae338652c66e98/integrations/osmosis-adapter/src/dex.rs#L389-L450

CyberHoward (approved):

PR #270, opened 2024-02-28
Generic dex and staking testing with Astroport and kujira testing

Buckram123:
This PR aims to add generic testing setup for dexes and staking adapters as well as adds tests for astroport and kujira
### Checklist

- [x] CI is green.
- [x] Changelog updated.


modules/contracts/adapters/dex/src/dex_tester.rs
```diff
@@ -0,0 +1,474 @@
...
+            DexInstantiateMsg {
+                recipient_account: 0,
+                swap_fee: Decimal::permille(3),
+            },
+            DeployStrategy::Force,
+        )?;
+
+        let lp_asset = {
+            let (pool, pool_metadata, lp_asset) = dex.create_pool()?;
+            let abstr = Abstract::load_from(abstr_deployment.environment())?;
```
CyberHoward, line 45: Are we sure we don't want to expose abstract in the client? xd

modules/contracts/adapters/dex/src/dex_tester.rs
```diff
@@ -0,0 +1,474 @@
...
+                balance.balance
+            }
+            _ => unreachable!(),
+        };
+
+        Ok(balance)
+    }
+}
+
+pub trait MockDex {
```
CyberHoward, line 460: Can we move this trait to the top?

cyberhoward (commented):
Amazing! Could we add a test for slippage assertion as well? So we know that that logic is implemented correctly.

modules/contracts/adapters/dex/src/dex_tester.rs
Buckram123: ah, I didn't need to use it though. forgot we have ans exposed, haha https://docs.rs/abstract-client/latest/abstract_client/struct.AbstractClient.html#method.name_service

Buckram123:
> Amazing! Could we add a test for slippage assertion as well? So we know that that logic is implemented correctly.

Sure! Will add kujira dex tests as well, since you claimed it as amazing :)

interchain/tests/src/clone_testing/dex/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+    app.set_sender(sender);
+
+    let abstr_deployment = AbstractClient::new(app)?;
+
+    // Migrate if needed
+    // TODO: can we expose it somehow for client?
+    {
+        let deployment = Abstract::load_from(abstr_deployment.environment())?;
+        deployment.migrate_if_version_changed()?;
+    }
```
Buckram123, line 27: @CyberHoward This one is more questionable, than me forgetting about `name_service`

CyberHoward: Yeah I think this method should stay internal for sure.

Buckram123:
> Will add kujira dex tests as well, since you claimed it as amazing :)

Was able to add only swap tests, because they use custom modules inside bow market maker, which CloneTesting doesn't support

CyberHoward:
If we place these tests in `interchain` we will need to update github actions to also run the tests there on changes in `modules`

CyberHoward:
But I don't really like that setup. I think the `interchain` dir is also kinda weird. Maybe we should split it up. 

We have: 
- Migration tests
- IBC tests
- Scripts
- Module clone tests

We could probably split that up into `scripts` and `tests` and have the different tests in different folders in the `tests` dir. 

Thoughts?

Buckram123:
> But I don't really like that setup. I think the `interchain` dir is also kinda weird. Maybe we should split it up.
> 
> We have:
> 
>     * Migration tests
> 
>     * IBC tests
> 
>     * Scripts
> 
>     * Module clone tests
> 
> 
> We could probably split that up into `scripts` and `tests` and have the different tests in different folders in the `tests` dir.
> 
> Thoughts?

We can move migration tests to the `framework` into separate package, same with module clone tests can be moved to the `modules`, and keep interchain scripts in `interchain` this way only needed tests are triggered.

Buckram123:
> If we place these tests in `interchain` we will need to update github actions to also run the tests there on changes in `modules`

Added interchain test trigger on change in modules and framework as well

framework/packages/abstract-interface/src/native/ans_host.rs
```diff
@@ -142,16 +142,18 @@ impl<Chain: CwEnv> ClientResolve<Chain> for AnsAsset {
     }
 }
 
-impl<Chain: CwEnv> ClientResolve<Chain> for AssetInfo {
```
CyberHoward, line 145: Don't we want this for both AssetInfo and AssetInfoUnchecked?

interchain/framework-clone-testing/src/common.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    let abstr_deployment = Abstract::load_from(app.clone())?;
+    Ok((abstr_deployment, app))
+}
+
+/// Sets up the CloneTesting for chain.
+/// Returns the abstract client
+pub fn load_abstr(chain: ChainInfo, sender: Addr) -> anyhow::Result<AbstractClient<CloneTesting>> {
+    let _ = env_logger::builder().is_test(true).try_init();
+    std::env::set_var("STATE_FILE", "../scripts/state.json");
```
CyberHoward, line 41: Is this still required, even with the new crate we use to set the path of the state file?

interchain/framework-clone-testing/src/common.rs
```diff
@@ -0,0 +1,63 @@
...
+        .expect("Should create a tokio runtime");
+    }
+    &RT
+}
+
+/// Sets up the CloneTesting for chain.
+/// Returns the abstract deployment and sender (=mainnet admin)
+pub fn setup(
+    chain: ChainInfo,
+    sender: &str,
```
CyberHoward, line 23: If this function only works for valid Addr, this should be Addr, not &str to prevent mis-use.

CyberHoward: You can use lazy static to create the addr at compile time.

Buckram123: > You can use lazy static to create the addr at compile time.

We use addresses of abstract admins here. We could determine it from querying one of the abstract core contract and remove this field then, would you prefer that instead?

framework/packages/abstract-interface/src/native/ans_host.rs
Buckram123: Yes

interchain/framework-clone-testing/src/common.rs
Buckram123: >We could determine it from querying one of the abstract core contract and remove this field then, would you prefer that instead?

But that would require "loading" abstract twice, because we can't edit `Chain` of `Abstract`, but I guess it's fine, because we only load local "state"

interchain/framework-clone-testing/src/common.rs
Buckram123: Shouldn't be, let's see if CI fails

Buckram123: @CyberHoward this fix is not present in the cw-orch-interchain tag version: https://github.com/AbstractSDK/cw-orchestrator/pull/333 so daemonstate doesn't creates empty state and clone test will try to read daemonstate file regardless

Buckram123: Nvm we need this line because clone-testing requires existing `state.json` file, since it opens `DaemonState` in read-only mode and fails at first read

CyberHoward:
We should have tests for the queries as well

cyberhoward (approved):

PR #277, opened 2024-03-06
Remove ibc callbacks for accounts.

Kayanski:
This PR aims at removing IBC callbacks on Abstract Accounts. 
This will be replaced in another PR by execution and queries between abstract modules (https://github.com/AbstractSDK/abstract/pull/259)

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-core/src/native/ibc_client.rs
```diff
@@ -259,20 +265,19 @@ mod tests {
...
+            .clone()
+            .into_cosmos_msg(receiver.clone())
+            .unwrap();
+
+        assert_that!(actual).is_equal_to(CosmosMsg::Wasm(cosmwasm_std::WasmMsg::Execute {
+            contract_addr: receiver,
+            // we can't test the message because the fields in it are private
+            msg: to_json_binary(&IbcCallbackMsg::IbcCallback(response_msg)).unwrap(),
+            funds: vec![],
+        }))
```
Buckram123, line 281: Could you add a test that would fail before this change?

Buckram123 (changes requested):
Looks good, but would really like to see test, where it's called by "attacker"

framework/packages/abstract-core/src/native/ibc_client.rs
Kayanski: And here it is ! 
https://github.com/AbstractSDK/abstract/pull/277/files#diff-c988e60c4037312258209e3bf737bff66c38698ab415e4fb4027cc89597df296R282

interchain/tests/src/interchain_accounts.rs
```diff
@@ -368,6 +491,20 @@ mod test {
         Ok(())
     }
 
+    fn assert_callback_sender(
```
CyberHoward, line 494:
```suggestion
    fn assert_callback_status_for_sender(
```

CyberHoward (commented):
Why would we want to support a module on another account doing an IBC callback on a module? Do you have an application that needs that? 

Can we keep IBC callbacks to same-account modules? I think that's the sensible default.

CyberHoward:
Convention over Configuration. 

Our convention is that inter-account calls (shielded calls) are the default security parameter. If we ever need to loosen those boundaries for a use-case that should be a configurable setting.

Kayanski:
> Why would we want to support a module on another account doing an IBC callback on a module? Do you have an application that needs that?
> 
> Can we keep IBC callbacks to same-account modules? I think that's the sensible default.

It's not really a question of support, it's a question of callback definiton. Our callback definition has a receiver address that can be set by the sender to anything. We can change that but to do that we need to redefine what the callback API looks like and what callback we want to allow. Remember, it's the user proxy that sends the callback

CyberHoward:
> > Why would we want to support a module on another account doing an IBC callback on a module? Do you have an application that needs that?
> > Can we keep IBC callbacks to same-account modules? I think that's the sensible default.
> 
> Remember, it's the user proxy that sends the callback

It's the ibc-client that sends it: 
```rust
IbcClientResponse::action("user_specific_callback")
                .add_message(callback.into_cosmos_msg(callback_info.receiver)?)
```

But yeah you have a point on the permissionless nature of the ibc callbacks! I'm just not sure if the AccountId is really the sender metric to use here. What if we use `ModuleInfo` as the sender identification instead and, like we're discussing in the Notion, the client asserts that this sender information is correct?

CyberHoward:
> > > > Why would we want to support a module on another account doing an IBC callback on a module? Do you have an application that needs that?
> > > > Can we keep IBC callbacks to same-account modules? I think that's the sensible default.
> > > 
> > > 
> > > Remember, it's the user proxy that sends the callback
> > 
> > 
> > It's the ibc-client that sends it:
> > ```rust
> > IbcClientResponse::action("user_specific_callback")
> >                 .add_message(callback.into_cosmos_msg(callback_info.receiver)?)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But yeah you have a point on the permissionless nature of the ibc callbacks! I'm just not sure if the AccountId is really the sender metric to use here. What if we use `ModuleInfo` as the sender identification instead and, like we're discussing in the Notion, the client asserts that this sender information is correct?
> 
> Sorry, I meant the callback is registered in the ibc-client by the proxy. Yeah but you may want to send a callback without a sending module. And would the sending module set this field themselves ? Again here, we have no way today of verifying which module is sending the message because it's the proxy that sends the message that registers the callback

Is it not safe to assume that if the sending module is registered on the account, the call is allowed to pass? I think we could assume that for now until we have direct module calls. Only case that would go wrong is with a maleficent module installed.

framework/packages/abstract-core/src/native/ibc_client.rs
```diff
@@ -106,33 +105,18 @@ pub enum ExecuteMsg {
         host_chain: String,
         // execute the custom host function
         action: HostAction,
-        // optional callback info
-        callback_info: Option<CallbackInfo>,
-    },
-    /// Allows to query something on a remote contract and act on that query result
-    /// This has to be an Execute variant for IBC queries
-    RemoteQueries {
```
CyberHoward, line 114: Why are we removing queries all together?

Kayanski: Because queries need callbacks and we agreed that callbacks can't be authentified so we can't use queries

Kayanski:
> Does this PR need to be merged before #259 ?

Yes

PR #287, opened 2024-03-12
Updated all modules readme

anistark:
and some link fixes.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


modules/contracts/apps/etf/README.md
CyberHoward, line 1: Would prefer to stick to one word for the product. Right now it's sometimes "ETF" and sometimes "Vault".

PR #288, opened 2024-03-13
Changed config field and added migration function

Kayanski:
This PR aims at adding permissions for claiming a new namespace.
By merging this PR, namespaces will only be claimable by admins on mainnet

We changed the name of the `allow_direct_module_registration_and_updates` config member of VC for `security_enabled` and added the migration function for it

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


CyberHoward:
Changing the name of the field is version-breaking. Could we split this up into a PR that only applies the assertion (using the old name). We could then apply that as a patch fix. 

Second PR would change the name of that field and be tagged breaking. We can keep it around until we have some other breaking changes lined up.

CyberHoward:
@Kayanski

framework/contracts/native/version-control/src/contract.rs
```diff
@@ -30,28 +33,13 @@ pub struct VcResponse;
...
-            {
-                // registration_fee is Option now, but was 0 in previous version
-                cfg.namespace_registration_fee = None;
-            }
-            VCResult::Ok(cfg)
-        })?;
-    }
-    // Remove old factory
-    deps.storage.remove(b"fac");
+    let old_config = Item::<OldConfig>::new(from_utf8(CONFIG.as_slice())?).load(deps.storage)?;
```
CyberHoward, line 36: What's this slice utf logic? @Kayanski

framework/packages/abstract-integration-tests/src/account_factory.rs
```diff
@@ -21,7 +21,7 @@ pub fn create_one_account_with_namespace_fee<T: MutCwEnv>(mut chain: T) -> AResu
     chain
         .set_balance(&sender, vec![namespace_fee.clone()])
         .unwrap();
-    version_control.update_config(None, None, Some(Clearable::Set(namespace_fee.clone())))?;
+    version_control.update_config(None, Some(Clearable::Set(namespace_fee.clone())), None)?;
```
Buckram123, line 24: It will shoot everywhere, should we disable alphabet sorting of fields? It will be pretty breaking change for everything in next version, but more stable later

framework/contracts/native/version-control/src/contract.rs
Kayanski: I wanted to make sure the config names were the same to avoid errors when copying the storage key

Buckram123: there's `deps.storage.get` if you want to use same key.

framework/contracts/native/version-control/src/error.rs
```diff
@@ -23,6 +25,9 @@ pub enum VCError {
     #[error("{0}")]
     Validation(#[from] ValidationError),
 
+    #[error("{0}")]
+    Utf8(#[from] Utf8Error),
+
```
Buckram123, line 30:
```suggestion
```
Don't think utf8 parse should be used for storage keys

Buckram123 (changes requested):

framework/packages/abstract-integration-tests/src/account_factory.rs
Kayanski: I think it's ok for it to shoot everywhere so we pay a little more attention to the changes

framework/contracts/native/version-control/src/migrate.rs
```diff
@@ -0,0 +1,37 @@
...
+pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> VCResult {
+    let to_version: Version = CONTRACT_VERSION.parse()?;
+
+    let old_config = CONFIG0_21.load(deps.storage)?;
+    let new_config = Config {
+        account_factory_address: old_config.account_factory_address,
+        security_disabled: old_config.allow_direct_module_registration_and_updates,
+        namespace_registration_fee: old_config.namespace_registration_fee,
+    };
+    CONFIG.save(deps.storage, &new_config)?;
```
Buckram123, line 32:
```suggestion
    };
    // No need to remove old config, because this uses same storage key
    CONFIG.save(deps.storage, &new_config)?;
```
Was curious why we don't remove old config, but then realized that it use same key. Let's add small comment about it

Buckram123 (approved):
Nice

PR #294, opened 2024-03-21
Make namespace claim permissioned

CyberHoward:
Adds an admin check to the namespace claim if the registration flag is false.

Implements the non-breaking logic of #288


Buckram123 (approved):
LGTM

PR #303, opened 2024-03-31
Added multiple messages to ibc action to allow for serial execution

Kayanski:
This PR aims at introducing the possiblity to send multiple manager messages in a single ibc packet. THis allows for sequential execution of remote IBC messages (when order matters). There is 2 parts to this PR : 
1. Allow for multiple messages inside 1 action
2. Remove the possiblity to send multiple IBCClient messages in one message on the Proxy Contract

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/contracts/native/ibc-client/src/contract.rs
+++ b/framework/contracts/native/ibc-client/src/contract.rs
@@ -485,11 +485,11 @@ mod tests {
             let msg = ExecuteMsg::RemoteAction {
                 host_chain: chain_name.to_string(),
                 action: HostAction::Dispatch {
-                    manager_msg: manager::ExecuteMsg::UpdateInfo {
+                    manager_msgs: vec![manager::ExecuteMsg::UpdateInfo {
                         name: None,
                         description: None,
                         link: None,
-                    },
+                    }],
                 },
                 callback_info: None,
             };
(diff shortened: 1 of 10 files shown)


CyberHoward:
Can't we allow for both? If I want to send funds AND execute, that should be possible in one proxy call, no?

CyberHoward:
Also what's the reason for this change? Does this block anything?

Kayanski:
> Can't we allow for both? If I want to send funds AND execute, that should be possible in one proxy call, no?

Yes that was my initial thought but then you have to specify multiple vectors in the API which could be confusing. Good documentation could remove this issue. 


There is only one comment associated with this, so it'd easily revertable.

Kayanski:
> Also what's the reason for this change? Does this block anything?

I need this change for the Cosmos Adventures cross-chain NFT project. Because I need to execute multiple messages on the cross-chain account (module installation+module interaction) in the same tx for UX purposes, I need to be able to do those Txs in the same packet. And I think it's a good design to allow for multiple messages async (just like txs include multiple messages)

CyberHoward (approved):

CyberHoward:
@Kayanski had to revert this merge because it's breaking. I want us to patch release first before we start applying breaking changes to main. 

See #316 for clone of this PR.

Kayanski:
Oups sorry

PR #307, opened 2024-04-04
Remove template from monorepo

CyberHoward:
We've stabilized to a point where it doesn't make sense for us to keep a clone of the app-template in this repo. This PR removes all the template content from the monorepo


Buckram123 (approved):
💯

PR #314, opened 2024-04-11
Re-expose important crates in abstract-adapter

CyberHoward:
Re-exposes the most important crates and features in the abstract-adapter crate similar to what is done in the abstract-app crate. We should figure out how we want to re-export the SDK. Currently it's exported as abstract_adapter::sdk in the adapter and abstarct_app::abstract_sdk in the app. This should be unified.


Kayanski:
> Re-exposes the most important crates and features in the abstract-adapter crate similar to what is done in the abstract-app crate. We should figure out how we want to re-export the SDK. Currently it's exported as abstract_adapter::sdk in the adapter and abstarct_app::abstract_sdk in the app. This should be unified.

`sdk` allows to avoid using the abstract_ prefix everywhere for imports. BUT this may not be clear to users that it's the same exact thing if imported from abstract_app or abstract_adapter. But imo they don't really need to have that in mind. 
I think `sdk` is the best way to go

Kayanski:
Added interface macro to adapter as well. This macro however needs to receive the CustomInitMsg as well. 
Haven't found a way to avoid passing it as an extra argument yet. 
Didn't see a reason why it was not implemented for adapter so went ahead and implemented it

Kayanski:
By the way, it looks like we don't have the same Cargo.toml settings ? 
https://github.com/AbstractSDK/abstract/pull/314/files#diff-bd6791a5e3b8b06cb6171060929c9a614a794817aad24e75cc85df3f85e7f8ed

CyberHoward:
> By the way, it looks like we don't have the same toml linter settings ? https://github.com/AbstractSDK/abstract/pull/314/files#diff-bd6791a5e3b8b06cb6171060929c9a614a794817aad24e75cc85df3f85e7f8ed

I use taplo with this alias:
ftoml='find . -type f -iname "*.toml" -print0 | xargs -0 taplo format'

https://taplo.tamasfe.dev/

Kayanski:
- Renamed abstract_core to std
- Renamed abstract_sdk to sdk and abstract_std to std inside app and adapter crates

modules/contracts/adapters/cw-staking/src/lib.rs
```diff
@@ -25,14 +25,14 @@ pub use msg::StakingQueryMsgFns;
 
 #[cfg(feature = "interface")]
 pub mod interface {
+    use abstract_adapter::abstract_interface::{
```
CyberHoward, line 28: Ahhh we can't re-expose this as `abstract_adapter::interface`? because that's a macro right?

Kayanski: No it's because interface is a module inside abstract adapter already

CyberHoward: I don't see that? `abstract_adapter::interface` doesn't exist yet. Anyway I don't think it's a good idea to re-export it that way anyway because it's not very intuitive.

PR #316, opened 2024-04-11
Allow multiple ibc messages to be executed on host chain

cyberhoward:
This PR aims at introducing the possiblity to send multiple manager messages in a single ibc packet. THis allows for sequential execution of remote IBC messages (when order matters). There is 2 parts to this PR : 
1. Allow for multiple messages inside 1 action
2. Remove the possiblity to send multiple IBCClient messages in one message on the Proxy Contract

Closes ABS-362

Created [ABS-400](https://linear.app/abstract-sdk/issue/ABS-400/allow-forwarding-all-the-data-from-executed-messages) to address the multiple response data issue.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Kayanski:
Added interface macro to adapter as well. This macro however needs to receive the CustomInitMsg as well. 
Haven't found a way to avoid passing it as an extra argument yet. 
Didn't see a reason why it was not implemented for adapter so went ahead and implemented it

framework/contracts/native/ibc-host/src/account_commands.rs
```diff
@@ -80,18 +80,22 @@ pub fn receive_register(
...
-    // Polytone handles all the necessary
-    Ok(Response::new()
-        .add_submessage(submsg)
-        .add_attribute("action", "receive_dispatch"))
+    // If there is only one executed message, we want to forward the data that this execution gets
+    Ok(if msgs.len() == 1 {
+        response.add_submessage(SubMsg::reply_on_success(msgs[0].clone(), RESPONSE_REPLY_ID))
+    } else {
+        response.add_messages(msgs)
+    })
```
Buckram123, line 98: I believe it can be useful on multi-action as well, shouldn't it just forward data of the last message?

Buckram123 (approved):
Nice :fire:

framework/contracts/native/ibc-host/src/account_commands.rs
Kayanski: Agreed that it can be useful. Do you think it's something the user expects ?

Kayanski: Should we do just like Polytone does, collecting all collected data ? 
https://github.com/DA0-DA0/polytone/blob/main/contracts/main/proxy/src/contract.rs#L73
We could do the same on account action !

Buckram123: > Should we do just like Polytone does, collecting all collected data ? https://github.com/DA0-DA0/polytone/blob/main/contracts/main/proxy/src/contract.rs#L73 We could do the same on account action !

Looks good

Kayanski: Ok, forwarding if there are multiple messages.
Forwarding the whole data will be in another release, creating tasks for it

Kayanski: Created [ABS-400](https://linear.app/abstract-sdk/issue/ABS-400/allow-forwarding-all-the-data-from-executed-messages) to address the multiple response data issue.

PR #319, opened 2024-04-16
Query rewards for osmosis

Buckram123:
TBD:
- [x] tests with v24+ osmosis-test-tube

### Checklist

- [x] CI is green.
- [x] Changelog updated.


integrations/osmosis-adapter/src/staking.rs
```diff
@@ -300,15 +301,62 @@ pub mod fns {
...
+#[serde(rename_all = "snake_case", crate = "::cosmwasm_schema::serde")]
+pub struct MiniGaugeByIdResponse {
+    pub gauge: Option<MiniGauge>,
+}
+
+#[cfg(feature = "full_integration")]
+#[derive(cosmwasm_schema::serde::Deserialize)]
+#[serde(rename_all = "snake_case", crate = "::cosmwasm_schema::serde")]
+pub struct MiniGauge {
+    pub coins: Vec<osmosis_std::types::cosmos::base::v1beta1::Coin>,
```
Kayanski, line 357: We're really fixing OSMOSIS API because it doesn't work ? Are they planning to update ?

Kayanski (approved):
This Osmosis Api error is really not good ...

integrations/osmosis-adapter/src/staking.rs
Kayanski: Are we sure that this is not only on OsmosisTestube ? Is that working on osmosis-1 or testnet ?

Buckram123: > Are we sure that this is not only on OsmosisTestube ? Is that working on osmosis-1 or testnet ?

Not sure if it's test tube bug, but it would work on main or test networks as it ignores unknown fields(and broken field is ignored in this case)

Kayanski: Great

PR #323, opened 2024-04-17
Fix clippy beta

Buckram123:
This PR aims at fixing beta clippy


CyberHoward (approved):
Wow, thanks sir! I attempted to fix it myself some day but aborted after 10 min because there were too many errs 😢

PR #324, opened 2024-04-17
Optimize astrovault dex for size

Buckram123:
We have issue with deploying contracts on archway, because of the size of contracts. This PR aims to optimize our archway wasm for dex adapter

### Checklist

- [x] CI is green.
- [x] Changelog updated.


modules/Cargo.toml
```diff
@@ -108,7 +108,7 @@ abstract-client = { path = "../framework/packages/abstract-client" }
 rpath = false
 lto = true
 overflow-checks = true
-opt-level = 3
```
CyberHoward, line 111: Can't we just do size optimization for the dex module specifically? We should optimize for speed in the others IMO

CyberHoward (changes requested):
We should discuss wether we want to optimize all for size or not

modules/Cargo.toml
Buckram123: > Can't we just do size optimization for the dex module specifically? We should optimize for speed in the others IMO

Yeah we can move release profile to each individual contract from the workspace

Buckram123: It's not possible to override release profile if it's specified in the workspace

CyberHoward: > > Can't we just do size optimization for the dex module specifically? We should optimize for speed in the others IMO
> 
> Yeah we can move release profile to each individual contract from the workspace

This was actually something that Oak security told us to do in their audit report. See: https://github.com/oak-security/audit-reports/blob/master/Abstract/2023-05-09%20Audit%20Report%20-%20Abstract%20Module%20Bases%2C%20Proxy%20and%20Manager%20Contracts%20v1.0.pdf

integrations/astrovault-adapter/src/dex.rs
```diff
@@ -36,59 +36,206 @@ pub const STABLE_POOL_FACTORY: &str =
...
-        if let Ok(config) = standard_config {
-            if config.factory.eq(STANDARD_POOL_FACTORY) {
-                return Ok(PoolType::ConstantProduct);
-            }
+        // We try the 3 different known factory addresses for one to match and get the factory address
+
+        #[cfg(feature = "full_integration")]
+        #[derive(cosmwasm_schema::serde::Deserialize)]
+        #[serde(rename_all = "snake_case", crate = "::cosmwasm_schema::serde")]
+        struct ConfigResponse {
```
CyberHoward, line 68: Can we put these types in a seperate mod?

Buckram123: This will be the only type we will have, do we really need separate mod for struct with 1 type?

Buckram123: moving it to separate mod in case of future use

integrations/astrovault-adapter/src/dex.rs
```diff
@@ -735,6 +684,124 @@ fn cw_asset_info_to_astrovault(
...
 
+/// Minimalistic versions of astrovault messages
+/// Currently messages not in use
+#[cfg(feature = "full_integration")]
+mod mini_astrovault {
+    use super::*;
+
+    #[derive(cosmwasm_schema::serde::Deserialize)]
+    #[serde(rename_all = "snake_case", crate = "::cosmwasm_schema::serde")]
+    // Ignoring unknown fields
```
CyberHoward, line 695: Won't this fail to deserialize if too many fields are provided?

Buckram123: No, serde ignores unknown fields unless https://serde.rs/container-attrs.html#deny_unknown_fields specified

CyberHoward (approved):

CyberHoward (approved):

PR #327, opened 2024-04-23
Add a parse_msg deserialize function to moduleibcmsg

Kayanski:
This PR aims a adding a deserialization helper for ibc module msgs.

It adds a `parse_msg` member to the `ModuleIbcMsg` struct.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Buckram123 (commented):
It's pretty weird, why do we need it? It just makes `ModuleIbcMsg` API slightly more complex, any cosmwasm developer should be capable of using `from_json` without this.

Kayanski:
> It's pretty weird, why do we need it? It just makes `ModuleIbcMsg` API slightly more complex, any cosmwasm developer should be capable of using `from_json` without this.

agreed that this is a bit overkill

CyberHoward:
Will close as unneeded for now.

PR #329, opened 2024-04-23
Added api to query wether an address is a module

Kayanski:
This PR aims at adding an API to query wether an address is registered Abstract SDK Module.

## Major Changes

We added a single query in the ModuleRegistry API

Here are the steps of the query : 
1. We query the address raw storage for the ModuleInfo config
2. We query the Version control to get the information associated with this config
3. We verify the stored version control information match the inputted address

## TODO

Clippy doesn't pass because AbstractSDkError has large differences between variants. We need to find a solution (not take Module but String instead in the WrongModule variant ?)

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-sdk/src/error.rs
```diff
@@ -65,6 +67,20 @@ pub enum AbstractSdkError {
         module_id: String,
         error: Box<AbstractError>,
     },
+
+    // Queried address is not a module
+    #[error("Queried address {addr} is not a module : {err}")]
```
Buckram123, line 72:
```suggestion
    #[error("Queried address {addr} is not a module: {err}")]
```

framework/packages/abstract-sdk/src/apis/version_registry.rs
```diff
@@ -142,4 +143,64 @@ impl<'a, T: ModuleRegistryInterface> ModuleRegistry<'a, T> {
             .query_standalone_info_raw(code_id, &self.deps.querier)
             .map_err(|error| self.wrap_query_error(error))
     }
+
+    /// Queries the Module information for an address.
+    /// This will error if the Address is not an Abstract Module (App, Adapter or Standalone)
+    pub fn module_info(&self, address: Addr) -> AbstractSdkResult<Module> {
```
Buckram123, line 149: Isn't `AccountBase` and `Native` module as well? Don't we have another name for App, Adapter or Standalone?

Buckram123 (approved):
Looks good, but I'm not sure if `Native` and `AccountBase` module types should be included as well

framework/packages/abstract-sdk/src/apis/version_registry.rs
Kayanski: Yeah that's something I thought about as well. It's not much more expensive to add them as well you are right

PR #330, opened 2024-04-23
Enabled IBC calls by default on remote accounts

Kayanski:
This PR aims at enabling IBC calls by default on remote accounts.
This is a starting point for duplex communication between remote accounts

Closes ABS-382

## Major Change

On `manager` instantiation, if the account id is remote, we install the ibc client on the account.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/contracts/account/manager/src/contract.rs
```diff
@@ -120,6 +117,11 @@ pub fn instantiate(
             .add_attribute(install_attribute.key, install_attribute.value);
     }
 
+    if msg.account_id.is_remote() {
+        let install_client_msg = install_ibc_client(deps.branch(), proxy_addr)?;
+        response = response.add_message(install_client_msg);
+    }
+
```
Buckram123, line 124: Can't we avoid extra message send in case !`install_modules.is_empty()`? Also, what would happen if user decides to specify install of the `ibc-client` for remote account. For example user wants to install lower version of the ibc-client for his remote account for whatever reason

Buckram123 (changes requested):
I feel like I'm missing something, can't this module install be added to the API instead of inside contract? For example here: `https://docs.rs/abstract-client/latest/abstract_client/struct.Account.html#method.create_ibc_account` we can check if `ibc-client` inside `install_modules` and if it's not just push latest version. But inside contract we just check that this module is installed.

Also changelog

framework/contracts/account/manager/src/contract.rs
Kayanski: You are right, let's push that inside the client API and not inside the contract.

Buckram123: > You are right, let's push that inside the client API and not inside the contract.

Do we need to check it inside contract, or it can be installed later?

Kayanski: no it doesn't have to be inside contract. For now it's inside the client and the account interface.

Buckram123 (approved):

PR #331, opened 2024-04-23
Duplex IBC for Accounts

Kayanski:
This PR aims at allowing duplex communication for abstract IBC accounts (remote can also call the local account).

## Current Status

When the IBC host receives an IBC message, it appends the sending chain name to the account trace before proceeding with executing the sent action. 

## Major changes

When the IBC host receives an IBC message, it modifies the account trace in the following way before proceeding with executing the sent action : 
- If the account_id is remote and the last trace matches the current chain, we pop the chain name
- Otherwise, we just add the sending chain to the trace

## Outstanding problem

The IBC host needs to be a manager of the resulting abstract account to be able to execute messaes on their behalf. 
--> The IBC host should be an admin of the manager of all abstract accounts that have IBC enabled.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/contracts/account/manager/src/commands.rs
+++ b/framework/contracts/account/manager/src/commands.rs
@@ -946,7 +946,7 @@ pub fn update_ibc_status(
         .add_message(proxy_callback_msg))
 }
 
-fn install_ibc_client(deps: DepsMut, proxy: Addr) -> Result<CosmosMsg, ManagerError> {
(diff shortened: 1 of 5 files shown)


CyberHoward:
Closing because of outstanding problem. Duplex module <> module IBC should account for all practical usecases.

PR #339, opened 2024-05-07
Feature flag module to module ibc in apps and adapters

Buckram123:
This PR aims to feature gate module to module ibc functionality to reduce wasm size for modules that don't want this functionality. Because most of the wasm size comes from serde generated code, module-to-module ibc related message variants had to be feature gated as well

Currently it's disabled only for astrovault dex adapter
### Checklist

- [ ] CI is green.
- [x] Changelog updated.

### DIFF
--- a/framework/packages/abstract-sdk/Cargo.toml
+++ b/framework/packages/abstract-sdk/Cargo.toml
@@ -15,7 +15,7 @@ module-ibc = ["abstract-std/module-ibc"]
 stargate = ["dep:cosmos-sdk-proto", "dep:prost-types"]
 
 # Expose MockModule for testing with other Adapters
-test-utils = ["dep:abstract-testing", "dep:cosmwasm-schema"]
+test-utils = ["dep:abstract-testing", "dep:cosmwasm-schema", "module-ibc"]
 
 [dependencies]
 cosmwasm-std = { workspace = true, features = ["iterator", "staking"] }
--- a/framework/packages/abstract-app/Cargo.toml
+++ b/framework/packages/abstract-app/Cargo.toml
@@ -36,7 +36,7 @@ abstract-std = { workspace = true }
 abstract-testing = { workspace = true, optional = true }
 cw-orch = { workspace = true }
 
-abstract-interface = { path = "../../packages/abstract-interface", version = "0.22.1" }
+abstract-interface = { version = "0.22.1", path = "../../packages/abstract-interface" }
 abstract-ibc-host = { version = "0.22.1", path = "../../contracts/native/ibc-host", default-features = false, features = [
     "module-ibc",
 ], optional = true }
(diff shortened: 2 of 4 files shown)


framework/packages/abstract-adapter/Cargo.toml
```diff
@@ -11,6 +11,13 @@ resolver = "2"
 crate-type = ["cdylib", "rlib"]
 
 [features]
+default = ["ibc"]
```
Buckram123, line 14: if we feature flag it in `abstract-adapter`, we need to feature flag it in `abstract-sdk` and `abstract-std`, to disable message variants

framework/packages/abstract-sdk/Cargo.toml
```diff
@@ -11,7 +11,7 @@ resolver = "2"
 
 # for quicker tests, cargo test --lib
 [features]
-default = []
+module-ibc = ["abstract-std/module-ibc"]
```
Buckram123, line 14: Since we feature flagged it inside `abstract-std` and `abstract-sdk` we need to add feature flag inside `ibc-host` and `ibc-client`, because messages will be different depending on this feature flag

framework/packages/standards/dex/Cargo.toml
```diff
@@ -15,7 +15,12 @@ resolver = "2"
 crate-type = ["cdylib", "rlib"]
 
 [features]
-default = ["export"]
+default = ["export", "module-ibc"]
+module-ibc = [
+    "abstract-std/module-ibc",
+    "abstract-sdk/module-ibc",
+    "abstract-adapter/ibc",
+]
```
Buckram123, line 23: because we feature flagged std we need to feature flag standards as well

modules/Cargo.toml
```diff
@@ -50,18 +50,18 @@ tokio = { version = "1.4", features = ["full"] }
...
+abstract-app = { version = "0.21.0", default-features = false }
 abstract-testing = { version = "0.21.0" }
 abstract-macros = { version = "0.21.0" }
 abstract-client = { version = "0.21.0" }
 
 abstract-adapter-utils = { version = "0.21.0" }
-abstract-dex-standard = { version = "0.21.0" }
-abstract-staking-standard = { version = "0.21.0" }
+abstract-dex-standard = { version = "0.21.0", default-features = false }
+abstract-staking-standard = { version = "0.21.0", default-features = false }
```
Buckram123, line 62: need to disable default features for everything inside workspace so contracts have control over features

modules/contracts/adapters/dex/Cargo.toml
```diff
@@ -23,6 +23,7 @@ required-features = ["schema"]
 
 [features]
 default = ["export"]
+ibc = ["abstract-adapter/ibc", "abstract-dex-standard/module-ibc"]
```
Buckram123, line 26: After all of that we finally can add feature flag to the adapters

Buckram123 (commented):
And now I'm at the step where I need to manage features for the integrations. Had to cargo-tree a lot to figure out what enables/disables feature of sub-dependencies in the progress because it's hard to keep all of the packages in the head at the same time

@CyberHoward What do you think about it? Is there a nicer solution where we can trim serde bloat?

Buckram123:
Not worth the effort

PR #340, opened 2024-05-13
FIX: different module id for different account id in abstract client

Buckram123:
Closes ORC-129

### Checklist

- [x] CI is green.
- [x] Changelog updated.


Kayanski:
Shouldn't we implement that directly on the account ? That way, we can centralize those module differences with the one for the account contracts. Does that make sense ?

framework/packages/abstract-interface/src/deployers.rs
```diff
@@ -17,6 +20,10 @@ pub trait RegisteredModule {
     fn module_id<'a>() -> &'a str;
     /// The version of the module.
     fn module_version<'a>() -> &'a str;
+    /// The id of the module for the account
```
CyberHoward, line 23:
```suggestion
    /// Create the unique contract ID for a module installed on an Account.
```

framework/packages/abstract-interface/src/deployers.rs
```diff
@@ -17,6 +20,10 @@ pub trait RegisteredModule {
...
+    fn account_module_id(account_id: &AccountId) -> String {
```
CyberHoward, line 24: Don't really like this function name... What do you think of `installed_module_contract_id`?

Problem is that we're using "id" in so many places. So we need to be increasingly specific about what kind of ID we're talking about. 

Think "contract_id" makes it clear that it's used for cw-orch.

CyberHoward:
> Shouldn't we implement that directly on the account ? That way, we can centralize those module differences with the one for the account contracts. Does that make sense ?

The problem is that we construct instances of the module in different places (to make them retrievable). 
With the account we have a simple object that encapsulates the contracts and there's really only one way of creating an instance of them. 

What we could do is update the implementation to have a central way of creating this `Contract` instance so we don't always construct it manually with this constructed ID. 

So have something like `fn module_contract_instance(account_id: AccountId) -> ContractInstance`

Buckram123:
> What we could do is update the implementation to have a central way of creating this `Contract` instance so we don't always construct it manually with this constructed ID.

I don't feel like we need to add any restrictions for `Contract`s abstract modules. The only place where it requires a fix is abstract-client, where we need smooth experience. Let's not overcomplicate this thing

cyberhoward (approved):

PR #342, opened 2024-05-14
Update/module ibc docs

Kayanski:
This PR aims at adding IBC documentation for accounts and modules. 
There are 3 parts to these docs : 
- Overview : Allows to understand what the feature allows. Good for application architects
- Integration : Allows to integrate with the feature. Good for application developers
- Maintenance: Allows to understand how the abstract framework functions. Good for abstract maintainers

For now I'm not using the SDK or Abstract-Client at all in those docs. We should find a spot for the SDK to intergate in or between those 3 parts 

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/docs/mdbook-admonish.css
CyberHoward, line 1: What happened here?

CyberHoward: nvm, due to version bump :)

framework/docs/book.toml
```diff
@@ -13,7 +13,7 @@ create-missing = false
 
 [preprocessor.admonish]
 command = "mdbook-admonish"
-assets_version = "2.0.1"    # do not edit: managed by `mdbook-admonish install`
+assets_version = "3.0.2" # do not edit: managed by `mdbook-admonish install`
```
CyberHoward, line 16: We should also update version for this in CI, or it will fail.

CyberHoward: done

cyberhoward (approved):

Buckram123 (approved):

CyberHoward:
Great job on these docs btw @Kayanski . Only thing that I think we should add are some diagrams.

PR #346, opened 2024-05-15
Fix automatic ibc client install

Kayanski:
This Pr aims at fixing the automatic installation of the IBC client onside the interface struct.
It actually allows the IBC client to be installed just like any other module on the manager contract.

The IBC client is now treated separately from all other modules in the install process. 

It also allows avoids panicking when an un-registered module type is installed on the account

## Discussion
Should we merge the install module flow with the enable IBC flow ? That could simplify the logic that we have here

### Pros
- Reduce the number of calls to the proxy to add modules (currently 2 calls, if the ibc-client is installed with other modules)
- Reduce the specificity of the ibc-client case inside install module
- IBC disable is as simple as an uninstall call

### Cons
- When just enabling IBC, you spend less gas because you're not preparing for other types of modules
- Ibc enable and disable will be separated

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/contracts/account/manager/src/commands.rs
```diff
@@ -902,14 +915,10 @@ pub fn update_suspension_status(
     Ok(response.add_abstract_attributes(vec![("is_suspended", is_suspended.to_string())]))
 }
 
-pub fn update_ibc_status(
+pub fn update_ibc_status_internal(
```
Buckram123, line 918: Can you add comment why it should only be used internally?

framework/contracts/account/manager/src/commands.rs
```diff
@@ -173,8 +186,7 @@ pub(crate) fn install_modules_internal(
 
                 Some(init_msg.unwrap())
             }
-            // TODO: do we want to support installing any other type of module here?
-            _ => unreachable!(),
+            _ => return Err(ManagerError::ModuleNotInstallable(module.info.to_string())),
```
Buckram123, line 189: Thanks for handling that

Buckram123 (commented):
Looks nice

framework/contracts/account/manager/src/commands.rs
CyberHoward: And in general the convention is to put on underscore before the name (and not include the "internal" in the name) 

So name would be `_update_ibc_status(...)`

CyberHoward:
> Should we merge the install module flow with the enable IBC flow ? That could simplify the logic that we have here

We can if we also single out disabling IBC through "uninstalling" the ibc module.

framework/packages/abstract-client/tests/integration.rs
```diff
@@ -1342,3 +1343,28 @@ fn install_same_app_on_different_accounts() -> anyhow::Result<()> {
...
+
+    let account2 = client
+        .account_builder()
+        .install_adapter::<IbcClient<MockBech32>>()?
+        .build()?;
+
+    let ibc_module_addr = account2.module_addresses(vec![IBC_CLIENT.to_owned()])?;
+    assert_eq!(ibc_module_addr.modules[0].0, IBC_CLIENT);
+    Ok(())
+}
```
Buckram123, line 1370: added small test that installs ibc_client. Should we consider it just as adapter for users?

framework/packages/abstract-interface/src/native/ibc_client.rs
```diff
@@ -30,7 +30,7 @@ impl<Chain: CwEnv> Uploadable for IbcClient<Chain> {
 }
 
 impl<Chain: CwEnv> RegisteredModule for IbcClient<Chain> {
-    type InitMsg = InstantiateMsg;
+    type InitMsg = cosmwasm_std::Empty;
```
Buckram123, line 33: had to fix this InitMsg as users don't need to provide InitMsg to install it

framework/packages/abstract-client/tests/integration.rs
Kayanski: Very nice ! But you can just use update_setting to do that !

Kayanski: Ok, yes, That's an adapter then, no init msg :)

framework/contracts/account/manager/src/commands.rs
```diff
@@ -915,55 +913,37 @@ pub fn update_suspension_status(
...
+                ModuleInfo::from_id_latest(IBC_CLIENT)?,
+                None,
+            )],
+            config.module_factory_address,
+            config.version_control_address,
+            vec![],
+        )?;
 
-        install_ibc_client(deps, proxy)?
+        install_msgs
```
Buckram123: Do we really need update ibc status if it's now just an installed module

Buckram123 (approved):
LGTM

Buckram123 (approved):

framework/packages/abstract-client/src/account.rs
```diff
@@ -549,7 +549,7 @@ impl<Chain: CwEnv> Account<Chain> {
 
     /// Set IBC status on an Account.
     pub fn set_ibc_status(&self, enabled: bool) -> AbstractClientResult<()> {
-        self.abstr_account.manager.update_settings(Some(enabled))?;
+        self.abstr_account.manager.set_ibc_status(enabled)?;
 
         Ok(())
     }
```
Buckram123, line 555: Do we need `set_ibc_status` method? I like that it's clear that it can be enabled or disabled any time on account. Should we have `toggle_adapter` instead?

Kayanski: You feel like `set_ibc_status` doesn't convey that it can be enabled and disabled ?

Kayanski: Yes that's a good thought !

CyberHoward: As a helper method yes could be nice!

PR #355, opened 2024-05-22
Update package licenses to MIT

CyberHoward:
See title


CyberHoward:
Update changelog to explicitly mention license inconsistency between abstract-interfaces and abstract-client.

CyberHoward:
Not using MIT

PR #356, opened 2024-05-22
Added incremental upgrade account

Kayanski:
This PR aims at creating a helper for upgrading accounts incrementally (they can only be updated to the next breaking version).

This function will upgrade the proxy and manager contracts incrementally and error if not update was done to neither contracts. 
It will have to do multiple sequential upgrades.

Here is of an update works : 
- Query the current contract version
- Query the available versions above the current version (version control)
- Then do either : 
  - If there is no breaking version above the current one, update to the highest version
      For instance, current version is "0.22.1" and available versions are  ["0.21.0", "0.22.1", "0.22.8"] it will upgrade to "0.22.8"
  - If there is a breaking version above the current one, update the to the highest version compatible to the next breaking version
     For instance, current version is "0.22.1", available version are ["0.22.1", "0.23.0", "0.23.4", "0.25.8"], it will upgrade to "0.23.4"

Added a clone-test to migrate the oldest account to the current version

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-interface/src/account/mod.rs
```diff
@@ -285,6 +291,119 @@ impl<Chain: CwEnv> AbstractAccount<Chain> {
...
+                .versions[0]
+                .clone()
+        };
+        let current_module_version = Version::parse(&current_cw2_module_version.version)?;
+
+        let module = ModuleInfo::from_id(module_id, current_module_version.to_string().into())?;
+
+        // We query all the module versions above the current one
+        let abstr = Abstract::load_from(chain.clone())?;
+        let all_next_manager_versions = abstr
```
Buckram123, line 343: nit: it's not always manager, rename var

framework/packages/abstract-interface/src/account/mod.rs
```diff
@@ -313,3 +432,14 @@ impl<Chain: CwEnv> std::fmt::Display for AbstractAccount<Chain> {
...
+fn are_versions_compatible(v1: &Version, v2: &Version) -> bool {
+    if v1.major == 0 && v2.major == 0 {
+        // For major version 0, compatibility can be more strict; typically, versions are considered
+        // compatible if they have the same minor version
+        v1.minor == v2.minor
+    } else {
+        // For major version >= 1, versions are compatible if they have the same major version
+        v1.major == v2.major
+    }
+}
```
Buckram123, line 445: It's different logic from the one we have in `assert_contract_upgrade`, should we update logic there as well? Wanna add source for the comment? https://semver.org/#spec-item-4

interchain/framework-clone-testing/Cargo.toml
```diff
@@ -26,3 +26,4 @@ env_logger = "0.10.0"
 lazy_static = "1.4.0"
 
 cw2 = { workspace = true }
+cw-semver.workspace = true
```
Buckram123, line 29:
```suggestion
cw-semver = { workspace = true }
```

framework/packages/abstract-interface/src/account/mod.rs
```diff
@@ -285,6 +291,119 @@ impl<Chain: CwEnv> AbstractAccount<Chain> {
...
+            }
+            one_migration_was_successful = true;
+        }
+
+        loop {
+            if self.upgrade_next_module_version(PROXY).is_err() {
+                break;
+            }
+            one_migration_was_successful = true;
+        }
```
Buckram123, line 310: Not sure if they should be updated separately. But it's possible versions won't be the same in the future(hopefully we will merge them into one contract sooner)

interchain/framework-clone-testing/tests/upgrade-account.rs
```diff
@@ -0,0 +1,60 @@
...
+                bail!("No owner for this account")
+            }
+        })
+        .min_by(|(_, _, v1), (_, _, v2)| v1.cmp(v2))
+        .unwrap();
+
+    // List the manager and proxy versions
+
+    Ok((chain, oldest_account_id, owner))
+}
```
Buckram123, line 48: How old was the account you found? I'm too lazy to run it :D

Buckram123 (approved):
Small nits LGTM

framework/packages/abstract-interface/src/account/mod.rs
CyberHoward: For now its fine but we should either do migrations in pairs or have them merged in the future as you say. One way to do that is to block migrations for proxy and do the proxy upgrade along with a manager upgrade when requested.

interchain/framework-clone-testing/tests/upgrade-account.rs
Kayanski: It was 0.19, so not that old, but still

framework/packages/abstract-interface/src/account/mod.rs
Kayanski: When requested by what ? Is there a dependency mechanism you are thinking of ?

framework/packages/abstract-interface/src/account/mod.rs
CyberHoward: What do you think of using the semver crate implementation instead of trying to implement it ourselves (just so we're sure). 

I.e. parse the v1 version into a `VersionReq` type and then assert v2 against it: https://docs.rs/semver/latest/semver/struct.VersionReq.html#method.parse

framework/packages/abstract-interface/src/account/mod.rs
```diff
@@ -328,6 +336,119 @@ impl<Chain: CwEnv> AbstractAccount<Chain> {
...
+
+        if !one_migration_was_successful {
```
CyberHoward, line 357: Why do we want to error if the upgrade wasn't needed? I'd prefer throwing Options here so the user can decide what to do more easily.

Buckram123: Yeah, agree, returning `one_migration_was_successful` would be nicer here

interchain/framework-clone-testing/tests/upgrade-account.rs
```diff
@@ -0,0 +1,60 @@
...
+        .account_factory
+        .config()?
+        .local_account_sequence;
+
+    let (oldest_account_id, owner, _) = (0..max_account_id.min(100))
+        .flat_map(|account_id| {
+            // Create the account helper
+            let account = AbstractAccount::new(&abstr_deployment, AccountId::local(account_id));
+            if let Some(owner) = account.manager.ownership()?.owner {
+                // We print the versions of Manager and Proxy for debugging if there an error
```
Buckram123, line 28: Outdated comment?
```suggestion
```

Buckram123 (approved):
LGTM

PR #358, opened 2024-05-23
Deploy strategy tests

Buckram123:
This PR aims to add tests for deploystrategy and fix behavior of strategies

Fixes ORC-133
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


CyberHoward:
Ok so you changed it to instantiate/register even if there's no change to the code? But that would/should never work because the version of the contract is asserted to be the same as the registered version. 

I.e. if you force update to a new version then either your version should be a new version (which means the bin will be different) or you're trying to update a contract on the same version (in which case the binary should be different as well, since why else are you trying to register this contract under the same version?)

Buckram123:
> Ok so you changed it to instantiate/register even if there's no change to the code? But that would/should never work because the version of the contract is asserted to be the same as the registered version.

Yes because you currently have no way to propose module if it's already uploaded

> I.e. if you force update to a new version then either your version should be a new version (which means the bin will be different) or you're trying to update a contract on the same version (in which case the binary should be different as well, since why else are you trying to register this contract under the same version?)

Could you explain differences between deploy strategies on how you expect them to be working?

CyberHoward:
> > Ok so you changed it to instantiate/register even if there's no change to the code? But that would/should never work because the version of the contract is asserted to be the same as the registered version.
> 
> Yes because you currently have no way to propose module if it's already uploaded
> 
> > I.e. if you force update to a new version then either your version should be a new version (which means the bin will be different) or you're trying to update a contract on the same version (in which case the binary should be different as well, since why else are you trying to register this contract under the same version?)
> 
> Could you explain differences between deploy strategies on how you expect them to be working?

Ahhh okey, didn't think about the case where you manually upload and then try to use this deploy fn.

CyberHoward (approved):

PR #363, opened 2024-05-27
Module to module query

Buckram123:
This PR aims to add module to module query. It adds new type `ModuleQuery{ target_module, msg }` integrated inside `QueryRequest<ModuleQuery>` to query modules from another chain
### Checklist

- [x] CI is green.
- [x] Changelog updated.

Closes ABS-435


framework/packages/abstract-std/src/native/ibc_client.rs
```diff
@@ -121,6 +121,9 @@ pub enum ExecuteMsg {
         msg: Binary,
         /// Callback info to identify the callback that is sent (acts similar to the reply ID)
         callback_info: Option<CallbackInfo>,
+        /// Set this field in case message is query, defaults to `false`
+        /// Don't forget to set callback_info to catch response
+        is_query: Option<bool>,
```
Buckram123, line 126: @CyberHoward It's an ugly message, but we have to be careful with new ser/de types. IMO if we slap some helpers for `IbcClient` that would be enough to make it nice

framework/contracts/native/ibc-host/src/endpoints/packet.rs
```diff
@@ -145,18 +149,36 @@ pub fn handle_host_module_action(
...
+            SystemResult::Err(system_err) => Err(StdError::generic_err(format!(
+                "Querier system error: {system_err}"
+            ))),
+            SystemResult::Ok(ContractResult::Err(contract_err)) => Err(StdError::generic_err(
+                format!("Querier contract error: {contract_err}"),
+            )),
+            SystemResult::Ok(ContractResult::Ok(value)) => Ok(value),
+        }?;
+
+        Ok(response.set_data(bin))
```
CyberHoward, line 169: What's the lifecycle of this bin data? It's catched by polytone and added to the execute response?

Buckram123: Polytone collect all responses, so it will return as ibc-host execution response with this data

Kayanski:
I would use an actual query on the ibc client instead of an execution. That makes more sense code-wise but it will overload the code with more serde derives that's for sure

Buckram123:
> I would use an actual query on the ibc client instead of an execution. That makes more sense code-wise but it will overload the code with more serde derives that's for sure

Do you know how to implement it that way?

Kayanski:
> > I would use an actual query on the ibc client instead of an execution. That makes more sense code-wise but it will overload the code with more serde derives that's for sure
> 
> Do you know how to implement it that way?

Yes I have a good idea

framework/packages/abstract-std/src/native/ibc_host.rs
```diff
@@ -136,6 +136,7 @@ pub enum ExecuteMsg {
         source_module: InstalledModuleIdentification,
         target_module: ModuleInfo,
         msg: Binary,
+        is_query: bool,
```
CyberHoward, line 139: Either rename enum type to `ModuleAction` or create a new variant `ModuleQuery`. 

If ModuleAction then `msg` can be an enum 

```rust
ModuleMsg {
   Execute(Binary)
   Query(Binary)
}
```

Buckram123: Raw could be useful, not sure about the rest

Buckram123: Also name could be confusing, because we already have `ModuleIbcMsg` should it be `ModuleAction` instead?

CyberHoward: > Raw could be useful, not sure about the rest

Yeah let's stick to Binary msg (smart query). We can add raw as a module action later if there's a request for it (but it's better solved with ICQ anyway)

CyberHoward: And `ModuleAction` sounds good!!

framework/packages/abstract-std/src/native/ibc_host.rs
```diff
@@ -131,14 +131,23 @@ pub enum ExecuteMsg {
...
-    /// Allows for remote execution from the Polytone implementation on a local module
-    ModuleExecute {
+    /// Allows for remote execution or request from the Polytone implementation on a local module
+    ModuleAction {
         source_module: InstalledModuleIdentification,
         target_module: ModuleInfo,
-        msg: Binary,
+        msg: ModuleActionMsg,
     },
 }
```
Buckram123, line 140: @CyberHoward ModuleAction inside ModuleAction variant felt weird, went with `ModuleActionMsg`, naming is hard :eyes:

framework/packages/abstract-std/src/native/ibc_host.rs
```diff
@@ -131,14 +131,23 @@ pub enum ExecuteMsg {
         action: HostAction,
     },
     // ANCHOR_END: ibc-host-execute
-    /// Allows for remote execution from the Polytone implementation on a local module
-    ModuleExecute {
+    /// Allows for remote execution or request from the Polytone implementation on a local module
```
CyberHoward, line 134:
```suggestion
    /// Performs an execution or query on a local module
```

CyberHoward (approved):

Buckram123:
@CyberHoward Ready to merge I think!

framework/contracts/native/ibc-host/src/endpoints/packet.rs
```diff
@@ -145,18 +148,39 @@ pub fn handle_host_module_action(
...
+                SystemResult::Ok(ContractResult::Err(contract_err)) => Err(StdError::generic_err(
+                    format!("Querier contract error: {contract_err}"),
+                )),
+                SystemResult::Ok(ContractResult::Ok(value)) => Ok(value),
+            }?;
 
-    Ok(Response::new()
-        .add_attribute("action", "module-ibc-call")
-        .add_message(msg))
+            Ok(response.set_data(bin))
```
Kayanski, line 183: I really don't like the fact that we use the execute data field to get a query result back. Do you have a reasoning behind not using actual queries to query data here and using the data field instead ?

Kayanski: Otherwise, nice implementation

Buckram123: > I really don't like the fact that we use the execute data field to get a query result back. Do you have a reasoning behind not using actual queries to query data here and using the data field instead ?

I don't really get what you mean by "using actual queries" here. Can you show what you mean? I don't know how to make polytone query desired module without calling execute on `ibc-host`

Kayanski: And this is not compatible with https://github.com/AbstractSDK/abstract/pull/369 (adding multiple IBC queries simultaneously)

Kayanski: > I don't really get what you mean by "using actual queries" here. Can you show what you mean? I don't know how to make polytone query desired module without calling execute on `ibc-host`

You can make polytone call the `query` on `ibc-host` instead of `execute

Buckram123: Okay I think I've got what you mean! Create new query method on ibc-host and query from there. I just didn't think of that, lol

CyberHoward: Ahhh yeah that's a good suggestion!!

framework/contracts/native/ibc-client/src/commands.rs
```diff
@@ -307,19 +307,74 @@ pub fn execute_send_query(
...
+            let ibc_infra = IBC_INFRA.load(deps.storage, &host_chain)?;
+            let remote_ibc_host = ibc_infra.remote_abstract_host;
+            // Query the sender module information
+            let module_info = cfg
+                .version_control
+                .module_registry(deps.as_ref())?
+                .module_info(info.sender.clone())?;
+
+            // We need additional information depending on the module type
+            let source_module = match module_info.reference {
```
Kayanski, line 338: This is not necessary here. Let's use target_module as an InstalledModuleIdentification type and let the user query any app associated with any account on the remote chain.

framework/contracts/native/ibc-host/src/endpoints/packet.rs
```diff
@@ -156,7 +160,49 @@ pub fn handle_host_module_action(
...
+    let target_module = InstalledModuleIdentification {
+        module_info: target_module,
+        account_id: source_module
+            .account_id
+            .map(|a| client_to_host_account_id(client_chain.clone(), a)),
+    };
+
+    let target_module_resolved = target_module.addr(deps, vc)?;
+
+    match target_module_resolved.reference {
```
Kayanski, line 185: Why can't we query native or accounts on the remote chain ?

framework/contracts/native/ibc-client/src/commands.rs
Buckram123: Thanks!

framework/packages/abstract-std/src/native/ibc_host.rs
```diff
@@ -131,14 +131,23 @@ pub enum ExecuteMsg {
...
     ModuleExecute {
         source_module: InstalledModuleIdentification,
         target_module: ModuleInfo,
         msg: Binary,
     },
 }
 
+/// Module to module action message
+#[cosmwasm_schema::cw_serde]
+pub enum ModuleActionMsg {
```
CyberHoward, line 144: This type is unused now I think?

framework/packages/abstract-std/src/native/ibc.rs
```diff
@@ -130,6 +115,21 @@ impl IbcResult {
...
+        } = &self
+        {
+            Ok((queries[index].clone(), result[index].clone()))
+        } else {
+            Err(StdError::generic_err(
+                "Failed to parse module to module query response",
+            ))
+        }
+    }
 }
```
Buckram123, line 133: Do we want to verify query[index] is custom query? Because right now it's just .get(index) query response

Buckram123: @CyberHoward

Buckram123: Besides that LGTM

CyberHoward: But don't really see why we would do it for Custom but not for all the other query types?

Buckram123: > We either do that for all query types or we do what we do right now and let the dev parse the result. We can add helpers to do the different query parsing later (maybe through a separate IBC query API).

Yeah, comment confused me a little

Buckram123: > But don't really see why we would do it for Custom but not for all the other query types?

```
    /// Get module to module query responses
```

CyberHoward:
Fixed interchain tests in #354

PR #365, opened 2024-05-28
Added public cw-orch-interchain crate

Kayanski:
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-adapter/src/endpoints/execute.rs
```diff
@@ -154,9 +154,10 @@ impl<Error: ContractError, CustomInitMsg, CustomExecMsg, CustomQueryMsg, Receive
             }
             None => account_registry
                 .assert_manager(sender)
-                .map_err(|_| unauthorized_sender())?,
+                .map_err(|_| unauthorized_sender())
+                .ok(),
         };
-        self.target_account = Some(account_base);
+        self.target_account = account_base;
```
CyberHoward, line 160: This change allows calling the adapter without adding a sender. while it's a good hackathon fix the standalone module should really handle that use case. I don't think we should merge this change.

interchain/interchain-tests/src/setup.rs
```diff
@@ -1,10 +1,32 @@
+use abstract_interface::connection::abstract_ibc_connection_with;
 use abstract_interface::Abstract;
-use abstract_scripts::abstract_ibc::abstract_ibc_connection_with;
 use anyhow::Result as AnyResult;
 use cw_orch::prelude::*;
+use cw_orch_interchain::prelude::*;
 use cw_orch_polytone::Polytone;
 use polytone::handshake::POLYTONE_VERSION;
 
+pub fn ibc_connect_abstract<Chain: IbcQueryHandler, IBC: InterchainEnv<Chain>>(
```
CyberHoward, line 9: Should this connect both directions?

framework/packages/abstract-interface/src/ibc.rs
```diff
@@ -62,3 +60,61 @@ impl<Chain: CwEnv> AbstractIbc<Chain> {
...
+    use abstract_std::ibc_client::QueryMsgFns;
+    use abstract_std::ibc_host::ExecuteMsgFns as _;
+    use abstract_std::objects::chain_name::ChainName;
+    use cw_orch_interchain::prelude::*;
+    use cw_orch_interchain::InterchainError;
+    use cw_orch_polytone::Polytone;
+    /// This is used for creating a testing connection between two Abstract connections using an existing polytone connection
+    ///
+    /// You usually don't need this function on actual networks if you're not an Abstract maintainer
+    pub fn abstract_ibc_connection_with<Chain: IbcQueryHandler, IBC: InterchainEnv<Chain>>(
```
CyberHoward, line 79: This should connect both ways.

Buckram123: Couldn't make it work, got `HostAddressExists` when trace is longer than 1

Buckram123: So there is `test_multi_hop_account_creation` test where I managed to connect in both ways, (polytone1-polytone-2 and polytone2-polytone-1) but then it's not possible to connect polytone2 to polytone3. Let's move this to separate task, so we don't dive too hard here

CyberHoward (approved):

framework/packages/abstract-interface/src/ibc.rs
Kayanski: Ok, I now understand why this is failing and attempting to fix that now.

Buckram123: > Ok, I now understand why this is failing and attempting to fix that now.

Is it related to this PR?

Kayanski: Yes, this is regarding your `test_multi_hop_account_creation` test. This will be fixed in a future PR, with better handling of the polytone complexity. See internal comms for suggestions

PR #366, opened 2024-06-02
Tokenized ownership transfer

hard-nett:
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


CyberHoward:
Just had a read through the `BITSONG.md` file. I'm hesitant to include the bitsong marketplace address inside of the manager and doing NFT minting in the factory. 

What I think we could do instead is add a different `GovernanceDetails` type that is `GovernanceDetails::NFT`. 

This new variant would contain the NFT token address that represents the account as well as the unique NFT-id. 
The `assert_admin` function that is used in all our contracts can then be updated to 

1. Query the NFT token address for the owner of the NFT associated with this account (given by the NFT-id). 
2. Assert that the caller is this owner. 

The marketplace can then be something completely separated from the rest of the framework. What do you think? Am I missing something?

hard-nett:
@CyberHoward thank you very much for taking some time to review.

The one feature that bitsong specifically requested here is when the nft ownership is transferred , the new owner becomes owner of the account. Making use of `GovernanceDetails::NFT` would end up being much more efficient than this current implementation, which semi-hacks the `cw_ownable` entry points to allow the marketplace to function as an escrow, until the new owner claims ownership of the manager. 

A fun thing about implementing a fork stargaze names is that the token_id is the name itself.


 is much more future proof than this first implementation, so I will create a fork with this implementation now!

hard-nett:
closed in reference to #367

PR #368, opened 2024-06-04
abstract-standalone

Buckram123:
This PR includes abstract-standalone package and template
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/contracts/native/module-factory/src/commands.rs
```diff
@@ -21,6 +23,30 @@ use crate::{
...
+// https://github.com/serde-rs/serde/issues/984#issuecomment-314143738
+// Any value that is present is considered Some value, including null.
+fn deserialize_some<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>
+where
+    T: cosmwasm_schema::serde::Deserialize<'de>,
+    D: cosmwasm_schema::serde::Deserializer<'de>,
+{
+    cosmwasm_schema::serde::Deserialize::deserialize(deserializer).map(Some)
+}
+
```
Buckram123, line 49: At first thought it's fine if user fills base themself, but after adding `AccountBase` inside it became annoying because if account is not created yet you need to get address of proxy or manager with instantiate2 first.
Alternative to this solution we could inherit `{base, module}` structure from apps (we will still need to verify base is `null` if we want to support arbitrary messages as well)

framework/packages/abstract-sdk/src/apis/ibc.rs
```diff
@@ -98,6 +98,66 @@ impl<'a, T: IbcInterface> IbcClient<'a, T> {
...
+                &self.deps.querier,
+                ibc_client_addr,
+                (&trace, sequence, &host_chain.parse()?),
+            )
+            .map_err(Into::into)
+    }
+}
+
+impl<'a, T: IbcInterface + AccountExecutor> IbcClient<'a, T> {
     /// Registers the ibc client to be able to use IBC capabilities
```
CyberHoward, line 161: I really love how extensible these APIs are, it's 🔥

framework/contracts/templates/standalone/src/contract.rs
CyberHoward, line 1: Wondering if we want to have this resemble the CosmWasm template contract as closely as possible?

I.e. we move the handlers in `handlers/*` to this file and move all the orchestrator related code to an `interface.rs` file?

framework/packages/abstract-standalone/src/state.rs
```diff
@@ -0,0 +1,80 @@
...
+use abstract_std::{
+    objects::{dependency::StaticDependency, module::ModuleInfo, nested_admin::NestedAdmin},
+    standalone::StandaloneState,
+    AbstractResult,
+};
+use cosmwasm_std::{StdResult, Storage};
+use cw_storage_plus::Item;
+
+/// The state variables for our StandaloneContract.
+pub struct StandaloneContract {
```
CyberHoward, line 15: I think it would be great to have dependencies on standalones as well. That way we can include the ibc client as a dependency, as well as enable multiple standalones to depend on each other (like a classical SC application)

framework/packages/abstract-standalone/src/features.rs
CyberHoward, line 1: Like discussed would be great to add `Dependency` feature here.

framework/packages/abstract-interface/src/account/mod.rs
```diff
@@ -177,6 +177,17 @@ impl<Chain: CwEnv> AbstractAccount<Chain> {
         self.install_module_parse_addr(app, Some(&custom_init_msg), funds)
     }
 
+    /// Installs an standalone from an standalone object
+    pub fn install_standalone<CustomInitMsg: Serialize, T: ContractInstance<Chain>>(
```
CyberHoward, line 181: add `install_standalone_witd_dependencies`. 

Also see https://discord.com/channels/976930997654405132/976931523582382090/1248567756740100146

framework/contracts/native/module-factory/src/commands.rs
CyberHoward: Why would they need to include the AccountBase? That can be interpolated from the admin right?

framework/contracts/templates/standalone/src/contract.rs
Buckram123: You're right, will also make easier migration for standard contracts that move to abstract if we use default structure!

framework/contracts/native/module-factory/src/commands.rs
CyberHoward: I.e. query query account_base from the caller (which should be a manager)

framework/packages/abstract-std/src/standalone.rs
```diff
@@ -0,0 +1,35 @@
...
+pub struct BaseInstantiateMsg {
+    pub ans_host_address: String,
+    pub version_control_address: String,
+    pub account_base: AccountBase,
+}
+
+/// The BaseState contains the main addresses needed for sending and verifying messages
+#[cosmwasm_schema::cw_serde]
+pub struct StandaloneState {
+    pub proxy_address: Addr,
```
CyberHoward, line 28: Why does the state need to hold this?

cyberhoward (commented):
Great PR :)

framework/contracts/native/module-factory/src/commands.rs
Buckram123: > I.e. query query account_base from the caller (which should be a manager)

Caller of instantiate msg is module-factory.

Do you know if "contract admin" set during instantiation?

framework/packages/abstract-interface/src/account/mod.rs
Buckram123: > add `install_standalone_witd_dependencies`.

Done
 
> Also see https://discord.com/channels/976930997654405132/976931523582382090/1248567756740100146

Let's have it in separate PR. It's easy to implement if we just check if it's installed and "pop" from the list if it is. But I'm not sure what we do with version miss-matches? Do we just let it fail or try to update if it's newer version?

Buckram123:
I guess not breaking anymore

framework/packages/abstract-standalone/src/endpoints/instantiate.rs
```diff
@@ -0,0 +1,59 @@
...
+    AbstractError,
+};
+use cosmwasm_std::{Addr, DepsMut, Env};
+use cw2::set_contract_version;
+
+use crate::state::StandaloneContract;
+
+impl StandaloneContract {
+    /// Call this method on instantiating of the standalone
+    pub fn instantiate(
```
Kayanski, line 17: So this means that users can't send a custom instantiate Msg to the Standalone (or migrate msg later ?)

Buckram123: Using those methods is optional and if you want to use it you can also have any structure, even without base standalone instantiate message(for example having vc and ans addrs included in your message), as long as you sent both here!

framework/packages/abstract-standalone/src/endpoints/instantiate.rs
```diff
@@ -0,0 +1,59 @@
...
+        &self,
+        deps: DepsMut,
+        env: &Env,
+        msg: BaseInstantiateMsg,
```
CyberHoward, line 21: Maybe we should rename this to `StandaloneInstantiateMsg`

framework/packages/abstract-standalone/src/endpoints/instantiate.rs
CyberHoward: > So this means that users can't send a custom instantiate Msg to the Standalone (or migrate msg later ?)

We don't implement any endpoints for the developer so it's completely optional.

Kayanski: Oh got it thanks

framework/packages/abstract-std/src/error.rs
```diff
@@ -79,6 +79,9 @@ pub enum AbstractError {
         from: Version,
         to: Version,
     },
+
+    #[error("Migration is not enabled on contract")]
```
CyberHoward, line 83: It's a bit weird that we include this in the abstract-std package when it's only used in standalone, no?

cyberhoward (commented):
Let's move the template to the template repo!

framework/packages/abstract-std/src/error.rs
Buckram123: Would you prefer it as `generic_err` then? Don't think it's worth adding `StandaloneErr` type just for one error

Buckram123:
> Let's move the template to the template repo!

Will be easier with new abstract version on crates.io! Just tried to publish this version, but this PR includes changes to abstract-std that this package relies on

cyberhoward (approved):

PR #369, opened 2024-06-07
Rationalize the Abstract IBC API

Kayanski:
This PR aims at rationalizing the Abstract IBC experience.

Changes : 
- Module Ibc Action message is already decomposed in module identification (src_module_info) and message (msg)
- IBC Callback IDs don't exist anymore
- IBC callback messages are now required when using a callback
- IBC callback messages are now called payload throughout to avoid confusion with `initiator_message` that the callback result includes
- Ability to send multiple IBC queries simultaneously 

If we want to make the experience similar to SubMgs, we need to create a new data structure (with a different name) that handles the case. The best solution is to have different API functions that are able to construct messages with a without callbacks and this is already sufficient.

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-std/src/native/ibc.rs
```diff
@@ -15,29 +15,24 @@ use crate::{
 #[cosmwasm_schema::cw_serde]
 // ANCHOR: callback-info
 pub struct CallbackInfo {
-    /// Used to identify the callback that is sent (acts like the reply ID)
-    pub id: String,
     /// Used to add information to the callback.
     /// This is usually used to provide information to the ibc callback function for context
-    pub msg: Option<Binary>,
+    pub payload: Binary,
```
CyberHoward, line 20: Why did you opt to use payload here? It's a name that isn't really used anywhere else while "msg" indicates that it's a message type, and should be deserialized as such.

Kayanski: Because payload can be anything. And msg can be confused with initiator_msg that we have in the Executor case

CyberHoward: How can it be anything? We want to set conventions and imo we should encourage the use of `CallbackMsg` as the convention

Buckram123: Pretty sure it have more usecases than just being a message. cosmwasm_std::Reply also use `payload` in this case https://docs.rs/cosmwasm-std/2.0.4/cosmwasm_std/struct.Reply.html#structfield.payload

Buckram123 (approved):
Very nice!

CyberHoward (approved):
LGTM appart from the `payload` rename.

framework/packages/abstract-std/src/native/ibc.rs
CyberHoward: On replies you still have the REPLY_ID that is used to uniquely identify replies. If we don't set the soft standard to use messages then it could be very easy to use it incorrectly.

If both of you agree then its fine the way it is. 

I do think we could rename `CallbackInfo` as well. It's kind of a weird name, would prefer just `Callback`

CyberHoward:
Made some minor changes: 
- Rename `CallbackInfo` to `Callback`
- Rename `payload` in `IbcResponseMsg` to `callback` (also changed type to match that).
- Rename `payload` in `Callback` to `msg`
- Accept `Callback` in ibc_handler fn instead of `Binary`. That way we can easily add fields to the `Callback` type in the future.

CyberHoward:
Clippy is fixed in #354

PR #373, opened 2024-06-11
Added integration test in documentation

Kayanski:
This Pr aims at adding an integration-test example inside the documentation

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


cyberhoward (changes requested):
I think we should showcase a test that uses the abstract-client, not the abstract-interface, because that's the library we expect devs to be using.

cyberhoward (approved):

PR #376, opened 2024-06-14
Verify IBC client installed for module to module app action

Buckram123:
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/contracts/native/ibc-client/src/commands.rs
```diff
@@ -251,6 +252,14 @@ pub fn execute_send_module_to_module_packet(
                 .query(&deps.querier, info.sender.clone())?
                 .proxy_address;
             let account_id = cfg.version_control.account_id(&proxy_addr, &deps.querier)?;
+            // Check that ibc_client is installed on account
+            let proxy_state = proxy::state::STATE.query(&deps.querier, proxy_addr)?;
```
CyberHoward, line 256: I would prefer this query to use the manager and check the module there. I don't think the ibc-client is actually registered to allow to execute on the proxy (and if it is we should re-consider that IMO)

Buckram123: Removed ibc client and standalone modules from proxy whitelist!

framework/packages/abstract-std/src/objects/module.rs
```diff
@@ -333,6 +333,19 @@ impl From<(ModuleInfo, ModuleReference)> for Module {
     }
 }
 
+impl Module {
+    // Helper to know if this module supposed to be whitelisted on proxy contract
+    pub fn proxy_whitelisted(&self) -> bool {
```
CyberHoward, line 338:
```suggestion
    pub fn should_be_whitelisted(&self) -> bool {
```

framework/packages/abstract-std/src/objects/module.rs
```diff
@@ -333,6 +333,19 @@ impl From<(ModuleInfo, ModuleReference)> for Module {
...
+        match &self.reference {
+            // Standalone not supposed to be whitelisted on proxy
+            ModuleReference::Standalone(_) => false,
+            // IBC Client not supposed to be whitelisted on proxy
+            ModuleReference::Native(_) => self.info.id() != IBC_CLIENT,
```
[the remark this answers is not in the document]
Buckram123: Don't think there are now! Updated to false on native

CyberHoward (approved):

PR #377, opened 2024-06-14
IBC Client as dependency checks

Buckram123:
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-sdk/src/base/endpoints/modules_ibc.rs
```diff
@@ -16,6 +17,19 @@ pub trait ModuleIbcEndpoint: Handler {
         info: MessageInfo,
         msg: ModuleIbcMsg,
     ) -> Result<Response, Self::Error> {
+        // Make sure module have ibc_client as dependency
+        if !self
```
CyberHoward, line 21: When a module receives an IBC message it isn't required to have a dependency on the IBC client (the message is sent from the host)

CyberHoward (changes requested):
Sweet! I'd like us to also add the assertion in the IBC APIs, like here: https://docs.rs/abstract-sdk/latest/src/abstract_sdk/apis/app.rs.html#78

framework/packages/abstract-sdk/src/apis/ibc.rs
```diff
@@ -87,11 +87,11 @@ pub struct IbcClient<'a, T: IbcInterface> {
 impl<'a, T: IbcInterface> IbcClient<'a, T> {
     /// Get address of this module
     pub fn module_address(&self) -> AbstractSdkResult<Addr> {
```
CyberHoward, line 89: Could we rename this to `address` instead of `module_address`?

CyberHoward (approved):

framework/packages/abstract-sdk/src/apis/ibc.rs
Buckram123: We use `module_address` in `Modules` and `CronCat` API's as well! Would you prefer to use `address` there too?

PR #379, opened 2024-06-18
Fix standalone instantiate package on chain

Buckram123:
The standalone contract attempted to use the wasm admin of the contract to figure out the account details as the admin is normally set to be the manager contract. However, on-chain this data is not available during instantiation as the contract address is not yet assigned, returning `Contract <address> does not exist`.

To fix this we now save the account details in the module factory if a standalone contract is created. The standalone instantiate fn then expects the `info.sender` to be the module factory and queries this state. 

### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/packages/abstract-standalone/src/endpoints/instantiate.rs
```diff
@@ -17,7 +16,7 @@ impl StandaloneContract {
     pub fn instantiate(
```
CyberHoward, line 16: Can you add a doc comment to ensure users know the info.sender must be the module factory?

```
/// Instantiates the `Standalone` state for this contract. 
/// **Note:** This contract can only be instantiated by the abstract module factory. 
```

CyberHoward (approved):

PR #382, opened 2024-06-20
NFT governance refactor

CyberHoward:
### Checklist

- [ ] CI is green.
- [x] Changelog updated.


framework/packages/gov-ownable/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+    let new_owner = new_owner.verify(deps.as_ref(), version_control)?;
+
+    if let GovernanceDetails::Renounced {} = new_owner {
+        return Err(GovOwnershipError::TransferToRenounced {});
+    }
+
+    OWNERSHIP.update(deps.storage, |ownership| {
+        // the contract must have an owner
+        check_owner(&deps.querier, &ownership, sender)?;
+
```
CyberHoward, line 155: What I think we should add here is a check for if the current owner is of type NFT. If so, we should prevent the ownership transfer. Otherwise any NFT holder can transfer the account away from the NFT (which is probably not something the NFT creators expect the user to be able to do!).

framework/packages/gov-ownable/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "cw-gov-ownable"
+version = "0.1.0"
+description = "Utility for controlling ownership of CosmWasm smart contracts"
+authors = { workspace = true }
```
CyberHoward, line 5: We should update authors to include larry!

framework/packages/abstract-std/src/core/manager.rs
```diff
@@ -174,7 +138,6 @@ impl ModuleInstallConfig {
 }
 
 /// Manager execute messages
-#[cw_ownable::cw_ownable_execute]
```
Buckram123, line 177: Would you prefer to keep it as macro or as variant? I personally dislike macros that introduce message variants, as they are harder to read IMO

framework/contracts/native/account-factory/src/commands.rs
```diff
@@ -364,3 +378,28 @@ pub fn execute_update_config(
...
+        &cw721::Cw721QueryMsg::OwnerOf {
+            token_id: nft_id,
+            include_expired: None,
+        },
+    )?;
+    // verify owner
+    ensure_eq!(
+        sender,
+        owner.owner,
+        AbstractError::Ownership(OwnershipError::NotOwner)
```
Buckram123, line 401: It's the only place where `AbstractError::Ownership` used, should we replace it to something else?

framework/packages/abstract-std/src/error.rs
```diff
@@ -18,6 +18,9 @@ pub enum AbstractError {
     #[error("cw math overflow error: {0}")]
     Overflow(#[from] OverflowError),
 
+    #[error("{0}")]
+    Ownership(#[from] cw_ownable::OwnershipError),
+
```
Buckram123, line 23: Should we replace it to just not owner error? See my previous comment

framework/packages/gov-ownable/derive/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "cw-gov-ownable-derive"
+version = "0.1.0"
```
Buckram123, line 3: Do we want to maintain this derive repo? I would happily throw it :)

Buckram123 (commented):
Looks safe. We have a ton of tests that covers ownership cases. The only scarry thing is migration, but it clone tests seems happy, so am I

framework/contracts/account/manager/src/commands.rs
```diff
@@ -571,8 +573,8 @@ pub(crate) fn renounce_governance(
         ManagerError::RenounceWithSubAccount {}
     );
 
-    let mut account_info = INFO.load(deps.storage)?;
-    if let GovernanceDetails::SubAccount { manager, proxy } = account_info.governance_details {
+    let ownership = cw_gov_ownable::get_ownership(deps.storage)?;
+    if let GovernanceDetails::SubAccount { manager, proxy } = ownership.owner {
         // If called by top-level owner, update the sender to let cw-ownable think it was called by the proxy.
         let top_level_owner = query_top_level_owner(&deps.querier, manager_addr)?;
         if top_level_owner == *sender {
```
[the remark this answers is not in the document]
Buckram123: Already discussed in voice and we should

framework/packages/gov-ownable/src/lib.rs
```diff
@@ -0,0 +1,572 @@
...
+fn renounce_ownership(
+    store: &mut dyn Storage,
+    querier: &QuerierWrapper,
+    sender: &Addr,
+) -> Result<Ownership<Addr>, GovOwnershipError> {
+    OWNERSHIP.update(store, |ownership| {
+        check_owner(querier, &ownership, sender)?;
+
+        Ok(Ownership {
+            owner: GovernanceDetails::Renounced {},
```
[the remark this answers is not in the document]
Buckram123: Already discussed in voice and we should

framework/packages/gov-ownable/derive/Cargo.toml
CyberHoward: We have to keep it (imports are different)

framework/packages/abstract-std/src/core/manager.rs
CyberHoward: Agreed! let's get rid of the macro

framework/contracts/native/account-factory/src/commands.rs
CyberHoward: Let's try to do your suggestion after moving the gov implementations to std. If this is still the only use of the OwnershipError type then we can remove it like suggested.

framework/packages/abstract-std/src/objects/ownership/gov_ownable.rs
```diff
@@ -0,0 +1,674 @@
...
+    version_control: Addr,
+    expiry: Option<Expiration>,
+) -> Result<Ownership<Addr>, GovOwnershipError> {
+    let new_owner = new_owner.verify(deps.as_ref(), version_control)?;
+
+    if let GovernanceDetails::Renounced {} = new_owner {
+        return Err(GovOwnershipError::TransferToRenounced {});
+    }
+
+    OWNERSHIP.update(deps.storage, |ownership| {
```
CyberHoward, line 240: How are we handeling transfers from a regular account to a sub-account? From this it seems like you can propose it but then somehow the manager of the top-level account must accept the transfer. Is this the expected behaviour? I thought we had a different system before where we asserted that the top-level owner of the *proposed* sub-account was the sender (this ensured that one can only add sub-accounts to accounts you already own.)

Buckram123: Well noted! Accidentally removed option of transferring account not as top level owner, just fixed it and covered it with tests, that don't use recursive querying :)

Buckram123: Moving check of pending direct/top level into gov-ownable as well from manager right now

interchain/framework-clone-testing/tests/upgrade-account.rs
```diff
@@ -51,10 +54,51 @@ fn find_old_account() -> anyhow::Result<(CloneTesting, u32, String)> {
...
+
+    account.upgrade(&abstr_deployment)?;
+    let info_account = account.manager.module_info(PROXY)?.unwrap();
+    assert_eq!(info_account.version.version, TEST_VERSION);
+
+    let info_sub_account = sub_account.manager.module_info(PROXY)?.unwrap();
+    assert_eq!(info_sub_account.version.version, TEST_VERSION);
+
+    let info_sub_sub_account = sub_sub_account.manager.module_info(PROXY)?.unwrap();
+    assert_eq!(info_sub_sub_account.version.version, TEST_VERSION);
```
Buckram123, line 101: `finished in 57.54s` Can't wait to add some caching to clone testing

framework/packages/abstract-std/src/objects/ownership/gov_ownable.rs
Buckram123: Done!

framework/packages/abstract-std/src/objects/ownership/gov_ownable.rs
```diff
@@ -0,0 +1,709 @@
...
+                pending_owner: None,
+                pending_expiry: None,
+            }
+        } else {
+            Ownership {
+                pending_owner: Some(new_owner),
+                pending_expiry: expiry,
+                ..ownership
+            }
+        };
```
Buckram123, line 300: @CyberHoward This auto-accept doesn't feel right to me. NFT contract cannot accept it, but we still can get owner of it.

We have 3 endpoints on UpdateOwnership

- Transfer ownership: propose owner
- Accept ownership: proposed governance accepts ownership
- Renounce: toss ownership to void

And this NFT case doesn't match to the regular usage of `Transfer Ownership`, so I would suggest either put it into pending_owner and check owner of NFT during accept, or create separate endpoint `NFT: Toss ownership to nft`

Buckram123: Implemented first suggestion, IMO that's preferred way to handle it

CyberHoward: Ok so you're suggesting that the owner of the NFT must accept the ownership for this so succeed? That makes sense.

Buckram123: > Ok so you're suggesting that the owner of the NFT must accept the ownership for this so succeed? That makes sense.

Yes, nice!

framework/packages/abstract-std/src/objects/ownership/gov_ownable.rs
```diff
@@ -39,7 +39,7 @@ pub enum GovOwnershipError {
     #[error("Cannot transfer ownership to renounced structure. use action::renounce")]
     TransferToRenounced,
 
-    #[error("Cannot change NFT ownership")]
+    #[error("Cannot change NFT ownership. Transfer the NFT to change account ownership.")]
```
Buckram123, line 42: We use this error for Renounce as well! Would you prefer to have separate error for renounce then?

CyberHoward (commented):
LGTM!!

Buckram123 (approved):
LGTM

framework/packages/abstract-std/src/objects/ownership/gov_ownable.rs
CyberHoward: I checked and it doesn't look like we do.

PR #383, opened 2024-06-20
Update Native contract in version control

Buckram123:
This PR aims to add updating of native contracts in version control

- update_natives_in_version_control added to the `abstract_interface::Abstract` for registering new versions of native contracts in Version Control (useful for refreshing VC, did run this method on pion)
- Added registration migrated native contracts to Version Control in `migrate_if_version_changed` method
- IBC client and host now deployed instead of migration on version change

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/packages/abstract-interface/src/migrate.rs
+++ b/framework/packages/abstract-interface/src/migrate.rs
@@ -148,10 +148,10 @@ impl<T: CwEnv> Abstract<T> {
                 ModuleInfo::from_id_latest(MODULE_FACTORY)?,
                 ModuleInfo::from_id_latest(VERSION_CONTROL)?,
                 ModuleInfo::from_id_latest(ANS_HOST)?,
-                ModuleInfo::from_id_latest(MANAGER)?,
-                ModuleInfo::from_id_latest(PROXY)?,
                 ModuleInfo::from_id_latest(IBC_CLIENT)?,
                 ModuleInfo::from_id_latest(IBC_HOST)?,
+                // ModuleInfo::from_id_latest(MANAGER)?,
+                // ModuleInfo::from_id_latest(PROXY)?,
             ])?
             .modules;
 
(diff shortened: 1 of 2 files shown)


framework/packages/abstract-interface/src/migrate.rs
```diff
@@ -151,39 +261,26 @@ fn contract_version<Chain: CwEnv, A: ContractInstance<Chain>>(
...
+        self.client
+            .upload_if_needed()?
+            .expect("IBC client wasm might be outdated");
+        self.host
+            .upload_if_needed()?
+            .expect("IBC host wasm might be outdated");
+        self.instantiate(abstr, &self.client.get_chain().sender())?;
 
-        Ok(has_migrated)
+        Ok(true)
```
Buckram123, line 284: revert

CyberHoward:
What we want is: 
When migrating, if IBC-contract versions are compatible, migrate them and register them. If they are breaking, deploy a new version and register them.
And then prevent migration of ibc-client module

framework/contracts/account/manager/tests/upgrades.rs
```diff
@@ -911,6 +912,26 @@ fn install_app_with_proxy_action() -> AResult {
     abstract_integration_tests::manager::install_app_with_proxy_action(chain)
 }
 
+#[test]
+fn native_not_migratable() -> AResult {
```
CyberHoward, line 916:
```suggestion
fn ibc_client_not_migratable() -> AResult {
```

Buckram123: The reason why it's not migratable is because module type native is not migratable. We just have only 1 installable native module right now :)

framework/packages/abstract-interface/src/migrate.rs
```diff
@@ -114,26 +133,130 @@ impl<T: CwEnv> Abstract<T> {
...
         }
 
+        self.version_control.register_natives(natives_to_register)?;
         self.version_control.approve_any_abstract_modules()?;
 
         Ok(has_migrated)
     }
+
+    /// Update native modules in version control  
+    pub fn update_natives_in_version_control(&self) -> Result<(), crate::AbstractInterfaceError> {
```
CyberHoward, line 168: The logic here is very similar to `fn migrate_if_version_changed`, is there any way to merge let them share a function?

framework/packages/abstract-interface/src/migrate.rs
```diff
@@ -162,28 +285,47 @@ impl<Chain: CwEnv> AbstractIbc<Chain> {
...
-                .client
-                .upload_and_migrate_if_needed(&ibc_client::MigrateMsg {})?;
-            if migration_result.is_some() {
-                has_migrated = true;
-            }
+    /// Deploys or Migrates IBC Client and IBC Host
+    /// - If no version changes - don't do anything
+    /// - If version change is breaking - new version is deployed
+    /// - If version change is non-breaking - ibc contracts migrated instead
+    pub fn deploy_or_migrate_if_version_changed(
```
CyberHoward, line 292: if this is only ibc contracts then the fn name should reflect that. 

`upgrade_ibc_contracts` or similar. 

We should have a clear naming scheme for all the upgrade/migrate/deploy functions.

framework/packages/abstract-interface/src/migrate.rs
Buckram123: This function checks module in VC, while `migrate_if_version_changed` checks cw2 of contracts.

So this function is supposed to be used just to hotfix outdated versions of native contracts in Version Control when cw2 != module version in VC

framework/packages/abstract-interface/src/migrate.rs
Buckram123: > if this is only ibc contracts then the fn name should reflect that.
> 
> `upgrade_ibc_contracts` or similar.
> 
> We should have a clear naming scheme for all the upgrade/migrate/deploy functions.

I don't get it, It's under `AbstractIbc`, you can't skip **.ibc** .deploy_or_migrate_if_version_changed() when you use it

PR #384, opened 2024-06-24
docs: IBC callback warning

Kayanski:
This PR aims at adding a little warning for sending remote modules messages. 
This allows developers to understand that a callback is needed if they want to revert state after a failed remote execution. 

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Buckram123 (approved):
Is it cheaper to set massive value in storage and remove on callback in case of failure or have a massive callback payload? Pretty sure removing state also pretty costy. We should propose cheaper (and IMO safer) solution

Kayanski:
> Is it cheaper to set massive value in storage and remove on callback in case of failure or have a massive callback payload? Pretty sure removing state also pretty costy. We should propose cheaper (and IMO safer) solution

I really can't tell which solution is cheaper. This is just like for the reply, you can use state or payload, dunno which is more efficient in terms of gas. I'm sure I know what is safer though (no storage), Will add an additional comment on this

cyberhoward (approved):

PR #386, opened 2024-06-27
Add client Interchain builder

Kayanski:
This PR aims at adding a Client Interchain Builder that registers everything needed across all chains (channels, packets...).

## Done

- [x] Create builder pattern
- [x] Connect all chains with IBC automatically on build

## TODO

- [ ] Propagate the denoms across chains 
- [ ] Propagate the dexes across chains

## Experiment

the AbstractInterchainClientBuilder::build_daemon method tries to register assets from chain A on chain B by figuring out their IBC denom. It's kind of difficult but goes along the following lines: 
- Find the channels on chain B that are connected to chain A that have an ics20-1 protocol version
- Find the right channel by querying the denom hash associated with the trace (defined here https://github.com/cosmos/ibc/blob/main/spec/app/ics-020-fungible-token-transfer/README.md#data-structures). Here I'm not sure wether the _denom_hash works or not when the trace doesn't exist on the chain.
- Register the IBC hash with the Abstract deployment on chain B

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


CyberHoward:
Closed in favor of #469

PR #387, opened 2024-07-01
Remove terraswap integration from terra

Buckram123:
Terra dex adapter wasm is too big. Terraswap not being used, so let's remove it for now
### Checklist

- [x] CI is green.
- [x] Changelog updated.


CyberHoward:
We should remove the terraswap code in `modules/contracts/adapters/dex/src/exchanges/terraswap.rs`

Buckram123:
> We should remove the terraswap code in `modules/contracts/adapters/dex/src/exchanges/terraswap.rs`

Do you think we should just remove it or comment for now, in case we find a way to use it?

CyberHoward:
> > We should remove the terraswap code in `modules/contracts/adapters/dex/src/exchanges/terraswap.rs`
> 
> Do you think we should just remove it or comment for now, in case we find a way to use it?

Remove, we can always go to a previous tag / release to find it!

cyberhoward (approved):

PR #394, opened 2024-07-03
Enable stargate feature on contracts

CyberHoward:
This enables the stargate feature on all contracts, allowing users to execute stargate messages on their accounts and over IBC


Buckram123 (commented):
New wasms is the same, stargate was already enabled for all contracts it seems

framework/Cargo.toml
```diff
@@ -32,7 +32,7 @@ license = "GPL-3.0-or-later"
 keywords = ["cosmos", "cosmwasm", "framework"]
 
 [workspace.dependencies]
-cosmwasm-std = { version = "1.5.0", features = ["cosmwasm_1_2", "stargate"] }
```
CyberHoward: even if the stargate feature is enabled implicitly I'd still prefer having it enabled explicitly, no?

interchain/interchain-tests/src/interchain_accounts.rs
```diff
@@ -161,6 +161,79 @@ mod test {
...
+            "ics20-1",
+            None, // Unordered channel
+        )?;
+
+        // The user on origin chain wants to change the account description
+        let ibc_transfer_result = origin_account.manager.execute_on_module(
+            PROXY,
+            &proxy::ExecuteMsg::ModuleAction {
+                msgs: vec![cosmwasm_std::CosmosMsg::Ibc(
+                    cosmwasm_std::IbcMsg::Transfer {
```
CyberHoward, line 201: This isn't a stargate message afaik.

Buckram123: It's not stargate message, but message enabled by stargate feature, what are we supposed to be checking here?

framework/Cargo.toml
Buckram123: It's enabled on proxy explicitly

interchain/interchain-tests/src/interchain_accounts.rs
CyberHoward: Ok, just looked it up, you're right. Would have expected it to be under the ibc feature. 
LGTM!

framework/Cargo.toml
CyberHoward: and ibc-client / host / manager?

PR #395, opened 2024-07-04
Fix account_sequence saving provided account_seq

CyberHoward:
The account_seq of the account_factory was being saved after account instantiation without taking into account if the account_seq was provided or not. This resulted in the sequence jumping around (instead of being sequential). We fix this by updating the sequence logic to save the sequence right after retrieving it


Buckram123 (approved):
Oopsie, almost halved our maximum number of account

PR #398, opened 2024-07-10
Add memo to the IBC Client method `send_funds`

Buckram123:
### Checklist

- [ ] CI is green.
- [x] Changelog updated.

### Failing CI

- [x] astroport migrated to a new contracts that use stargate, fix: #396
- [x] tests on neutron fails because we have `pion-1` migrated to 0.23.0-beta, but we have some changes to the contracts after merge, fix: #405


framework/packages/abstract-std/src/native/ibc_client.rs
```diff
@@ -96,7 +96,7 @@ pub enum ExecuteMsg {
         /// host chain to be executed on
         /// Example: "osmosis"
         host_chain: TruncatedChainId,
-        funds: Vec<Coin>,
+        funds: Vec<InterchainSend>,
```
Buckram123, line 99: Feel like it's worth to rename `funds` field, but not sure to what? Maybe name of `SendFunds` as well

framework/contracts/native/ibc-client/src/commands.rs
```diff
@@ -442,25 +443,71 @@ pub fn execute_send_funds(
...
+            }
+        }
+        None => IbcMsg::Transfer {
+            channel_id: ics20_channel_id,
+            to_address: receiver,
+            amount: coin,
+            timeout: env.block.time.plus_seconds(PACKET_LIFETIME).into(),
+        }
+        .into(),
+    }
```
Buckram123, line 508: If we don't have memo we just send cosmwasm ibc message instead of stargate. Did it this way so tests can still use this method

framework/packages/abstract-std/src/native/ibc_client.rs
CyberHoward: I'm sorry that I haven't been clear enough. I'd expect the API to be 

```rust
host_chain: TruncatedChainId,
funds: Vec<Coin>,
memo: Option<String>,
```

If memo is provided, it is attached to **every** ibc transfer msg.

CyberHoward: Because that is what the next version of ics20 will support!

Buckram123: > I'm sorry that I haven't been clear enough. I'd expect the API to be
> 
> ```rust
> host_chain: TruncatedChainId,
> funds: Vec<Coin>,
> memo: Option<String>,
> ```
> 
> If memo is provided, it is attached to **every** ibc transfer msg.

Updated!

CyberHoward (approved):

PR #408, opened 2024-07-17
Fix oversized dex contracts

Buckram123:
- Provide liquidity symmetric no longer supported on Dex Adapters
- Dex Ans Action from Dex Adapters: now resolve should be done by the caller of the Dex Adapter

### Checklist

- [ ] CI is green.
- [x] Changelog updated.

P.S. osmosis test tube passing with fresh wasms, but didn't push them to avoid merge conflicts


modules/contracts/adapters/dex/src/exchanges/junoswap.rs
CyberHoward, line 1: we removed junoswap from match but still have it in here (with a wynd flag?)

We should get rid of this file if we drop junoswap and wynd

CyberHoward (approved):

PR #416, opened 2024-07-24
Abstract sdk feedback

Buckram123:
- Added `execute_with_funds` to Executor to attach funds to execution.
- Account action on executor takes `impl IntoIter<Item = impl Into<AccountAction>>` instead of `Vec<AccountAction>`

### Checklist

- [ ] CI is green.
- [x] Changelog updated.


modules/contracts/adapters/tendermint-staking/src/contract.rs
```diff
@@ -76,27 +75,23 @@ pub fn handle_request(
                     &deps.querier,
                     &source_validator,
                     &destination_validator,
-                    adapter.target()?,
+                    module.target()?,
                 )?,
             };
-            executor.execute(vec![redelegate_msg.into()])
+            executor.execute(iter::once(redelegate_msg))
```
CyberHoward, line 81: hmm. I thought if we accepted that new type we could just provide `CosmosMsg` directly. Is there no way to enable that? I.e. accept both a single item and a vec.

cyberhoward (approved):

modules/contracts/adapters/tendermint-staking/src/contract.rs
CyberHoward: Then does it make sense to take iterator as arg? Kinda weird for devs to have to create an iterator manually

Buckram123: It can be any item that can be converted to iterator like array/vector/hashset

CyberHoward: Yeah but if you only want to provide one message it's kinda weird no?

CyberHoward: or should we differentiate between adding one message and adding multiple? Like what CosmWasm does.

`executor.execute_msg` and `executor.execute_msgs`?

Buckram123: Yeah it's slightly weird(but imo reasonable). Current change already simplifies usage of Executor API, not sure if having distinct method for single/multiple makes it easier or more confusing in our case, as we don't push messages to response, but just return `CosmosMsg`

Buckram123: Maybe it's easier, but IMO this way it's clearer intention.

Putting preferences aside `[]` is bigger than `iter::once()` or `Some()` when used with .into_iter(), source: https://users.rust-lang.org/t/psa-iter-once-can-now-be-spelled-as/77392/7

cyberhoward (approved):

PR #418, opened 2024-07-24
Use private grpc url in ci

Buckram123:
This PR aims to use private grpc networks config for interchain tests

Unspecified ChainKind ended up being pretty breaking change, because some of our dependencies was still using cw-orch 0.22 or 0.23.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Buckram123:
Nice, now it fails on expected action

CyberHoward:
fixed clippy

PR #419, opened 2024-07-25
Resolve feedback abstract client

Buckram123:
### Checklist
- [x] `install_on_sub_account` should be false by default on account builder
- [x] If account is found by builder, let it reflect the requested state instead of skipping entirely. For example when an account is fetched with a namespace, the builder will load the account and stop there. Instead, it should compare the loaded account's state (which modules it has) with the requested modules and install any modules that should be installed ~~and remove any modules that aren't in the builder request~~.
- [x] fn module_installed(module: ModuleId) on account. Returns true if a module is installed, false if it isnt.
- [x] fn `is_module_version_installed(module: ModuleInfo)` on account. Returns true if a module with the specified version is installed, false if it isnt.
- [x] fn with_modules on account builder that installs the provided modules on build.
- [x] Review @feedback comments on [github.com/abstractsdk/icaa-scripts](http://github.com/abstractsdk/icaa-scripts) 
- [x] Upgrade modules on account. (?) Do we want to migrate modules through their `Application<ModuleI>` interfaces or through the Account?
- [x]  fn `ibc_enabled()` helper on account builder (and default it to true?) @CyberHoward I don't like this idea since ibc-client is just a module and you can have it as dependency, don't think we want it to behave not like a module
- [x] `is_published` fn on `Application` @CyberHoward what is the point of it? If you got `Application` object it means you already have it deployed
- [ ] CI is green.
- [x] Changelog updated.


CyberHoward:
- Enabling IBC is something that devs will want to do frequently (we've been doing it all the time lately) so IMO it's fine for there to be a helper to enable it. We can prevent conflicts by only adding the ibc-client to the to-be-installed modules if it's not already in the list. 
- agree, let's not add `is_published`. Tho maybe we can add a similar type of function to the client? `client.module_is_published(module: ModuleId)`?

framework/packages/abstract-client/src/account.rs
```diff
@@ -293,7 +326,46 @@ impl<'a, Chain: CwEnv> AccountBuilder<'a, Chain> {
...
+        Ok(Account::new(abstract_account, self.install_on_sub_account))
+    }
+
+    /// Modules to install without duplicates
+    fn install_modules(&self) -> Vec<ModuleInstallConfig> {
+        let mut install_modules = self.install_modules.clone();
+        if self.ibc_enable.unwrap_or(true) {
+            install_modules.push(IbcClient::<Chain>::install_config(&Empty {}).unwrap());
+        }
+        install_modules.dedup();
```
CyberHoward, line 358: This dedup might not be enough as it will keep two modules of the same kind but different versions. I.e. if I add two DCA apps with different versions, the dedup won't remove one of them.

cyberhoward (approved):

framework/packages/abstract-client/src/account.rs
Buckram123: Also it will fail on same module with same version if init message is different. Wasn't aiming to solve all issues of installation, just this basic one

PR #424, opened 2024-08-02
Abstract-interface should have only non-wasm32 dependencies

Buckram123:


framework/packages/standards/dex/Cargo.toml
```diff
@@ -39,8 +39,6 @@ abstract-adapter = { version = "0.23.0", path = "../../abstract-adapter" }
 
 ibc-chain-registry = { version = "0.25.0", optional = true }
 
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-
```
Buckram123, line 43: This compile will be 0 seconds on dex standard

framework/packages/abstract-interface/Cargo.toml
```diff
@@ -12,7 +12,7 @@ daemon = ["cw-orch/daemon"]
 integration = []
 interchain = ["dep:cw-orch-interchain", "dep:cw-orch-polytone"]
 
-[dependencies]
```
CyberHoward, line 15: won't this fail to compile without an architecture flag in the library as well? (in lib.rs)

framework/packages/standards/dex/Cargo.toml
```diff
@@ -39,8 +39,6 @@ abstract-adapter = { version = "0.23.0", path = "../../abstract-adapter" }
 
 ibc-chain-registry = { version = "0.25.0", optional = true }
```
CyberHoward, line 40: We should be able to remove this dep

framework/packages/abstract-interface/Cargo.toml
Buckram123: > won't this fail to compile without an architecture flag in the library as well? (in lib.rs)

We already have it!
https://github.com/AbstractSDK/abstract/blob/e5f742d505b3934b2c83223b7464980b07fff47d/framework/packages/abstract-interface/src/lib.rs#L1

CyberHoward (approved):

PR #426, opened 2024-08-07
Check dependencies on publishing modules

Buckram123:
This PR adds check before publishing module via AppDeployer/AdapterDeployer/StandaloneDeployer of the module to ensure dependencies also published.

Beyond that this PR removes `cw-semver` dependency and replaces it with `semver`, making it breaking
### Checklist

- [ ] CI is green.
- [x] Changelog updated.


CyberHoward (approved):
LGTM!

PR #427, opened 2024-08-07
Register polytone connection on ibc infrastructure

Buckram123:
This PR adds `register_infrastructure()` method on `AbstractIbc`, that can be used to easily register all of the infrastructures to this chain 
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


CyberHoward:
> @CyberHoward not sure where we're supposed to get this data, should we add `connected_chains: Vec<Chain>` or `Vec<(chain_id, host, note)>` to DeployData?
> 
> ### Checklist
> * [ ]  CI is green.
> * [ ]  Changelog updated.

Is `cw-orch-polytone` not a dependency of this crate? If it is we can use the `Polytone::load_from` fn to get the deployment afaik.

framework/packages/abstract-interface/src/migrate.rs
```diff
@@ -323,7 +324,13 @@ impl<Chain: CwEnv> AbstractIbc<Chain> {
                 .expect("IBC host supposed to be migrated, but skipped instead");
         } else {
             // Version change is breaking, need to deploy new version
```
CyberHoward, line 326: I've actually come around on this issue, I don't think this solution works will with our workflow. When we release a breaking version of our contracts that doesn't always mean we break the API of the contract. Also, the fact that this contract stores the counterparty contract means that when "upgrading" users lose their account on the other chain. 

Could you think of any way for us to improve this? 

I'm thinking we could stabilize the API and we publish breaking versions under a different module-id, but maybe that's weird as well?

Buckram123:
> > @CyberHoward not sure where we're supposed to get this data, should we add `connected_chains: Vec<Chain>` or `Vec<(chain_id, host, note)>` to DeployData?
> > ### Checklist
> > 
> > * [ ]   CI is green.
> > * [ ]   Changelog updated.
> 
> Is `cw-orch-polytone` not a dependency of this crate? If it is we can use the `Polytone::load_from` fn to get the deployment afaik.

We need dst chain(or chains) to know all of the data, note address defaults to null in cw-orch state without dst chain.
```json
{ "polytone:note": null }
 ```

File: framework/packages/abstract-interface/src/migrate.rs (hunk 1 above)
Comment (Buckram123) [reply framework/packages/abstract-interface/src/migrate.rs:326]: > I've actually come around on this issue, I don't think this solution works will with our workflow. When we release a breaking version of our contracts that doesn't always mean we break the API of the contract. Also, the fact that this contract stores the counterparty contract means that when "upgrading" users lose their account on the other chain.

That's why we shouldn't make breaking bump to ibc modules without breaking change!

> Could you think of any way for us to improve this?

Pretty sure if proxy contract will own data about remote accounts then update will become much easier

**Comment by CyberHoward** [comment]
> > > @CyberHoward not sure where we're supposed to get this data, should we add `connected_chains: Vec<Chain>` or `Vec<(chain_id, host, note)>` to DeployData?
> > > ### Checklist
> > > 
> > > * [ ]    CI is green.
> > > * [ ]    Changelog updated.
> > 
> > 
> > Is `cw-orch-polytone` not a dependency of this crate? If it is we can use the `Polytone::load_from` fn to get the deployment afaik.
> 
> We need dst chain(or chains) to know all of the data, note address defaults to null in cw-orch state without dst chain.
> 
> ```json
> { "polytone:note": null }
> ```

We should be able to get the whole state and find all connections to the current state right?

File: framework/packages/abstract-interface/src/ibc.rs [hunk 2]
```diff
@@ -59,14 +59,33 @@ impl<Chain: CwEnv> AbstractIbc<Chain> {
             ),
         ])
     }
+
+    #[cfg(feature = "interchain")]
+    /// Register infrastructure on connected chains
+    ///
+    /// Note: this should only be called on real chains (for example using [`cw_orch::daemon::Daemon`])
```
CyberHoward, line 66: If so, should it not be implemented for `Abstract<Daemon>` instead of `Abstract<CwEnv>`?

PR #428, opened 2024-08-08
Forward stargate feature for abstract modules

Buckram123:
You can see here: https://docs.rs/crate/abstract-app/0.23.0/features
That we don't have a way to enable "abstract-sdk/stargate" when only using abstract-app as a dependency. This PR adds `stargate` feature for module packages
### Checklist

- [ ] CI is green.
- [x] Changelog updated.


CyberHoward (approved):

CyberHoward:
nice!

PR #430, opened 2024-08-09
Allow developers to register contracts as service contracts

Buckram123:
See ABS-485
### Checklist

- [ ] CI is green.
- [x] Changelog updated.

### DIFF
--- a/framework/docs/src/releases/CHANGELOG.md
+++ b/framework/docs/src/releases/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - `execute_with_funds` to Executor to attach funds to execution.
+- New module type: `Service`, acts the same as Native, but can be registered by any namespace
 
 ### Changed
 
--- a/framework/docs/src/releases/CHANGELOG.md
+++ b/framework/docs/src/releases/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - `execute_with_funds` to Executor to attach funds to execution.
 - New module type: `Service`, acts the same as Native, but can be registered by any namespace
+- `AbstractClient`: `service` to get api of Service module
 
 ### Changed
 
--- a/framework/packages/abstract-client/src/account.rs
+++ b/framework/packages/abstract-client/src/account.rs
@@ -870,15 +870,8 @@ impl<Chain: CwEnv> Account<Chain> {
                 .manager
                 .install_modules(modules, Some(funds))?;
         }
-        let module_addresses = self.module_addresses(vec![M::module_id().to_owned()])?;
 
-        let contract = Contract::new(
-            M::installed_module_contract_id(&self.id()?),
-            self.environment(),
-        );
-        contract.set_address(&module_addresses.modules[0].1);
-
-        let module: M = contract.into();
+        let module = self.module::<M>()?;
 
         Application::new(
             Account::new(self.abstr_account.clone(), self.install_on_sub_account),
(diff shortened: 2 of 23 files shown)


framework/packages/abstract-client/src/account.rs
```diff
@@ -927,30 +961,6 @@ impl<Chain: CwEnv> Account<Chain> {
...
-        // When there are multiple modules registered the addresses are returned in a common
-        // separated list. We want the last one as that is the "top-level" module while the rest
-        // are dependencies, since in the sub-account creation call, we pass in the top-level
-        // module last.
-        let module_address = module_addresses.split(',').last().unwrap();
-
-        // We install only one module
-        Addr::unchecked(module_address)
-    }
-
```
Buckram123, line 953: This method is not useful in combination with removing duplicate modules from install list, so got replaced to the more reliable method

framework/docs/src/releases/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### Added
 
 - `execute_with_funds` to Executor to attach funds to execution.
+- New module type: `Service`, acts the same as Native, but can be registered by any namespace
```
CyberHoward, line 8:
```suggestion
- New module type: `Service`, behaves the same as Native, but can be registered by any namespace.
```

framework/packages/abstract-client/src/account.rs
```diff
@@ -521,6 +539,22 @@ impl<Chain: CwEnv> Account<Chain> {
         }
     }
 
+    /// Install an service on the account.
+    /// if `install_on_sub_account` is `true`, the application will be installed on new a sub-account. (default)
```
CyberHoward, line 543: We disabled the sub-account install by default.
```suggestion
    /// if `install_on_sub_account` is `true`, the application will be installed on new a sub-account.
```

framework/packages/abstract-client/src/account.rs
```diff
@@ -831,24 +865,24 @@ impl<Chain: CwEnv> Account<Chain> {
...
-            .manager
-            .install_modules(modules, Some(funds))?;
-
-        let module_addr = Self::parse_modules_installing_response(install_module_response);
         let contract = Contract::new(
             M::installed_module_contract_id(&self.id()?),
             self.environment(),
         );
-        contract.set_address(&module_addr);
+        contract.set_address(&module_addresses.modules[0].1);
```
CyberHoward, line 879: is this guaranteed to be index 0 on this array?

Buckram123: Yeah, we error when we get no modules
https://github.com/AbstractSDK/abstract/blob/6a6abf82ca05bdefc6f21f58de31bc7ffb302452/framework/contracts/account/manager/src/queries.rs#L187-L191

framework/packages/abstract-interface/src/deployers.rs
```diff
@@ -242,3 +242,58 @@ pub trait StandaloneDeployer<Chain: CwEnv>: Sized + Uploadable + ContractInstanc
         Ok(())
     }
 }
+
+/// Trait for deploying Services
+pub trait ServiceDeployer<Chain: CwEnv>:
+    Sized + Uploadable + ContractInstance<Chain> + CwOrchInstantiate<Chain>
+{
+    /// Deploys the module. If the module is already deployed, it will return an error.
+    /// Use `maybe_deploy` if you want to deploy the module only if it is not already deployed.
```
CyberHoward, line 251: there is no `maybe_deploy` fn, that's what the strategies are for right?

CyberHoward (approved):

framework/packages/abstract-client/src/account.rs
Buckram123: Updated this query method on manager!

PR #431, opened 2024-08-12
Support interchain ICA

CyberHoward:
This PR adds an `ica-client` contract that handles ICA management through the proxy contract. 

Any ICA action is performed by providing the proxy contract with a correctly formed `ica_client::QueryMsg::IcaAction` query. The result of this query is a `Vec<CosmosMsg>` that is executed by the proxy.

### Checklist
- [x] Update deps to use published polytone-evm crate
- [x] Improve test coverage
- [ ] CI is green.
- [x] Changelog updated.


framework/packages/abstract-ica/src/chain_type.rs
```diff
@@ -0,0 +1,55 @@
+use std::fmt::Display;
+
+use abstract_sdk::std::objects::TruncatedChainId;
+use abstract_sdk::std::registry::*;
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum ChainType {
+    Evm,
+    Cosmos,
+}
```
Buckram123, line 10: Shouldn't it be: EVM+CosmwasmVM or Eth+Cosmos?

Buckram123 (approved):
LGTM

framework/packages/abstract-ica/src/chain_type.rs
Buckram123: > Not sure what you mean by this

Isn't EVM run SC code on ethereum chains and Cosmwasm VM on cosmos chains?

CyberHoward: THe `Cosmos` chain type will be for cw-ica-controller (so ICA). We can add CosmWasm in the future as well if we want (and use polytone straight up).

framework/packages/abstract-ica/src/action.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/// Queries first
+/// Execute second
+/// Funds transfers last
+impl PartialOrd for IcaAction {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
Buckram123, line 42: I see 3 definitions for order here:
1) Execute first, Fund transfers second, Query last
2) Queries first, Execute second, Funds transfers last
(third will be in a next comment)

framework/contracts/native/ica-client/src/queries.rs
```diff
@@ -0,0 +1,278 @@
...
+    })?;
+
+    // todo: what do we do for msgs that contain both cosmos and EVM messages?
+    // Best to err if there's conflict.
+
+    // sort actions
+    // 1) Transfers
+    // 2) Calls
+    // 3) Queries
+    actions.sort_unstable();
```
Buckram123, line 42: (see previous comment)
3) Fund transfers first, Execute second, Query last

I'm confused, which order should it be?

Beyond that, is it okay to mix order inside of the group? Won't it be better to create a new vector to which we push first group of actions, second group of actions, ...?

Buckram123 (commented):
How actions should be sorted?

framework/contracts/native/ica-client/src/queries.rs
CyberHoward: I'm following this logic: https://github.com/srdtrk/cw-ica-controller/blob/44962d06a69852176b9417caa3326090af548aaf/src/types/msg.rs#L51

I.e. execute queries after calls, which makes sense to me because then you can easily request some post-action query data (if relevant). 

The problem is that some implementations will support this while others wont. For example, with cw-ica-controller we know that the actions and queries will be executed in that order, but for other solutions (like our current polytone implementation) we don't know if the execution will be performed before the queries because they are two different packets...

Buckram123: I don't get the point of sorting it then at all

CyberHoward: Problem is that cw-ica-controller sorts them so we either need to do the same or we risk confusing devs.

framework/contracts/native/ica-client/src/queries.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    let cfg = CONFIG.load(deps.storage)?;
+
+    // TODO: should we match ChainType first, so it doesn't get so nested?
```
adairrr, line 46: Yes because then each ChainType's functionality can be separated into functions more clearly as well.

framework/contracts/native/ica-client/README.md
```diff
@@ -0,0 +1,39 @@
+# Abstract ICA Client
+
+The ICA client contract provides an interface for interacting with Interchain Accounts (ICAs) on different blockchains using the IBC protocol.
```
adairrr, line 3: What we have with the ibc-client and host is something similar to the "controlled note" because they form a protocol built upon Polytone as a messaging layer. 

Is the ICA-client going to be separate from the ibc-client and host and be more of a "protocol-level" contract? Asking because if the user calls through Polytone with this they'll have a separate Polytone Proxy on the other chain. Or is the ibc-client ultimately going to use this?

framework/contracts/native/ica-client/src/queries.rs
CyberHoward: The type of chain is inside of the action so then assertions need to be moved. Let's refactor it once we support another ICA impl, then it'll be clearer how to structure it.

framework/packages/abstract-ica/src/action.rs
CyberHoward: Removed sorting

framework/contracts/native/ica-client/src/queries.rs
CyberHoward: Removed the sorting for now.

framework/contracts/native/ica-client/README.md
CyberHoward: The ibc-client could use this to interact with its polytone instance yes (if we added polytone as an ICA provider)

This is basically an ICA protocol adapter.

adairrr: Gotcha. So are we imagining that we'll create a new `evm-ibc-client` or execute EVM ica actions through this one? 

I suppose it relates to the question whether we'll support "ibc-host"-like functionality on the EVM side.

CyberHoward: We execute EVM ICA actions "through" this adapter. This contract only constructs messages based on a query. 

ibc-host-like functionality is a different game altogether and would broaden the scope of our EVM work significantly. Ideally we update the proxy implementation to use some available smart-account solution. 

It's too easy to assume that certain patterns in CW also make sense in Solidity. The complexity of building an abstract-like setup on EVM (which would be a requirement to supporting ibc-host functionality) is probably an order of magnitude harder than CW. 

I really think we should keep playing on our strengths (CW) and keep the EVM side either as minimal as possible (with some additional runtime overhead) or finding an EVM SA provider that is willing to play this game with us.

PR #432, opened 2024-08-13
Remove value calculation logic from proxy contract

Buckram123:
This PR removes everything that's related to the calculation logic in proxy

ETF was relying on that, so it got removed as well as mentioned in ABS-488 description
### Checklist

- [x] CI is green.
- [x] Changelog updated.


framework/contracts/account/proxy/src/queries.rs
```diff
@@ -68,36 +26,6 @@ pub fn query_config(deps: Deps) -> StdResult<ConfigResponse> {
...
-        .map_err(Into::into)
-}
-
-pub fn query_base_asset(deps: Deps) -> ProxyResult<BaseAssetResponse> {
-    let oracle = Oracle::new();
-    let base_asset = oracle.base_asset(deps)?;
-    Ok(BaseAssetResponse { base_asset })
-}
-
 pub fn query_holding_amount(
```
CyberHoward, line 29: Should we remove this query as well? 

IMO this can be done by either the FE or the in the abstract-sdk bank API. Doing a seperate smart-query to this contract for this information is a bit weird.

cyberhoward (approved):

framework/contracts/account/proxy/src/queries.rs
Buckram123: Agree! Removing it!

Buckram123: Proxy don't need ANS_HOST anymore, should we remove it from state?

CyberHoward: > Proxy don't need ANS_HOST anymore, should we remove it from state?

Yup! (we should add that to the migrate logic as well then)

CyberHoward: we should probably also remove the migrate logic from the manager that was required for the last version?

Buckram123: Tbh, now I REALLY wonder that if we merge manager and proxy, even manager will become smaller(because it have to Serialize ProxyMsg for Adding/Removing modules)

CyberHoward: it's pretty easy to figure out, 

- create new contract
- define the new msgs as 
```rust
ExecuteMsg {
   Proxy(proxy::ExecuteMsg),
   Manager(...),
}
```

And straight up delegate the logic to those in the match statement: 

```rust
fn execute() {
   match msg {
      Proxy(msg) => proxy::contract::execute(msg,...)
      ...
   }
}
```

and then compile it!

PR #436, opened 2024-08-29
Wasm-flag workspace hack for framework

CyberHoward:
Add a wasm-target flag to prevent the workspace hack dependencies from being included in wasm builds


Buckram123 (approved):
hakari still not happy with excluded ica and macros

CyberHoward:
Don't see a way to allow them to be excluded in the command.

CyberHoward:
Maybe this? 
https://docs.rs/cargo-hakari/0.9.30/cargo_hakari/config/index.html#final-excludes

PR #440, opened 2024-09-06
Remove useless objects from abstract-std

Buckram123:
Removed unused objects from abstract-std.

In a future it would make sense to split objects used inside abstract-std into separate packages. To give an example of such objects: `TimeWeightedAverage`, `GovernanceDetails`, `SimpleVoting` and some other modules  
### Checklist

- [ ] CI is green.
- [x] Changelog updated.


framework/packages/abstract-std/src/objects/paged_map.rs
CyberHoward, line 1: I was so proud of this one 😿

CyberHoward (approved):

framework/packages/abstract-std/src/objects/paged_map.rs
Buckram123: Looks cool, but it's hard to think where it can be useful on-chain. What was it used for?

CyberHoward: For batched execution. I.e. if you need to migrate the state of a Map for example, then you'll need do split it up into multiple txs.

PR #441, opened 2024-09-09
Pull out used types of cw721

Buckram123:
This PR aims to pull out used types for nft ownership from cw721 package.
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-std/src/objects/ownership/cw721.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub expires: Expiration,
+}
+
+#[cw_serde]
+pub struct OwnerOfResponse {
+    /// Owner of the token
+    pub owner: String,
+    /// If set this address is approved to transfer/send the token as well
+    pub approvals: Vec<Approval>,
+}
```
Buckram123, line 31:
```suggestion
#[cw_serde]
pub struct OwnerOfResponse {
    /// Owner of the token
    pub owner: String,
}
```
We only care about owner right now, should we skip deserializing approvals?

Buckram123: BTW https://github.com/CosmWasm/cosmwasm/blob/3b15bfba6406cac2f807b9af93077c04d27e79c5/CHANGELOG.md?plain=1#L157

CyberHoward: ahhh nice! Yeah then we can keep it out but leave it in as a commented value. 
```rust
pub struct OwnerOfResponse {
	    /// Owner of the token
	    pub owner: String,
	    /// If set this address is approved to transfer/send the token as well
	    // Left out as an optimization.
	    // pub approvals: Vec<Approval>,
	}
```

CyberHoward (approved):

framework/packages/abstract-std/src/objects/ownership/cw721.rs
Buckram123: Sure!

PR #442, opened 2024-09-09
Pull out polytone

Buckram123:
This PR aims to pull out used types for polytone from polytone and polytone_note packages.
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-std/src/native/ibc.rs
```diff
@@ -12,6 +13,7 @@ use crate::{
     ibc_client,
     objects::{module::ModuleInfo, TruncatedChainId},
 };
+use polytone_callbacks::{Callback as PolytoneCallback, ErrorResponse, ExecutionResponse};
```
Kayanski, line 16: Whats'up here ?

Kayanski:
Very nice, could you add references to the code copied (with URLs for instance), because we might/will touch on this later and for reference ?

framework/packages/abstract-std/src/native/ibc.rs
Buckram123: What do you mean?

Buckram123:
> Very nice, could you add references to the code copied (with URLs for instance), because we might/will touch on this later and for reference ?

Added links to docs.rs of copied types

CyberHoward (approved):

PR #443, opened 2024-09-09
Optimize account state

Buckram123:
See ABS-499
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/packages/abstract-std/src/core/manager.rs
+++ b/framework/packages/abstract-std/src/core/manager.rs
@@ -36,7 +36,9 @@ pub mod state {
     #[cosmwasm_schema::cw_serde]
     pub struct AccountInfo {
         pub name: String,
+        #[serde(skip_serializing_if = "Option::is_none")]
         pub description: Option<String>,
+        #[serde(skip_serializing_if = "Option::is_none")]
         pub link: Option<String>,
     }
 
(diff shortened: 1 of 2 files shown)


framework/packages/abstract-std/src/core/manager.rs
```diff
@@ -47,9 +49,11 @@ pub mod state {
         pub const ACCOUNT_MODULES: &str = "d";
         pub const DEPENDENTS: &str = "e";
         pub const SUB_ACCOUNTS: &str = "f";
+        pub const STATE: &str = "g";
     }
 
     /// Suspension status
+    // TODO: Pull it inside Config as `suspended: Option<String>`, with reason of suspension inside a string?
     pub const SUSPENSION_STATUS: Item<SuspensionStatus> = Item::new(namespace::SUSPENSION_STATUS);
```
Buckram123, line 57: @CyberHoward What do you think?

CyberHoward: Think bool is fine for now. We'll want to sync on this functionality with XION to see if they even want it.

Buckram123: > Think bool is fine for now. We'll want to sync on this functionality with XION to see if they even want it.

It can be `Option<bool>` as well, I just don't get why we want to have it in the state when most of the time account won't be suspended

Buckram123: *inside Config if option

CyberHoward: True, we can pull it out into an optional that we don't save if not set!

PR #444, opened 2024-09-09
Remove proxy and manager contracts and packages

Kayanski:
This PR aims at removing the manager and proxy contracts completely from the repository

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-std/src/account/state.rs
```diff
@@ -0,0 +1,57 @@
...
+pub const ACCOUNT_MODULES: Map<ModuleId, Addr> = Map::new(namespace::ACCOUNT_MODULES);
+/// Stores the dependency relationship between modules
+/// map module -> modules that depend on module.
+pub const DEPENDENTS: Map<ModuleId, HashSet<String>> = Map::new(namespace::DEPENDENTS);
+/// List of sub-accounts
+pub const SUB_ACCOUNTS: Map<u32, cosmwasm_std::Empty> = Map::new(namespace::SUB_ACCOUNTS);
+// Additional states, not listed here: cw_gov_ownable::GovOwnership
+
+#[cosmwasm_schema::cw_serde]
+pub struct State {
```
CyberHoward, line 55: We could reduce this state by storing `State(Vec<Addr>)` or even just storing `Item<Vec<Addr>>`?

CyberHoward: (As context @Kayanski, XION wants us to really minimize the state that each account takes in, because they plan on having **a lot** of accounts.

CyberHoward: So in this case the `state` will be part of the serialized data `{"state": ["xyz"]}` vs `["xzy"]`

Buckram123: I would prefer to rename `STATE`(and `State`, if we decide to keep it this type) to something else, it doesn't give any information about what this list used for

framework/packages/abstract-std/src/account/state.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub const SUSPENSION_STATUS: &str = "a";
+    pub const CONFIG: &str = "b";
+    pub const INFO: &str = "c";
+    pub const ACCOUNT_MODULES: &str = "d";
+    pub const DEPENDENTS: &str = "e";
+    pub const SUB_ACCOUNTS: &str = "f";
+}
+
+pub const STATE: Item<State> = Item::new("a");
+pub const ADMIN: Admin = Admin::new(ADMIN_NAMESPACE);
```
Buckram123, line 37: Put `STATE` namespace inside `namespaces`, you have 2 "a"

Buckram123 (commented):
I'm not sure why we have `msgs`, `types` and `responses` in separate files for the account? Is it something we want to follow up with other contracts as well?

CyberHoward:
> I'm not sure why we have `msgs`, `types` and `responses` in separate files for the account? Is it something we want to follow with other contracts as well?

I kind of like the current convention we have as well. 

```rust
pub mod state {
   // Constants 
   
   // Types 
}

// Messages

// Responses
```

framework/packages/abstract-std/src/account/state.rs
Kayanski: Renamed to WHITELISTED_MODULES instead here

framework/packages/abstract-sdk/src/base/features/executor.rs
```diff
@@ -20,3 +20,5 @@ pub trait AccountExecutor: AccountIdentification {
             .map_err(Into::into)
     }
 }
+
+impl<T> AccountExecutor for T where T: AccountIdentification {}
```
Kayanski, line 24: I think this is interesting to add, so added here. There is no reason to re-implement that explicitely on structures that have AccountIdentification

Buckram123: Thanks for bringing attention to that change, unfortunately I think we would want to revert it. The point of implementing that explicitly is to limit which contract can and which cannot execute as account. Unless we have a better way to limit it with types?

Kayanski: I'm not sure what you mean here. That `Standalone` can't execute as the account ?

framework/contracts/account/src/reply.rs
```diff
@@ -0,0 +1,70 @@
...
+        .set_data(data)
+    } else {
+        AccountResponse::new(
+            "forward_response_data_reply",
+            vec![("response_data", "false")],
+        )
+    };
+
+    Ok(resp)
+}
```
Buckram123, line 31: @CyberHoward do we even need this method in cosmwasm 2? Won't it be possible to get data from the new field of SubMsgResponse: https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.SubMsgResponse.html#structfield.msg_responses? Let's mark this method as deprecated for reminder as well if so

Kayanski: I think it's the same behavior here, because it's still a nested execution

framework/packages/abstract-adapter/src/features.rs
```diff
@@ -32,11 +30,6 @@ impl<Error: ContractError, CustomInitMsg, CustomExecMsg, CustomQueryMsg, SudoMsg
     }
 }
 
-impl<Error: ContractError, CustomInitMsg, CustomExecMsg, CustomQueryMsg, SudoMsg> AccountExecutor
```
CyberHoward, line 35: Why was this removed?

Buckram123: https://github.com/AbstractSDK/abstract/pull/444#discussion_r1756393019 Check this discussion

framework/packages/abstract-sdk/src/base/features/executor.rs
CyberHoward: Yup, correct. We want to implement AccountExecutor on App and Adapter but not implement it for Standalone because in the install flow a standalone is not added to the permission set of account.

framework/packages/abstract-app/src/features.rs
```diff
@@ -43,18 +41,6 @@ impl<
...
 
-impl<
-        Error: ContractError,
-        CustomInitMsg,
-        CustomExecMsg,
-        CustomQueryMsg,
-        CustomMigrateMsg,
-        SudoMsg,
-    > AccountExecutor
-    for AppContract<Error, CustomInitMsg, CustomExecMsg, CustomQueryMsg, CustomMigrateMsg, SudoMsg>
```
CyberHoward, line 54: We should keep this.

framework/contracts/account/src/reply.rs
Kayanski: Asked for clarification here : https://github.com/CosmWasm/cosmwasm/issues/2253

framework/packages/abstract-sdk/src/mock_module.rs
```diff
@@ -26,8 +26,6 @@ impl AccountIdentification for MockModule {
     }
 }
 
-impl AccountExecutor for MockModule {}
```
CyberHoward, line 29: keep impl

Kayanski: ok, let me put it back again

framework/contracts/account/src/reply.rs
CyberHoward: > @CyberHoward do we even need this method in cosmwasm 2? Won't it be possible to get data from the new field of SubMsgResponse: https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.SubMsgResponse.html#structfield.msg_responses? Let's mark this method as deprecated for reminder as well if so

Yeah I removed the `allow(deprecated)`

But we should do this in a different PR imo.

PR #446, opened 2024-09-10
Account self registration

CyberHoward:
This PR adds and tests the self-registration logic for new Abstract Accounts.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/contracts/account/account/src/contract.rs
```diff
@@ -54,19 +56,25 @@ pub fn instantiate(
...
+    // ## Manage ##
+    let module_factory_address = deps.api.addr_validate(&module_factory_address)?;
+    let version_control_address = deps.api.addr_validate(&version_control_address)?;
+
+    let account_id = account_id.unwrap_or_else(|| {
+        AccountId::local(
+            LOCAL_ACCOUNT_SEQUENCE
+                .query(&deps.querier, version_control_address.clone())
+                // default to 0 for first account
+                .unwrap_or_default(),
```
Kayanski, line 68: Isn't it dangerous that any error will turn the account sequence to 0 on instantiation ? I think we should be more careful on the error handling here !

Kayanski: This could mean parsing the error or simply initializing the account sequence on the version control on instantiation

framework/contracts/native/version-control/src/commands.rs
```diff
@@ -31,32 +35,95 @@ use crate::{
...
 /// Add new Account to version control contract
-/// Only Factory can add Account
+/// Only Account can add itself.
 pub fn add_account(
     deps: DepsMut,
     msg_info: MessageInfo,
-    account_id: AccountId,
-    account_base: Account,
     namespace: Option<String>,
+    creator: String,
```
Kayanski, line 43: Creator is never used, but to check that it's the ibc-host in case of remote accounts. So this seems to be a useless parameter

Kayanski (changes requested):
very nice ! A little change on the creator parameter that seems useless

framework/contracts/account/account/src/contract.rs
CyberHoward: Think that second option makes the most sense :) good comment

framework/contracts/native/version-control/src/commands.rs
CyberHoward: How do we check that the ibc-host created the account if not for this parameter?

framework/contracts/account/account/src/contract.rs
CyberHoward: Done

framework/contracts/account/account/Cargo.toml
```diff
@@ -34,11 +32,15 @@ abstract-sdk = { workspace = true }
 abstract-std = { workspace = true }
 semver = { workspace = true }
 abstract-macros = { workspace = true }
-workspace-hack = { version = "0.1", path = "../../../workspace-hack" }
+cw721 = { version = "0.18.0" }
+cw721-base = { version = "0.18.0", features = ["library"] }
```
Buckram123, line 36: Use/add types from `framework/packages/abstract-std/src/objects/ownership/cw721.rs` instead, or put it only in dev-deps if you need it for testing
```suggestion
```

framework/contracts/native/version-control/src/commands.rs
```diff
@@ -31,32 +35,95 @@ use crate::{
...
 
-    ACCOUNT_ADDRESSES.save(deps.storage, &account_id, &account_base)?;
+    let acc_module_info = maybe_acc_module_info;
+
+    let account_code_id = REGISTERED_MODULES
+        .load(deps.storage, &acc_module_info)?
+        .unwrap_account()?;
+
+    // provided and smaller, assert is eq to sequence
+    // provided and larger, just register
```
Buckram123, line 72: This logic should be removed if we can't create account without instantiate2, and have sequences for all accounts

framework/contracts/native/version-control/src/commands.rs
```diff
@@ -31,32 +35,95 @@ use crate::{
...
+
+    ensure_eq!(
+        account_code_id,
+        sender_contract_info.code_id,
+        VCError::NotAccountCodeId {
+            account_info: acc_module_info,
+            expected_code_id: account_code_id,
+            actual_code_id: sender_contract_info.code_id
+        }
+    );
```
Buckram123, line 121: Let's move all validations to the start of the method

framework/packages/abstract-testing/src/lib.rs
```diff
@@ -3,16 +3,138 @@ pub mod map_tester;
...
+        } else {
+            Err(format!(
+                "attempt to query {} with key {}",
+                contract, str_key
+            ))
+        }
+    };
+    let abstr = AbstractMockAddrs::new(mock_api);
+    // panic!("{}", abstr.account.addr().to_string());
+    // cosmwasm1s0t3nh7snq5xz2nfkgqdd4wzcnm8d2x7ercmqf9v4rw8du96ud2qe23ef4
```
Buckram123, line 55:
```suggestion
```
It's true only when `cosmwasm` is prefix in `MockApi`

framework/contracts/native/version-control/src/commands.rs
```diff
@@ -31,32 +35,95 @@ use crate::{
...
         !ACCOUNT_ADDRESSES.has(deps.storage, &account_id),
         VCError::AccountAlreadyExists(account_id)
     );
 
-    ACCOUNT_ADDRESSES.save(deps.storage, &account_id, &account_base)?;
+    let acc_module_info = maybe_acc_module_info;
+
+    let account_code_id = REGISTERED_MODULES
+        .load(deps.storage, &acc_module_info)?
+        .unwrap_account()?;
```
Buckram123, line 69: What is the benefit of checking anything except cw2 id?
User can migrate from any code id before registration and to any code id after registration, so we can't do anything against custom account contracts

Buckram123 (commented):
Very cool! Should we decrease amount of checks for account registration, or is it helping user during account creation?

framework/packages/abstract-testing/src/lib.rs
CyberHoward: this was left from debugging. :)

framework/contracts/native/version-control/src/commands.rs
CyberHoward: The goal is that they can't. For this we need to ensure that the migration function of the contract is properly protected.
[ORC-169: Ensure Account migration is properly protected](https://linear.app/abstract-sdk/issue/ORC-169/ensure-account-migration-is-properly-protected)

framework/contracts/native/version-control/src/commands.rs
Kayanski: Sorry my bad, this parameter can be trusted from the caller, so it's necessary, removing this comment

framework/contracts/native/version-control/src/commands.rs
CyberHoward: Will do in different PR
[ORC-169: Ensure Account migration is properly protected](https://linear.app/abstract-sdk/issue/ORC-169/ensure-account-migration-is-properly-protected)

framework/contracts/native/version-control/src/commands.rs
CyberHoward: done

framework/contracts/native/version-control/src/commands.rs
Buckram123: And because registration can only happen during account instantiation, it can't be migrated from any contract, so we only need to protect migrate, ok, nice

CyberHoward:
Merge #444 First!!

framework/packages/abstract-client/src/account.rs
Buckram123, line 1: In `AccountBuilder::build` at the end of the method, when `self.owner_account` is None, we should call instantiate on account. How should it be executed?
- Get code id of the "abstract::account" from `chain` and
- `chain.instantiate2()`?
or maybe it would be better to add some method to `AbstractAccount` to instantiate it?

Buckram123: Ok, nice

PR #447, opened 2024-09-11
Merged account and abstractaccount

Kayanski:
This PR aims at merging the 2 account structures into one. 

### Design choices

Merged the two Accounts definitions into a single `AccountI` to indicate that it is a cw-orch interface. 

The `Account` exposed by the `abstract-client` should be the canonical type used by external tests. 

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


CyberHoward:
Interesting! Can't we move all the `impl AbstarctAccount` to `impl Account`? 

and maybe rename `Account` to `AccountI` or something (indicating that it's different from the abstract-client `Account`?

CyberHoward (approved):

PR #448, opened 2024-09-12
V2/Feat/Standardize the deployed contracts

Buckram123:
This PR aims to make abstract native contracts have predictable canonical address between chains. It's done by first uploading blob with defined checksum, creator and salt and migrating to the actual contract right after that
### Checklist

- [ ] CI is green.
- [x] Changelog updated.


framework/packages/abstract-std/src/native_addrs.rs
Buckram123, line 1: This file will have to be re-generated when we decide who is the creator

framework/packages/abstract-std/src/native_addrs.rs
```diff
@@ -0,0 +1,28 @@
...
+// Default local-juno used right now(for testing)
+pub const TEST_ABSTRACT_CREATOR: [u8; 20] = [
+    210, 20, 62, 221, 52, 61, 116, 51, 8, 49, 75, 191, 7, 17, 231, 72, 143, 140, 113, 214,
+];
+
+pub const ANS_ADDR: [u8; 32] = [
+    240, 207, 249, 98, 43, 113, 214, 211, 30, 136, 11, 140, 16, 81, 121, 153, 203, 165, 42, 12,
+    186, 176, 183, 241, 120, 71, 168, 99, 151, 45, 246, 17,
+];
+pub const VERSION_CONTROL_ADDR: [u8; 32] = [
```
Kayanski, line 11: I'm not sure why we need to store all addresses as canonical addresses. Can't we just use instantiate_addr2 using the blob checksum, the creator and the salt. 
In that way, there is only 2 bytes array and N salts to store here ?

framework/contracts/native/ans-host/src/contract.rs
```diff
@@ -103,13 +103,29 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
 }
 
 #[cfg_attr(feature = "export", cosmwasm_std::entry_point)]
-pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> AnsHostResult {
-    let version: Version = CONTRACT_VERSION.parse().unwrap();
+pub fn migrate(deps: DepsMut, env: Env, msg: MigrateMsg) -> AnsHostResult {
+    match msg {
+        MigrateMsg::Instantiate(instantiate_msg) => {
```
Kayanski, line 108: Love that logic hah. Instantiate on Migrate is fun

Kayanski: We should make sure that instantiate is callable only once I think !

framework/packages/abstract-std/src/native_addrs.rs
Buckram123: > I'm not sure why we need to store all addresses as canonical addresses. Can't we just use instantiate_addr2 using the blob checksum, the creator and the salt. In that way, there is only 2 bytes array and N salts to store here ?

We don't need to store it as canonical, but why compute instantiate2 of already known address every time we use any of the native contracts?

framework/contracts/native/ans-host/src/contract.rs
Buckram123: > We should make sure that instantiate is callable only once I think !

Yeah we can add check that there is no cw2 in state. Is it good enough?

framework/contracts/native/ans-host/src/contract.rs
```diff
@@ -103,13 +103,29 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+            let contract_info = deps
+                .querier
+                .query_wasm_contract_info(&env.contract.address)?;
+            // Only admin can call migrate on contract
+            let sender = contract_info.admin.unwrap();
```
CyberHoward, line 113: This is ensured by the VM no?

Buckram123: Yeah, it's just for construction of `MessageInfo`, in case instantiate will need "sender"

CyberHoward: Can we pull this logic into a function in abstract-std?

CyberHoward: Something like `fn migrate_instantiation(init_fn: Fn<(...)> -> ...)`

Buckram123: > Something like `fn migrate_instantiation(init_fn: Fn<(...)> -> ...)`

Yep

CyberHoward (approved):

CyberHoward (approved):

Buckram123:
Some of the doctests failling now

Buckram123:
> Some of the doctests failling now

Not anymore

framework/packages/abstract-testing/src/lib.rs
```diff
@@ -191,6 +196,21 @@ pub mod module {
...
+    pub fn mock_bech32_admin(bech32: &MockBech32) -> Addr {
+        bech32
+            .app
+            .borrow()
+            .api()
+            .addr_humanize(&CanonicalAddr::from(TEST_ABSTRACT_CREATOR))
+            .unwrap()
+    }
+}
+
```
Buckram123, line 213: A bit annoying to refactor, maybe annoying to use? But we need predetermined sender

framework/packages/abstract-interface/src/deployment.rs
```diff
@@ -32,21 +38,25 @@ pub struct Abstract<Chain: CwEnv> {
...
     pub ibc: AbstractIbc<Chain>,
     pub(crate) account: AccountI<Chain>,
+    pub(crate) blob: CwBlob<Chain>,
 }
 
 impl<Chain: CwEnv> Deploy<Chain> for Abstract<Chain> {
     // We don't have a custom error type
     type Error = AbstractInterfaceError;
-    type DeployData = String;
+    type DeployData = Chain::Sender;
```
Buckram123, line 47: Now passing sender to deployment, which should be the admin of the abstract

framework/packages/abstract-client/src/builder.rs
```diff
@@ -128,26 +128,28 @@ impl<Chain: CwEnv> AbstractClientBuilder<Chain> {
...
-    pub fn build(&self) -> AbstractClientResult<AbstractClient<Chain>> {
-        let abstr =
-            Abstract::deploy_on(self.chain.clone(), self.chain.sender_addr().into_string())?;
-        self.update_ans(&abstr)?;
+    pub fn build(
+        &self,
+        abstract_admin: Chain::Sender,
+    ) -> AbstractClientResult<AbstractClient<Chain>> {
+        let abstr = Abstract::deploy_on(self.chain.clone(), abstract_admin.clone())?;
+        self.update_ans(&abstr, abstract_admin)?;
```
Buckram123, line 136: A bit painful, can we avoid it?

Buckram123 (commented):
Highlighted changes after previous review. The rest is applying it on tests

framework/packages/abstract-client/src/builder.rs
Kayanski: what's painful exactly there ?

Buckram123: > what's painful exactly there ?

Passing sender on client builder. I guess it's fine as long as it's easy to get this sender

framework/packages/abstract-testing/src/lib.rs
CyberHoward: Can't we set the sender when deploying the contracts only?

CyberHoward: I.e. in `deploy_on`

CyberHoward: Would this be better as a trait?

Buckram123: > Would this be better as a trait?

Not sure what to implement it on. For Abstract builder we can add `build_test()` for MockBech32 Env without sender param

CyberHoward: On `MockBase<A, S>`? 

But probably easier to have a `deploy_on_mock` fn? Would be nice to stick to the `deploy_on` naming to stay consistent.

framework/packages/abstract-client/src/builder.rs
CyberHoward: We should be able to add a `build_test` fn that excludes sender and applies for `Chain = Mock<A, S>`?

framework/packages/abstract-client/src/lib.rs
```diff
@@ -27,6 +27,10 @@ pub use publisher::{Publisher, PublisherBuilder};
 pub use service::Service;
 pub use source::AccountSource;
 
+// Re-export abstract testing for test-utils
+#[cfg(feature = "test-utils")]
+pub use abstract_testing;
```
CyberHoward, line 32: Why would we want to do this? The abstract-client is supposed to be used exclusively for integration testing while abstract-testing is a (mostly internal) unit-testing library.

framework/packages/abstract-std/src/native_addrs.rs
CyberHoward: Can we add an insta snapshot test that saves these addresses? We'll want to have tests fail if this ever changes.

framework/packages/abstract-std/src/objects/gov_type.rs
```diff
@@ -77,6 +77,7 @@ impl GovernanceDetails<String> {
     pub fn verify(
         self,
         deps: Deps,
+        // TODO: remove!
```
CyberHoward, line 80: I don't think we can remove this.

Buckram123: > I don't think we can remove this.

We have deps, we don't need to pass vc

framework/contracts/account/src/modules.rs
```diff
@@ -65,18 +62,15 @@ pub fn install_modules(
...
     let mut installed_modules = Vec::with_capacity(modules.len());
     let mut manager_modules = Vec::with_capacity(modules.len());
     let account_id = ACCOUNT_ID.load(deps.storage)?;
-    let version_control = VersionControlContract::new(version_control_address);
+    let version_control = VersionControlContract::new(deps.api)?;
 
-    let canonical_module_factory = deps
-        .api
-        .addr_canonicalize(module_factory_address.as_str())?;
+    let canonical_module_factory = CanonicalAddr::from(native_addrs::MODULE_FACTORY_ADDR);
```
CyberHoward, line 72: Can we use the object here?

Buckram123: We can, but that will add extra `addr_canonicalize`, because we need canonical address of module factory for whitelisting modules

framework/packages/abstract-std/src/objects/gov_type.rs
CyberHoward: What if we don't want to use the deterministic address?

Buckram123: > What if we don't want to use the deterministic address?

Where we wouldn't want to use deterministic address?

framework/packages/abstract-interface/src/deployment.rs
```diff
@@ -321,6 +312,20 @@ impl<Chain: CwEnv> Abstract<Chain> {
     }
 }
 
+impl Abstract<MockBech32> {
```
CyberHoward, line 315: Can we do this for arbitrary `MockBase<A, S>` or not?

framework/packages/abstract-client/src/lib.rs
Buckram123: Removed

framework/contracts/account/tests/snapshots/subaccount__creating_on_subaccount_should_succeed.snap
```diff
@@ -63,49 +57,43 @@ expression: all_storage
...
+    - "{\"native\":\"mock19jf2y29m2qvgfezcyy5zhq8phfdcysngew508eeumdsg2dyzqrqshlx44z\"}"
   - - "\u0000\u0002cb\u0000\babstract\u0000\nibc-client0.23.0"
-    - "{\"native\":\"mock1ufs3tlq4umljk0qfe8k5ya0x6hpavn897u2cnf9k0en9jr7qarqqgfzm3q\"}"
+    - "{\"native\":\"mock1u377ag8ewzrskj6r0pqutnpyvh7hd2yqq6uh3rh2pclex2gzhslsq3frn0\"}"
   - - "\u0000\u0002cb\u0000\babstract\u0000\u000emodule-factory0.23.0"
-    - "{\"native\":\"mock1qg5ega6dykkxc307y25pecuufrjkxkaggkkxh7nad0vhyhtuhw3sep3g9s\"}"
+    - "{\"native\":\"mock10ymkph7y47lmxkyan49cq2yjm5m3n2nk83efppcwwpunx7gs374qg8fkqs\"}"
   - - "\u0000\u0002cb\u0000\babstract\u0000\u000fversion-control0.23.0"
-    - "{\"native\":\"mock1wug8sewp6cedgkmrmvhl3lf3tulagm9hnvy8p0rppz9yjw0g4wtqwm38hv\"}"
+    - "{\"native\":\"mock18aq2nqrm809gdktwetuk6c48c6yj379ptkdeewv6j797xv3j2gssdzgurv\"}"
```
Buckram123, line 90: @CyberHoward We already save those addresses inside all snapshot tests. Address change is also covered by `deploy2` test

framework/packages/abstract-std/src/native_addrs.rs
Buckram123: See https://github.com/AbstractSDK/abstract/pull/448#discussion_r1772805653

framework/packages/abstract-client/src/builder.rs
```diff
@@ -155,6 +155,13 @@ impl<Chain: CwEnv> AbstractClientBuilder<Chain> {
     }
 }
 
+impl<A: cosmwasm_std::Api, S: StateInterface> AbstractClientBuilder<MockBase<A, S>> {
+    /// Deploy mock abstract with current configuration
+    pub fn build_test(&self) -> AbstractClientResult<AbstractClient<MockBase<A, S>>> {
```
CyberHoward, line 160: I prefer `build_mock`, wdyt?
```suggestion
    pub fn build_mock(&self) -> AbstractClientResult<AbstractClient<MockBase<A, S>>> {
```

Buckram123: I assume we will have such helper for clone testing env as well. Would you prefer to call it `build_clone()`?

framework/contracts/account/tests/subaccount.rs
```diff
@@ -137,7 +136,7 @@ fn recursive_updating_on_subaccount_should_succeed() -> AResult {
 fn installed_app_updating_on_subaccount_should_succeed() -> AResult {
     let chain = MockBech32::new("mock");
     let sender = chain.sender_addr();
-    let deployment = Abstract::deploy_on(chain.clone(), mock_bech32_admin(&chain))?;
+    let deployment = Abstract::deploy_on_test(chain.clone())?;
```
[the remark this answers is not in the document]
Buckram123: See my previous comment

framework/packages/abstract-client/src/builder.rs
CyberHoward: Which one? The helper for clone-testing? 

Yeah, but maybe there it would be `build_clone_env` otherwise the "clone" is a bit ambiguous.

CyberHoward (approved):

PR #455, opened 2024-09-13
Fix `Account` tests

CyberHoward:
This PR aims to fix the tests for the `Account` that were migrated from the proxy and manager contracts.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/contracts/account/src/config.rs
```diff
@@ -96,117 +117,34 @@ pub fn update_info(
...
 
     mod set_owner_and_gov_type {
-        use ownership::GovAction;
+        use ownership::{GovAction, GovOwnershipError};
 
         use super::*;
 
         #[test]
-        fn only_owner() -> ManagerTestResult {
+        fn only_owner() -> anyhow::Result<()> {
```
Buckram123, line 147: Using anyhow result in tests is much simpler, great idea!

framework/contracts/account/src/contract.rs
```diff
@@ -182,19 +197,41 @@ pub fn instantiate(
...
+    for fee in total_fee.clone() {
+        let mut total_received = Coins::try_from(info.funds.clone()).unwrap();
+
+        total_received.sub(fee).map_err(|_| {
+            abstract_std::AbstractError::Fee(format!(
+                "Invalid fee payment sent. Expected {}, sent {:?}",
+                total_fee, info.funds
+            ))
+        })?;
+    }
```
Buckram123, line 233: It looks incorrect(`total_received` created inside loop). This can allow to sub same fund multiple times. For example Carrot App and Dex Adapter both cost 5 juno to install, sending 5 juno will be enough to pass this loop

framework/contracts/account/src/execution.rs
```diff
@@ -108,16 +108,27 @@ pub fn ica_action(deps: DepsMut, msg_info: MessageInfo, action_query: Binary) ->
...
+
+        use cosmwasm_std::{testing::MOCK_CONTRACT_ADDR, wasm_execute, CosmosMsg};
 
         use super::*;
 
         #[test]
         fn only_whitelisted_can_execute() {
             let mut deps = mock_dependencies();
-            mock_init(&mut deps);
+            mock_init(&mut deps).unwrap();
```
Buckram123, line 131: NIT if you want all tests to return `anyhow::Result`

framework/contracts/account/src/execution.rs
```diff
@@ -178,23 +186,22 @@ mod test {
...
-        use cosmwasm_std::coins;
+        use crate::modules::update_module_addresses;
+
+        use super::*;
 
         #[test]
         fn add_module() {
             let mut deps = mock_dependencies();
-            mock_init(&mut deps);
+            mock_init(&mut deps).unwrap();
```
Buckram123, line 198: NIT: anyhow result

framework/contracts/account/src/modules.rs
```diff
@@ -347,30 +339,65 @@ fn _whitelist_modules(deps: DepsMut, modules: Vec<String>) -> AccountResult<()>
...
+            if !whitelisted_modules.0.contains(&module_addr) {
+                return Err(AccountError::NotWhitelisted(module_addr.to_string()));
+            }
+            // Remove contract from whitelist.
+            whitelisted_modules.0.retain(|addr| *addr != module_addr);
+            Ok(whitelisted_modules)
+        })?;
+    }
     Ok(())
 }
```
Buckram123, line 357: Let's have single load or move loop inside `update`. This will `load`+`save` `WHITELISTED_MODULES` for every module, instead of 1 load and 1 save

framework/contracts/account/src/modules.rs
```diff
@@ -657,196 +657,148 @@ mod tests {
...
-                .is_equal_to(ProxyError::NotWhitelisted(test_module_addr.to_string()));
-        }
-    }
+    // TODO: move these tests to integration tests
+
+    // mod add_module {
+    //     use super::*;
+
+    //     use cw_controllers::AdminError;
+
```
Buckram123, line 675: This TODO will be done in different PR?

framework/contracts/account/src/queries.rs
```diff
@@ -194,64 +194,111 @@ mod test {
...
         let abstr = AbstractMockAddrs::new(deps.api);
+        deps.querier = mock_querier_builder(deps.api.clone())
+            .with_contract_version(&abstr.module_address, TEST_MODULE_ID, "1.0.0")
+            .build();
+        mock_init(&mut deps).unwrap();
+
+        execute_as_admin(
+            &mut deps,
+            ExecuteMsg::UpdateInternalConfig(InternalConfigAction::UpdateModuleAddresses {
+                to_add: vec![(TEST_MODULE_ID.into(), abstr.module_address.to_string())],
```
Buckram123, line 223: NIT: many unwraps in this test, want to return anyhow result instead?

framework/contracts/account/src/sub_account.rs
```diff
@@ -184,3 +185,56 @@ pub fn maybe_update_sub_account_governance(deps: DepsMut) -> AccountResult<Vec<C
...
+        // Remove the namespace that this account holds.
+        msgs.push(
+            wasm_execute(
+                vc.address,
+                &abstract_std::version_control::ExecuteMsg::RemoveNamespaces {
+                    namespaces: vec![namespace.to_string()],
+                },
+                vec![],
+            )?
+            .into(),
```
Buckram123, line 236: This namespace can be used by other users after that right? Remove confusing name for it. And do we still want it to use array of namespaces?

Buckram123 (changes requested):
Small fixes, and couple of NITS(optional)

framework/contracts/account/src/contract.rs
CyberHoward: Great catch, thanks Misha!

framework/contracts/account/src/modules.rs
CyberHoward: Yes, let's 

I think this was already commented out before the refactor

framework/contracts/account/src/sub_account.rs
CyberHoward: It's nice for us to be able to batch removals if we want. 

Which name would you suggest? "ForgoNamespace"?

framework/contracts/account/src/modules.rs
Buckram123: Not pushed? 👀

framework/contracts/account/src/sub_account.rs
Buckram123: Yeah or return/disclaim

framework/contracts/account/src/modules.rs
CyberHoward: pushed :)

Buckram123 (approved):
LGTM

framework/contracts/account/src/modules.rs
Buckram123: > Better now? x)

Yeah, nice!

framework/contracts/account/src/sub_account.rs
CyberHoward: @Kayanski what do you think?

CyberHoward: Let's do it in a different PR: https://linear.app/abstract-sdk/issue/ABS-538/rename-removenamespace-on-vc

PR #458, opened 2024-09-16
Update Adapter Access Control Logic

Kayanski:
This PR aims at updating the Access Control for actions via the account.

## Proposed changes: 

- Merge All actions on the account (ExecOnModule, ModuleActions) to a single LocalAction variant
  - ExecOnModule simply fetches the module address and dispatches the message
- Add a `ConfigureModule` message on the account for sending admin calls to modules
- Enforce this access control inside Adapter Base Variant
- Modify the NestedAmin structure to be able to account for this admin change

This is Options 2 here, there is a mermaid diagram at the end to understand the situation a little better https://www.notion.so/abstract-money/Merging-Abstract-Account-Implementation-6d366bb35b3242adb6ae5787fb759cc8?pvs=4

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/contracts/account/src/modules.rs
```diff
@@ -234,29 +234,6 @@ pub fn uninstall_module(mut deps: DepsMut, info: MessageInfo, module_id: String)
     Ok(response)
 }
 
-/// Execute the [`exec_msg`] on the provided [`module_id`],
-pub fn exec_on_module(
```
Kayanski, line 238: moved with the other actions

framework/packages/abstract-client/src/account.rs
```diff
@@ -632,14 +632,7 @@ impl<Chain: CwEnv> Account<Chain> {
...
         let msgs = execute_msgs.into_iter().map(Into::into).collect();
-        self.configure(
-            &account::ExecuteMsg::ExecOnModule {
-                module_id: ACCOUNT.to_owned(),
-                exec_msg: to_json_binary(&abstract_std::account::ExecuteMsg::ModuleAction { msgs })
-                    .map_err(AbstractInterfaceError::from)?,
-            },
-            funds,
-        )
+        self.configure(&account::ExecuteMsg::AccountActions { msgs }, funds)
```
Kayanski, line 635: Proxy removal, no need for nested execution here

framework/contracts/account/src/contract.rs
```diff
@@ -283,12 +280,18 @@ pub fn execute(mut deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg)
...
-                    execute_module_action_response(deps, info, msg).map_err(AccountError::from)
+                ExecuteMsg::AccountActionWithData { msg } => {
+                    execute_account_action_response(deps, &info.sender, msg)
+                        .map_err(AccountError::from)
                 }
+
+                ExecuteMsg::ExecOnModule {
+                    module_id,
+                    exec_msg,
+                } => exec_on_module(deps, info, module_id, exec_msg).map_err(AccountError::from),
```
Buckram123, line 294: It looks like one module can call base message on another module. It's fine, as long as user trust base method of the adapter, right?

In case it's not clear how:
- Admin calls `admin_on_module` on whitelisted module with custom base handler
- Now this module can call base via `exec_on_module` on any module or modules (`ADMIN_CALL_TO_CONTEXT` will not be cleared until this module execution is done)

P.S. Sorry if I ruined your backdoor :rofl:

Buckram123 (approved):
Looks good! Noticed one edge case, which I hope is unlikely to happen

framework/contracts/account/src/actions.rs
```diff
@@ -3,48 +3,114 @@ use abstract_sdk::std::{
...
+    }
+
+    // If whitelisted address, ok
+    let whitelisted_modules = WHITELISTED_MODULES.load(deps.storage)?;
+    if whitelisted_modules.0.contains(sender) {
+        return Ok(());
+    }
+
+    Err(AccountError::SenderNotAuthorized {})
+}
```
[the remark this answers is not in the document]
Kayanski: Good question, but why do we care ?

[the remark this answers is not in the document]
Buckram123: This function will be executed pretty often, it does make sense to do small gas optimization if possible

framework/contracts/account/src/contract.rs
Buckram123: Oh, nice! Missed that there's mermaid graphs as well, thank you.

framework/contracts/account/src/actions.rs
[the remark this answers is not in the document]
Kayanski: I think that owners will first call a module action and then the module might call other modules, so this order seems ok, wdyt ?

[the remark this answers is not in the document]
Kayanski: But I see that there was already a function that did that in the current develop/v2 branch, so didn't touch what was already there

CyberHoward:
> Namely, today only the admin (might be the account) OR the top level owner can execute on the admin variants. We need to modify the NestedAdmin structure to be able to verify that IF the admin is an account, it should be in an admin state. Is that an acceptable behavior

Yes, I think this is the behavior we want. A nucance here is that we want to do the admin-flag assertion on the direct admin of the app (i.e. the account that the app is installed on) vs the top-level account. Because an admin action on a sub-account should not require an admin flag be set on the owner account.

Kayanski:
> > Namely, today only the admin (might be the account) OR the top level owner can execute on the admin variants. We need to modify the NestedAdmin structure to be able to verify that IF the admin is an account, it should be in an admin state. Is that an acceptable behavior
> 
> Yes, I think this is the behavior we want. A nucance here is that we want to do the admin-flag assertion on the direct admin of the app (i.e. the account that the app is installed on) vs the top-level account. Because an admin action on a sub-account should not require an admin flag be set on the owner account.

I'm not sure we can keep the current Admin design (same API as cw_controllers::Admin) with this new constraint of checking the admin variable on the admin only if it's an account.
That's is mainly because now `Admin` is not just an address but an address with some conditions. We should sync on this one if possible

framework/packages/abstract-std/src/account.rs
```diff
@@ -124,9 +127,17 @@ pub struct CallbackMsg {}
 #[derive(cw_orch::ExecuteFns)]
 pub enum ExecuteMsg {
     /// Executes the provided messages if sender is whitelisted
-    ModuleAction { msgs: Vec<CosmosMsg<Empty>> },
+    AccountActions { msgs: Vec<CosmosMsg<Empty>> },
```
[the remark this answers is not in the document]
Buckram123: > Can we just call this `Execute`?

Agree, but then we will have to rename cw-orch method for it, what would you prefer to call it?

CyberHoward (commented):
Could you update `packages/abstract-std/src/objects/ownership/README.md` with an explanation of how the `CALLING_TO_AS_ADMIN` value is used?

framework/contracts/account/src/execution.rs
```diff
@@ -26,30 +31,83 @@ pub(crate) fn assert_whitelisted_or_owner(deps: Deps, sender: &Addr) -> AccountR
...
 
+/// Execute the [`exec_msg`] on the provided [`module_id`],
+/// This is a simple wrapper around `LocalAction`
+pub fn execute_on_module(
+    deps: DepsMut,
+    info: MessageInfo,
+    module_id: String,
+    exec_msg: Binary,
+) -> AccountResult {
+    let module_addr = load_module_addr(deps.storage, &module_id)?;
```
CyberHoward, line 66: Forgot authentication here!

framework/packages/abstract-std/src/objects/ownership/README.md
```diff
@@ -92,11 +96,87 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+### Usage inside a module
+
+To use this functionality, Abstract provides helpers in form of the `NestedAdmin` structure. This structure should be used to handle `Abstract Accounts` as admin of a contract.
 
-### NFT governance type
+The `NestedAdmin::assert_admin` function will only return an `Result::Ok` if any of those conditions is true:
 
-In case NFT contract does not return owner of `owner_of`, ownership will act as renounced. For example NFT got burned or something happened with NFT contract.
+- The caller is the saved Account AND the `CALLING_TO_AS_ADMIN` variable is set on the account to either:
+  - The contract account address (`env.contract.address` is supposed to be fed to the self_addr variable)
```
Kayanski, line 126: Added a Readme. Because of this line, I think we should take in `Env` instead of `self_addr` otherwise, it's simple to exchange the 2 addresses required.

Kayanski: @CyberHoward wdyt ?

CyberHoward: Yeah agree. Realized it was a bad change when I was reviewing as well.

CyberHoward: `&Env` preferably

framework/contracts/account/src/contract.rs
```diff
@@ -337,18 +335,33 @@ pub fn execute(mut deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg)
...
+                    install_modules,
+                    account_id,
+                )
+                .map_err(AccountError::from),
+                ExecuteMsg::UpdateSubAccount(action) => {
+                    handle_sub_account_action(deps, info, action).map_err(AccountError::from)
                 }
+
+                // ## Other ##
+                // TODO: Update module migrate logic to not use callback!
```
Kayanski, line 364: @CyberHoward I don't remember, did you add this TODO here ?

CyberHoward: I did. In theory we should be able to use the reply payload field.

Kayanski: Ok, that's for another PR !

Buckram123: *Not without `cosmwasm_2_0` feature

.github/workflows/framework-test.yml
```diff
@@ -40,6 +40,17 @@ jobs:
         if: hashFiles('Cargo.lock') == ''
         working-directory: ./framework
         run: cargo generate-lockfile
+      - name: Free up disk space on runner
+        run: |
+          df -h
+          rm -rf /usr/share/dotnet/
+          sudo apt-get remove -y 'php.*'
+          sudo apt-get remove -y '^dotnet-.*'
+          sudo apt-get remove -y azure-cli firefox powershell mono-devel
```
Buckram123, line 49: :exploding_head: why it has all this stuff installed haha

PR #459, opened 2024-09-17
Fix ans-host tests

Buckram123:
This PR aims to fix ANS host tests

### DIFF
--- a/framework/contracts/native/ans-host/src/commands.rs
+++ b/framework/contracts/native/ans-host/src/commands.rs
@@ -365,7 +365,10 @@ fn validate_pool_assets(
 mod test {
     #![allow(clippy::needless_borrows_for_generic_args)]
     use abstract_testing::{map_tester::CwMapTester, prelude::*};
-    use cosmwasm_std::{testing::*, Addr, DepsMut};
+    use cosmwasm_std::{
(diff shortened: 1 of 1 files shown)


framework/contracts/native/ans-host/src/commands.rs
```diff
@@ -365,7 +365,10 @@ fn validate_pool_assets(
 mod test {
     #![allow(clippy::needless_borrows_for_generic_args)]
     use abstract_testing::{map_tester::CwMapTester, prelude::*};
-    use cosmwasm_std::{testing::*, Addr, DepsMut};
+    use cosmwasm_std::{
+        testing::{message_info, mock_env, MockApi},
+        Addr, DepsMut,
+    };
```
Buckram123, line 371: We added `mock_dependencies` to `abstract_testing`, but now it's conflicting with `cosmwasm_std::testing::mock_dependencies`. Should we rename it to something else?

Kayanski: I think we should NOT import cosmwasm-std::testing::*, it's a better practice if we have this kind of conflicts

Buckram123:
Closed in favor of #460(no changes needed without conflicting name from abstract-testing)

PR #460, opened 2024-09-17
Remove `mock_dependencies`

Buckram123:
This PR removes `mock_dependencies` from abstract-testing, because of the name conflict with `cosmwasm_std::testing::mock_dependencies`

### DIFF
--- a/framework/packages/abstract-testing/src/lib.rs
+++ b/framework/packages/abstract-testing/src/lib.rs
@@ -120,7 +120,7 @@ pub fn mock_querier(mock_api: MockApi) -> MockQuerier {
 /// Abstract-specific mock dependencies.
 ///
 /// Sets the required queries for native contracts and the root Abstract Account.
-pub fn mock_dependencies() -> MockDeps {
+pub fn mock_deps() -> MockDeps {
     let api = MockApi::default();
     let querier = mock_querier(api.clone());
 
@@ -207,7 +207,7 @@ pub mod module {
 }
 
 pub mod prelude {
-    pub use super::{mock_dependencies, mock_querier};
+    pub use super::{mock_deps, mock_querier};
     pub use abstract_mock_querier::AbstractMockQuerier;
     use abstract_std::objects::{AccountId, AccountTrace};
     pub use addresses::*;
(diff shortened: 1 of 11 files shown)


Kayanski:
Are we renaming it OR do we make sure that we don't import the cosmwasm-std testing mock_dependencies and keep the names ?

Buckram123:
> Are we renaming it OR do we make sure that we don't import the cosmwasm-std testing mock_dependencies and keep the names ?

Where we had `mock_dependencies` in tests, it was supposed to be `cosmwasm_std::testing::mock_dependencies`, instead of more complex abstract mock environment(Don't want to slow down tests for no reason). So I would prefer to rename abstract-testing

Kayanski:
If it's a question of only some places having the abstract one, I think that this mock_deps new name is not explicit enough. We should either use: 
`abstract-testing::mock_dependencies` explicitely or have it renamed `abstract_mock_dependencies`, so that it's clear that it's abstract we're talking about. I think it's worth the precision so that it's easier to deal with later

Buckram123:
> `abstract_mock_dependencies`

I like it as `abstract_mock_dependencies`

Buckram123:
Thank you! Added `abstract_` prefix to other mock functions for `abstract-std` as well, what do you think? @Kayanski

framework/packages/abstract-testing/src/lib.rs
```diff
@@ -113,23 +113,8 @@ pub fn mock_querier_builder(mock_api: MockApi) -> MockQuerierBuilder {
...
-pub fn mock_dependencies() -> MockDeps {
-    let api = MockApi::default();
-    let querier = mock_querier(api.clone());
-
-    OwnedDeps {
-        storage: MockStorage::default(),
-        api,
-        querier,
-        custom_query_type: std::marker::PhantomData,
-    }
```
Buckram123, line 132: It's not very useful when `abstract_mock_querier()` available

Buckram123:
Closes ABS-525

CyberHoward (approved):

Kayanski (approved):

PR #461, opened 2024-09-17
V2/Fix abstract sdk tests

Buckram123:


framework/packages/abstract-sdk/src/apis/authz.rs
```diff
@@ -460,13 +469,26 @@ impl AuthZ {
 mod tests {
     use super::*;
 
+    use crate::std::version_control::Account;
     use crate::{apis::stargate::convert_stamp, mock_module::*};
+    use abstract_testing::abstract_mock_querier_builder;
+    use abstract_testing::prelude::*;
     use cosmwasm_std::testing::mock_dependencies;
 
+    pub fn setup() -> (MockDeps, Account, MockModule) {
```
CyberHoward, line 478: This setup fn seems to be the same everywhere and the package already has a `test-utils` feature so IMO it makes sense to pull this fn out of every file into a single location.

framework/packages/abstract-sdk/src/apis/splitter.rs
```diff
@@ -83,8 +84,12 @@ mod test {
     };
 
     fn split() -> Result<Response, AbstractSdkError> {
-        let deps = mock_dependencies();
-        let module = MockModule::new(deps.api);
+        let mut deps = mock_dependencies();
+        let account = test_account_base(deps.api);
```
CyberHoward, line 88: use common setup fn here

CyberHoward (approved):
Small nit on setup fn :)

framework/packages/abstract-sdk/src/apis/splitter.rs
Buckram123: Yep

PR #463, opened 2024-09-17
V2/Fix version control query unit tests

Buckram123:


Kayanski (approved):

framework/contracts/native/version-control/src/queries.rs
```diff
@@ -260,822 +260,856 @@ fn filter_modules_by_namespace(
...
+                security_disabled: Some(true),
+                namespace_registration_fee: None,
+            },
+        )?;
+
+        Ok(())
+    }
+
+    /// Initialize the version_control with admin as creator and test account
+    fn mock_init_with_account(deps: &mut MockDeps) -> VCResult {
```
CyberHoward, line 373: aren't these fns also present in other files? (commands.rs / contract.rs)
is it possible for them to use the same setup fns?

CyberHoward (approved):
Nit on setup fn

framework/contracts/native/version-control/src/queries.rs
Buckram123: Couldn't make `mock_init_with_account()` compatible with both tests. Don't see a reason to complicate tests structure by reusing `execute_as()`

PR #465, opened 2024-09-18
V2/Fix ibc host tests

Buckram123:


framework/contracts/native/ibc-host/tests/integration.rs
```diff
@@ -175,7 +181,7 @@ fn account_creation_full() -> anyhow::Result<()> {
         ConfigResponse {
             ans_host_address: abstr_origin.ans_host.address()?,
             version_control_address: abstr_origin.version_control.address()?,
-            account_factory_address: abstr_origin.account_factory.address()?,
+            module_factory_address: abstr_origin.module_factory.address()?,
```
Kayanski, line 184: Why is module_factory now included in the config response ?

Kayanski (approved):

framework/contracts/native/ibc-host/tests/integration.rs
Buckram123: Account need it during instantiation, it was forwarded by account factory before

framework/contracts/native/ibc-host/tests/integration.rs
```diff
@@ -52,7 +52,7 @@ fn account_creation() -> anyhow::Result<()> {
...
```
CyberHoward, line 55: Does the ibc-host ever use the module factory?

CyberHoward (approved):

PR #467, opened 2024-09-18
V2/Fix abstract app tests

Buckram123:


framework/packages/abstract-app/Cargo.toml
```diff
@@ -39,6 +39,5 @@ workspace-hack = { version = "0.1", path = "../../workspace-hack" }
 
 [dev-dependencies]
 cosmwasm-schema = { workspace = true }
-speculoos = { workspace = true }
```
Buckram123, line 42: Also did small sidequest, hehe

Kayanski: Don't we like speculoos ? @CyberHoward

Kayanski (approved):

framework/packages/abstract-app/Cargo.toml
Buckram123: We discussed it already when you were gone, but there is no benefit for using it. It's also not maintained and no one uses it.

framework/packages/abstract-app/src/endpoints/query.rs
```diff
@@ -55,11 +55,10 @@ impl<
 
     fn dapp_config(&self, deps: Deps) -> StdResult<AppConfigResponse> {
         let state = self.base_state.load(deps.storage)?;
-        let admin = self.admin.get(deps)?.unwrap();
         Ok(AppConfigResponse {
-            account: state.account.into_addr(),
+            account: state.account.addr().to_owned(),
```
CyberHoward, line 59:
```suggestion
            account: state.account.into_addr(),
```

CyberHoward (approved):
LGTM!

framework/packages/abstract-app/src/endpoints/query.rs
Buckram123: Thank you

PR #462, opened 2024-09-17
V2/Fix module factory tests

Buckram123:
Depends on #455

PR #470, opened 2024-09-20
module-factory tests

Buckram123:
Cherrypicked tests fix commit from #462


Buckram123:
Will just merge it, there is nothing to check here

PR #474, opened 2024-09-23
 Better Interchain API for tests / execution

Kayanski:
This PR aims at making the Abstract IBC dev experience even better in tests / actual executions.
This is Abstract cw1 because cw2 tests don't pass and I need tests
Proposed changes: 
- `AbstractInterchainClient` that is responsible for deploying Abstract to all chains of an `InterchainEnv` and linking them via an IBC connection

Depends on https://github.com/AbstractSDK/polytone/pull/9

### Checklist

- [ ] CI is green.
- [x] Changelog updated.


framework/packages/abstract-client/src/lib.rs
```diff
@@ -32,14 +32,83 @@ pub use source::AccountSource;
...
+    pub struct AbstractInterchainClient<Chain: IbcQueryHandler> {
+        /// All clients registered within this Interchain Client
+        pub clients: HashMap<String, AbstractClient<Chain>>,
+    }
+
+    impl<Chain: IbcQueryHandler> AbstractInterchainClient<Chain> {
+        /// Deploys and connects Abstract instances across all chains specified
+        pub fn build<Interchain: InterchainEnv<Chain>>(
+            interchain: &Interchain,
+        ) -> AbstractClientResult<Self> {
```
Buckram123, line 59: Why is it called `build` if we never get to the builder?

framework/packages/abstract-client/src/lib.rs
```diff
@@ -32,14 +32,83 @@ pub use source::AccountSource;
 mod interchain {
```
CyberHoward, line 32: Can we move this to a file?

framework/packages/abstract-client/src/lib.rs
Buckram123: What if we duplicate AbstractClientBuilder methods but with "chain_id" passed, so you can edit each abstract individually?

framework/packages/abstract-client/src/lib.rs
```diff
@@ -32,14 +32,83 @@ pub use source::AccountSource;
...
+
+    /// Client to interact with Abstract and setup interchain capabilities
+    pub struct AbstractInterchainClient<Chain: IbcQueryHandler> {
+        /// All clients registered within this Interchain Client
+        pub clients: HashMap<String, AbstractClient<Chain>>,
+    }
+
+    impl<Chain: IbcQueryHandler> AbstractInterchainClient<Chain> {
+        /// Deploys and connects Abstract instances across all chains specified
+        pub fn build<Interchain: InterchainEnv<Chain>>(
```
CyberHoward, line 57: let's move `fn new` above `fn build`

CyberHoward (changes requested):
small nits

framework/packages/abstract-client/src/lib.rs
```diff
@@ -32,14 +32,83 @@ pub use source::AccountSource;
...
+            Ok(AbstractInterchainClient {
+                clients: clients
+                    .into_iter()
+                    .map(|c| (c.environment().chain_id(), c))
+                    .collect(),
+            })
+        }
+
+        /// Getter for an abstract client within this object
+        pub fn get_abstract(&self, chain_id: &str) -> AbstractClientResult<AbstractClient<Chain>> {
```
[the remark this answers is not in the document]
Buckram123: or just `client`

Buckram123 (changes requested):
What is the meaning of `build` here?

framework/packages/abstract-client/src/lib.rs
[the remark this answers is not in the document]
Kayanski: Danke schön,

framework/packages/abstract-client/src/interchain/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use cw_orch_interchain::{IbcQueryHandler, InterchainEnv, InterchainError};
+pub use remote_account::RemoteAccount;
+pub use remote_application::RemoteApplication;
+
+use crate::{client::AbstractClientResult, AbstractClient, Environment};
+
+// TODO: Why are we not returning ibc tx analysis after await
+/// IbcTxAnalysis after waiting for interchain action
+pub struct IbcTxAnalysisV2<Chain: cw_orch::environment::CwEnv>(
```
CyberHoward, line 13: why is this a new type?

Buckram123: `IbcTxAnalysis` raises a warning for being `must_use`, but it's checked analysis

CyberHoward: @Kayanski

Kayanski: I discovered that 2 days ago yes, I don't really like this because this nesting is just there to mitigate the must_use. 
I don't have a solution right now if people want this result, they should use it. If they don't want it, they should not care, and we should not return it. Don't knoow what the devX looks like. But V2 seems weird that's for sure

CyberHoward: Yeah there's something that we should improve on the design of that type.

CyberHoward:
Waiting with this until we fix the unused tag

modules/contracts/apps/ping-pong/tests/better_interchain_api.rs
CyberHoward, line 1: What's this file used for?

CyberHoward (approved):
LGTM, just wandering what `better_interchain_api` is used for. Maybe it needs a better name or maybe we can update `ping_pong_tests` to use the better API instead?

modules/contracts/apps/ping-pong/tests/better_interchain_api.rs
Kayanski: This is here to test that the client builder has the right behavior and loads the right addresses.

Kayanski: Moved to abstract-client package

Buckram123 (approved):

PR #475, opened 2024-09-24
Fix modules and interchain tests

Buckram123:
Tests compile, but many of the tests ignored right now as we don't have usable wyndex-bundle
Clone tests only usable on astrovault dex as it's the only dex migrated to cosmwasm 2


CyberHoward (approved):

Buckram123:
@CyberHoward InterchainEnv now have clone, so refactored remote accounts and applications for clients to fix and improve tests for ping-pong. So instead of holding lifetimed reference to IbcEnv now it's cloned IbcEnv

framework/packages/abstract-interface/src/deployment.rs
```diff
@@ -314,19 +314,21 @@ impl<Chain: CwEnv> Abstract<Chain> {
...
+    pub fn mock_admin(chain: &Chain) -> <MockBase as TxHandler>::Sender {
+        // Getting prefix
+        let sender_addr: cosmrs::AccountId = chain.sender().as_str().parse().unwrap();
+        let prefix = sender_addr.prefix();
+        // Building mock_admin
+        let mock_admin =
+            cosmrs::AccountId::new(prefix, &native_addrs::TEST_ABSTRACT_CREATOR).unwrap();
+        Addr::unchecked(mock_admin)
     }
 }
```
Buckram123, line 333: Made those helpers even broader as well

CyberHoward (approved):

PR #476, opened 2024-09-26
Add abstract account to govdetails

Buckram123:
Adding AbstractAccount Governance and admin check based on first bit of signature id. This bit will set AUTH_ADMIN in state and used to verify admin call

#473 is base


framework/contracts/account/src/execution.rs
```diff
@@ -13,7 +13,17 @@ use crate::{
 };
 
 /// Check that sender either whitelisted or governance
-pub(crate) fn assert_whitelisted_or_owner(deps: Deps, sender: &Addr) -> AccountResult<()> {
+pub(crate) fn assert_whitelisted_or_owner(deps: &mut DepsMut, sender: &Addr) -> AccountResult<()> {
+    #[cfg(feature = "xion")]
+    {
+        if let Some(is_admin) = crate::state::AUTH_ADMIN.may_load(deps.storage)? {
+            // Clear auth if it was set
```
CyberHoward, line 20: maybe this should be cleared in the post_tx hook?

CyberHoward (approved):

framework/contracts/account/src/execution.rs
Buckram123: It's tough one! Clearing it here disallows more than 1 admin action per tx, but clearing it in post-tx allows module to execute as admin

Buckram123: Actually clearing it in post tx will pretty much have same result as having check of if account caller is account address and no AUTH_ADMIN at all

PR #481, opened 2024-10-02
Fix predictable addrs

Buckram123:
This PR aims to fix predictable address of the abstract native contracts.
Downside is we need to get hrp from somewhere so `&Env` is used everywhere for now


CyberHoward:
Another option would be to feature-flag the prefix for each chain and compile with the corresponding feature ?

Buckram123:
> Another option would be to feature-flag the prefix for each chain and compile with the corresponding feature ?

Sure, we can do the same for canonical addresses for each chain, not sure it's a good idea to have native wasm for each chain, just to reduce state tho

Kayanski:
Before I review further, just to make sure I agree the rationale behind this change. 
Most of the changes here come from the fact that Canonical Addresses are not common between chains? So that a same account won't have the same canonical address on multiple chains?

Buckram123:
> Before I review further, just to make sure I agree the rationale behind this change. Most of the changes here come from the fact that Canonical Addresses are not common between chains? So that a same account won't have the same canonical address on multiple chains?

Yes, so we only store public address, who instantiated blob

interchain/scripts/src/bin/full_deploy.rs
```diff
@@ -44,11 +44,6 @@ fn full_deploy(mut networks: Vec<ChainInfoOwned>) -> anyhow::Result<()> {
     // write_deployment(&deployment_status)?;
 
     for network in networks {
-        let urls = network.grpc_urls.to_vec();
```
Kayanski, line 47: Yes, nice

interchain/scripts/src/bin/full_deploy.rs
```diff
@@ -23,15 +23,15 @@ fn full_deploy(mut networks: Vec<ChainInfoOwned>) -> anyhow::Result<()> {
...
-    let deployment_status = read_deployment()?;
-    if deployment_status.success {
-        log::info!("Do you want to re-deploy to {:?}?", networks);
-        let mut input = String::new();
-        std::io::stdin().read_line(&mut input)?;
-        if input.to_lowercase().contains('n') {
-            return Ok(());
-        }
-    }
+    // let deployment_status = read_deployment()?;
```
Kayanski, line 26: what's happening here ? Is this part not relevant anymore ?

Kayanski (approved):
Ok ! 
A side question on this. We have some traits in the sdk that necessitate access to registry and ANS host, those traits can be used more generally now that the address is deterministic and fixed (and not stored anymore), it might be worth thinking about that in the future

framework/contracts/account/src/absacc/auth/util.rs
```diff
@@ -6,9 +6,7 @@ pub const CHAIN_BECH_PREFIX: &str = "xion";
 use crate::contract::AccountResult;
 
 pub fn sha256(msg: &[u8]) -> Vec<u8> {
-    let mut hasher = Sha256::new();
-    hasher.update(msg);
-    hasher.finalize().to_vec()
+    Sha256::digest(msg).to_vec()
```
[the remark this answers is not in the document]
Buckram123: No, just noticed this small function could be simplified

framework/contracts/native/ans-host/src/commands.rs
```diff
@@ -1230,6 +1242,7 @@ mod test {
         #[test]
         fn add_five_asset_pool() -> AnsHostTestResult {
             let mut deps = mock_dependencies();
+            let env = mock_env_validated(deps.api);
```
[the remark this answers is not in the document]
Buckram123: You're right, thank you

Buckram123:
> Ok ! A side question on this. We have some traits in the sdk that necessitate access to registry and ANS host, those traits can be used more generally now that the address is deterministic and fixed (and not stored anymore), it might be worth thinking about that in the future

Yeah, we can pass `Api` instead of deps as I don't think we need storage anymore, and only Api methods

interchain/scripts/src/bin/full_deploy.rs
Buckram123: I don't have deployment status file and it's not inside this repo either. This script was erroring without this file, so I commented it out for now.

framework/packages/abstract-testing/src/lib.rs
```diff
@@ -117,11 +118,23 @@ pub fn abstract_mock_querier(mock_api: MockApi) -> MockQuerier {
     abstract_mock_querier_builder(mock_api).build()
 }
 
+/// cosmwasm_std::mock_env_validated(deps.api), but address generated with MockApi
+pub fn mock_env_validated(mock_api: MockApi) -> Env {
```
CyberHoward, line 122: We should open an issue on cosmwasm to change the env.contract.addr to a valid addr instead of the string that it currently is.

CyberHoward (approved):

framework/packages/abstract-testing/src/lib.rs
Kayanski: Yes ! Just like they removed mock_info to message_info !

CyberHoward:
> > Ok ! A side question on this. We have some traits in the sdk that necessitate access to registry and ANS host, those traits can be used more generally now that the address is deterministic and fixed (and not stored anymore), it might be worth thinking about that in the future
> 
> Yeah, we can pass `Api` instead of deps as I don't think we need storage anymore, and only Api methods. Let's fix it in the next PR

We need `deps` to perform queries on the addresses tho. 

What I think we could do to simplify things is to move to a `Context` based system where the `Context` contains deps + env + info.

PR #482, opened 2024-10-04
Moved ica-client

Kayanski:
This PR aims at placing the ica-client definitions in their right place and creating tests for the implementation (EVM and Cosmos)

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-std/src/native/ica_client.rs
```diff
@@ -0,0 +1,173 @@
...
+        fn chain_type(&self) -> Option<ChainType>;
+    }
+
+    impl CastChainType for TruncatedChainId {
+        // Return the type of chain based on the chain-id.
+        // Note: chain-ids for EVM chains are numbers!
+        fn chain_type(&self) -> Option<ChainType> {
+            // TODO: Should be a better way to have lookup table, maybe this will be nice:
+            // https://crates.io/crates/phf
+            let chains = map_macro::hash_map! {
```
Buckram123, line 90: Have you considered lookup table alternative? What do you think about phf?

framework/packages/abstract-std/src/native/ica_client.rs
```diff
@@ -0,0 +1,173 @@
...
+        // ... other actions?
+    }
+
+    #[cosmwasm_schema::cw_serde]
+    #[non_exhaustive]
+    pub enum IcaExecute {
+        Evm {
+            // TODO: move types to `abstract-ica` crate
+            msgs: Vec<polytone_evm::evm::EvmMsg<String>>,
+            callback: Option<polytone_evm::callbacks::CallbackRequest>,
```
Buckram123, line 149: Don't think this comment makes sense if we removing abstract-ica
```suggestion
    #[cosmwasm_schema::cw_serde]
    #[non_exhaustive]
    pub enum IcaExecute {
        Evm {
            msgs: Vec<polytone_evm::evm::EvmMsg<String>>,
            callback: Option<polytone_evm::callbacks::CallbackRequest>,
```

Buckram123 (changes requested):
Let's clean those TODO's here

framework/packages/abstract-std/src/native/ica_client.rs
```diff
@@ -0,0 +1,173 @@
...
+use cosmwasm_std::CosmosMsg;
+use cw_ownable::{cw_ownable_execute, cw_ownable_query};
+
+pub use action::{IcaAction, IcaActionResponse, IcaExecute};
+pub use chain_type::{CastChainType, ChainType};
+
+pub use polytone_evm::EVM_NOTE_ID;
+pub use polytone_evm::POLYTONE_EVM_VERSION;
+
+/// This needs no info. Owner of the contract is whoever signed the InstantiateMsg.
```
CyberHoward, line 12: We can get rid of the addresses here and use the deterministic addresses. 

Also would be nice to have an owner field here because it's easier to deploy that way if we want a different owner.

Buckram123: base branch is main here, it's already removed in develop/v2

Kayanski: yes, rebasing right now

Kayanski:
Closed in favor of https://github.com/AbstractSDK/abstract/pull/486 (on version 2)

framework/packages/abstract-std/src/native/ica_client.rs
Kayanski: This is not possible because hf doesn't support keys of the type ARCHWAY[0] it seems here

PR #486, opened 2024-10-07
ICA Client testing

Kayanski:
This Pr aims at moving and testing the ICA client based on Abstract V2
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/Cargo.toml
```diff
@@ -60,6 +60,14 @@ polytone = { package = "abstract-polytone", version = "2.0.0" }
...
 cw-orch-polytone = { package = "abstract-cw-orch-polytone", version = "5.0.0" }
 
+## evm
+evm-note = { version = "0.3.0", features = ["library"] }
+polytone-evm = { version = "0.3.0" }
+# polytone-evm = { git = "https://github.com/abstractsdk/evm-ibc", branch = "development/create-union-bundle" }
+# evm-note = { git = "https://github.com/abstractsdk/evm-ibc", branch = "development/create-union-bundle", features = [
+#   "library",
+# ] }
+
```
Buckram123, line 70:
```suggestion
## evm
evm-note = { version = "0.3.0", features = ["library"] }
polytone-evm = { version = "0.3.0" }
```

Don't think it's really useful, as you won't be able to compile wasms using optimizer with private repo link anyways, right?

framework/contracts/native/ica-client/src/queries.rs
```diff
@@ -10,8 +10,8 @@ pub const PACKET_LIFETIME: u64 = 60 * 60;
 
 pub fn config(deps: Deps, env: &Env) -> IcaClientResult<ConfigResponse> {
     Ok(ConfigResponse {
-        ans_host: AnsHost::new(deps.api, env)?.address,
-        registry_address: RegistryContract::new(deps.api, env)?.address,
+        ans_host_address: AnsHost::new(deps.api, env)?.address.to_string(),
+        registry_address: RegistryContract::new(deps.api, env)?.address.to_string(),
     })
 }
```
Buckram123, line 16: @CyberHoward prefers to return `Addr` from queries if we have verified Address. Let's stick to one standard

framework/Cargo.toml
```diff
@@ -69,6 +77,8 @@ abstract-standalone = { path = "packages/abstract-standalone" }
 
 # Keep these as path, creates cirular dependency otherwise
 # Only need to re-publish all contracts if a re-publish of abstract-interface is required
+
+abstract-client = { version = "0.24.1-beta.1", path = "packages/abstract-client" }
 abstract-interface = { version = "0.24.1-beta.1", path = "packages/abstract-interface" }
 module-factory = { package = "abstract-module-factory", path = "contracts/native/module-factory" }
```
Buckram123, line 83: Just noticed we have `beta.1` version for interface and client

framework/packages/abstract-std/src/native/ibc/ica_client.rs
```diff
@@ -1 +1,170 @@
...
+            funds: Vec<Coin>,
+            // Optional receiver address
+            // Should be formatted in expected formatting
+            // EVM: HexBinary
+            // Cosmos: Addr
+            receiver: Option<Binary>,
+            memo: Option<String>,
+        },
+        // ... other actions?
+    }
```
Buckram123, line 139:
```suggestion
        },
    }
```
Are we still moving TODO's to linear tasks?

framework/packages/abstract-std/src/native/ibc/ica_client.rs
```diff
@@ -1 +1,170 @@
...
+    // pub enum IcaQuery {
+    // 	Evm {
+    // 		// encoded data
+    // 		// ...
+    // 	},
+    // 	Cosmos {
+    // 	    // Encoded data
+    // 		// ...
+    // 	}
+    // }
```
Buckram123, line 163:
```suggestion
         }
```

framework/packages/abstract-std/src/native/ibc/ica_client.rs
```diff
@@ -1 +1,170 @@
...
+
+    pub trait CastChainType {
+        fn chain_type(&self) -> Option<ChainType>;
+    }
+
+    impl CastChainType for TruncatedChainId {
+        // Return the type of chain based on the chain-id.
+        // Note: chain-ids for EVM chains are numbers!
+        fn chain_type(&self) -> Option<ChainType> {
+            let chains = map_macro::hash_map! {
```
Buckram123, line 88: I don't like that we spawn this hashmap on each `chain_type()` call. In theory we shouldn't have to call it more than once, but would be nice to ensure it's not generated more than once

interchain/interchain-end_to_end_testing/src/bin/gov-prop-abstract.rs
Buckram123, line 1: nit: we use snake_case for file names

scripts/wasm-framework.sh
Buckram123, line 1: Why not to use `just wasm` inside `framework`?

Buckram123 (approved):
Great! Would be nice to update chain type lookup table, other than that small nits :rocket:

scripts/wasm-framework.sh
Kayanski: Not sure anymore, deleting

framework/packages/abstract-std/src/native/ibc/ica_client.rs
Kayanski: Ok, something with OnceCell or something. Because static variables can't really be used here, because we are using constants arrays

framework/packages/abstract-std/src/native/ibc/ica_client.rs
Kayanski: yes !

framework/packages/abstract-std/src/native/ibc/ica_client.rs
Buckram123: > Ok, something with OnceCell or something. Because static variables can't really be used here, because we are using constants arrays

Also maybe this https://crates.io/crates/phf, haven't tried, but look like would fit here

Kayanski: I have already tried

Buckram123: > I have already tried and it's not working because they don't support this pattern (ARCHWY[0]) for instance

Ah, that's unfortunate

PR #487, opened 2024-10-08
New feature --> No ID replies

Kayanski:
This PR aims at removing the reply ids all-together from the Abstract Framework. 
The rationale behind that is that they are hard to manage, easy to miss and difficult to maintain.
Instead, the user will register the replies to a specific endpoint by returning the following new structure inside their endpoints: 
```rust
// Old 
// SubMsg::reply_always(cosmos_msg, REPLY_ID)
// New
SubMsg::reply_always(cosmos_msg, reply_func)
``` 

In addition, they will need to register `reply_func` with the `App` object for the rust engine to be able to match it and call it accordingly. 

This allows for code to be easier to read (register a callback function instead of a random ID) and less error-prone.
We could also use this concept for IBC Callbacks 

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/packages/abstract-adapter/src/endpoints/execute.rs
+++ b/framework/packages/abstract-adapter/src/endpoints/execute.rs
@@ -157,7 +157,8 @@ impl<Error: ContractError, CustomInitMsg, CustomExecMsg, CustomQueryMsg, SudoMsg
                 .map_err(|_| unauthorized_sender())?,
         };
         self.target_account = Some(account);
-        self.execute_handler()?(deps, env, info, self, request.request)
+        self.execute_handler()?(deps, env, info, &self, request.request)
+            .map(|response| response.into_cosmwasm_response(self.contract()))
     }
 
     /// Update authorized addresses from the adapter.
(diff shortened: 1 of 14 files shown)


adairrr:
Since it removes reply ids entirely, how would developers build examples with dynamic replies like:

https://github.com/CyberHoward/interchain-gov/blob/main/contracts/interchain-gov/src/contract.rs#L37-L45

adairrr:
> Since it removes reply ids entirely, how would developers build examples with dynamic replies like:
> 
> https://github.com/CyberHoward/interchain-gov/blob/main/contracts/interchain-gov/src/contract.rs#L37-L45

And makes it more difficult to migrate existing cosmwasm contracts to abstract (if this is a concern).

Kayanski:
> > Since it removes reply ids entirely, how would developers build examples with dynamic replies like:
> > https://github.com/CyberHoward/interchain-gov/blob/main/contracts/interchain-gov/src/contract.rs#L37-L45
> 
> And makes it more difficult to migrate existing cosmwasm contracts to abstract (if this is a concern).

It does !

CyberHoward:
Don't think this is worth pursuing either. Seems to add complexity for not too much gain.

Kayanski:
This was a POC, closing as this introduces too much complexity for very little gain

PR #489, opened 2024-10-08
Improve framework coverage v2

Buckram123:
Adding new tests and clearing weird outcomes or errors


framework/contracts/account/src/modules.rs
```diff
@@ -297,52 +294,41 @@ fn configure_adapter(
...
-            } else {
-                true
-            }
-        });
-        Ok::<_, StdError>(whitelisted_modules)
-    })?;
+    whitelisted_modules.retain(|addr| {
+        // retain any addresses that are not in the list of addresses to remove
+        !to_remove_module_addresses.contains(addr)
+    });
```
Buckram123, line 325: Sometimes we do add+remove, for example on UpdateWhitelist action and Upgrade if that's adapter. It was inconsistent when we add or remove first, also takes 2 loads and saves, decided to merge it instead

framework/packages/abstract-sdk/src/cw_helpers/cw_storage_plus.rs
```diff
@@ -1,81 +0,0 @@
...
-{
-    let mut res: Vec<(K, V)> = vec![];
-
-    for key in keys.into_iter() {
-        let value = map.load(storage, key.clone())?;
-        res.push((key, value));
-    }
-
-    Ok(res)
-}
```
Buckram123, line 23: This helper was causing ton of redundant clones in queries. Code is much cleaner without this helper IMO

framework/packages/abstract-sdk/src/apis/traits.rs
```diff
@@ -2,13 +2,33 @@ use cosmwasm_std::Deps;
...
 pub trait AbstractApi<T: ModuleIdentification> {
+    const API_ID: &'static str;
     fn base(&self) -> &T;
     fn deps(&self) -> Deps;
+    /// Get the api identifier.
+    fn api_id() -> String {
+        Self::API_ID.to_owned()
+    }
+}
+
```
Buckram123, line 14: Decided to merge those traits, as they only used together

Buckram123 (commented):
Rest of changes is tests

framework/contracts/account/src/modules.rs
CyberHoward: What happens if an address is both in to-add and to-remove? Should be clarified in a doc-comment.

CyberHoward (approved):

CyberHoward:
Framework tests seem to be failing but I think this was fixed in the interchain improvement PR!

Kayanski:
> Framework tests seem to be failing but I think this was fixed in the interchain improvement PR!

I confirm this is fixed in https://github.com/AbstractSDK/abstract/pull/474

Kayanski (approved):

framework/packages/abstract-std/src/native/registry.rs
```diff
@@ -58,27 +58,8 @@ pub mod state {
...
-
-    /// Primary index for namespaces.
-    pub const NAMESPACES_INFO: IndexedMap<&Namespace, AccountId, NamespaceIndexes> =
-        IndexedMap::new(
-            "nmspc",
-            NamespaceIndexes {
-                account_id: MultiIndex::new(|_pk, d| d.clone(), "nmspc", "nmspc_a"),
-            },
-        );
+    pub const NAMESPACES: Map<&Namespace, AccountId> = Map::new("bbbmb");
```
Kayanski, line 61: Wait what

framework/packages/abstract-std/src/native/registry.rs
```diff
@@ -58,27 +58,8 @@ pub mod state {
     /// Account sequences
     pub const LOCAL_ACCOUNT_SEQUENCE: Item<AccountSequence> =
         Item::new(storage_namespaces::registry::LOCAL_ACCOUNT_SEQUENCE);
-    /// Sub indexes for namespaces.
```
Kayanski, line 61: This is needed, because we use idx at some point, so what's the decision here ?

framework/packages/abstract-std/src/native/registry.rs
Buckram123: > Wait what

Nice catch

framework/packages/abstract-std/src/native/registry.rs
Buckram123: What do you mean? One account can only have one namespace. And indexedmap is overkill just for that

Kayanski: Ok, reverse namespace it is

PR #491, opened 2024-10-09
Smaller wasms

Buckram123:
Changing optimization for wasm size.
- opt-level `3` -> 's'
- Account: ibc_action -> execute_on_module(ibc_client)
- ibc_client: cosmos_sdk msg_transfer -> anybuf msg_transfer

Beyond that updated state file to the deployment that Adair did and tested deployment on local xion without issues


framework/contracts/native/ibc-client/src/anybuf.rs
```diff
@@ -0,0 +1,145 @@
+use anybuf::Anybuf;
```
Kayanski, line 1: Are we sure this removes a lot of wasm size ? Doesn't seem particularly healthy to provide our own ibc-proto types

framework/packages/abstract-interface/src/deployment.rs
```diff
@@ -354,7 +354,7 @@ mod test {
     fn have_some_state() {
         State::get("state.json").unwrap();
         let state = State::load_state();
-        let vc_juno = &state["juno-1"]["code_ids"].get(REGISTRY);
+        let vc_juno = &state["pion-1"]["code_ids"].get(ANS_HOST);
```
Kayanski, line 357: That's doesn't seem to be the contract you're trying to query (or mayeb the var name is not good)

framework/packages/abstract-std/src/native_addrs.rs
```diff
@@ -16,7 +16,7 @@ const TEST_ABSTRACT_CREATOR: [u8; 33] = [
 
 // Salts for deployments
 pub const ANS_HOST_SALT: &[u8] = b"ans";
-pub const VERSION_CONTROL_SALT: &[u8] = b"vc";
+pub const REGISTRY_SALT: &[u8] = b"vc";
```
Kayanski, line 19: can't we change the salt to reg instead or is it a task for another release ?

Kayanski (approved):
Very nice, a few small changes to clarify

framework/Cargo.toml
```diff
@@ -112,7 +112,7 @@ opt-level = 3
 rpath = false
 lto = true
 overflow-checks = true
-opt-level = 3
+opt-level = 's'
```
Kayanski, line 115: Oh, can we also change the opt-level in all crate profiles so that it's not a source of confusion ?

Buckram123: Workspace overrides profile of packages, I'll remove package profiles to remove confusion and warning from cargo

framework/contracts/native/ibc-client/src/anybuf.rs
Buckram123: Yep, a lot, with opt-level =3 wasm don't fit, but with this change we have extra ~20 kb (It's still not enough if we ever decide to add any method)

framework/Cargo.toml
CyberHoward: > Workspace overrides profile of packages, I'll remove package profiles to remove confusion and warning from cargo

During our last audit Oak recommended having the profile in multiple places to ensure that even if one is remove somehow, the profile still applies (explicitly because of the overflow checks that could be omitted if not set)

CyberHoward (approved):

PR #492, opened 2024-10-10
Creator pubkey script

Buckram123:
Small helper script to get pubkey bytes of the abstract creator


CyberHoward:
Would it make sense to have a `build.rs` with an optional env-var that overwrites the pub-key? 

`ABSTRACT_DEPLOYER_ADDR`. If it's set we compute the addr bytes and write to the const file.

Kayanski (approved):
Very nice, could you document the usage of this script ? Or have the current deployment flow documented (doesn't have to be now, but as soon as possible) ?

Buckram123:
> Would it make sense to have a `build.rs` with an optional env-var that overwrites the pub-key?
> 
> `ABSTRACT_DEPLOYER_ADDR`. If it's set we compute the addr bytes and write to the const file.

After setting it user of it will still have to recompile it.

What do you think about updating this script later to also optionally override if deployer addr provided and use it with justfile wasm actions

`just wasm $addr`

CyberHoward:
> > Would it make sense to have a `build.rs` with an optional env-var that overwrites the pub-key?
> > `ABSTRACT_DEPLOYER_ADDR`. If it's set we compute the addr bytes and write to the const file.
> 
> After setting it user of it will still have to recompile it.
> 
> What do you think about updating this script later to also optionally override if deployer addr provided and use it with justfile wasm actions
> 
> `just wasm $addr`

Yup, sounds good!

PR #494, opened 2024-10-11
Remove funds from IBC client SendFunds

CyberHoward:
This PR removes `funds` field from the IBC client SendFunds function. `SendFunds` now uses `msg_info.funds` instead.

### DIFF
--- a/framework/contracts/native/ibc-client/src/commands.rs
+++ b/framework/contracts/native/ibc-client/src/commands.rs
@@ -371,7 +371,6 @@ pub fn execute_send_funds(
     env: Env,
     info: MessageInfo,
     host_chain: TruncatedChainId,
-    funds: Vec<Coin>,
     memo: Option<String>,
 ) -> IbcClientResult {
     host_chain.verify()?;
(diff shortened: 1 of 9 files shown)


Kayanski:
Wait so the user can't sent funds that are located inside their account already ? Or in a complicated way only ?

CyberHoward:
Sorry, description wasn't very clear. 

Instead of using the `funds` field specified within the message, this pr removes it to just use the `msg_info.funds` instead. That way we don't have to replicate that data.

framework/contracts/native/ibc-client/src/commands.rs
```diff
@@ -397,7 +396,7 @@ pub fn execute_send_funds(
     let ics20_channel_id = ics20_channel_entry.resolve(&deps.querier, &ans)?;
 
     let mut transfers: Vec<CosmosMsg> = vec![];
-    for coin in funds {
+    for coin in info.funds.into_iter() {
```
Kayanski, line 399:
```suggestion
    for coin in info.funds {
```

Kayanski (approved):
LGTM

PR #495, opened 2024-10-14
Added accounts query and tests

Kayanski:
This Pr aims at 
- changing the `Account` query on the registry contract to an `Accounts` query
- Adding an paginated `AccountList` query on the registry contract

### Checklist

- [ ] CI is green.
- [x] Changelog updated.

### DIFF
--- a/framework/contracts/account/tests/create.rs
+++ b/framework/contracts/account/tests/create.rs
@@ -78,7 +78,7 @@ fn create_one_account() -> AResult {
 
     assert_that!(&vc_config).is_equal_to(&expected);
 
-    let account_list = registry.get_account(TEST_ACCOUNT_ID)?;
+    let account_list = registry.account(TEST_ACCOUNT_ID)?;
 
     assert_that!(&account_list.into()).is_equal_to(Account::new(account));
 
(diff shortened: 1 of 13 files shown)


CyberHoward (approved):

framework/contracts/account/tests/create.rs
```diff
@@ -78,9 +78,9 @@ fn create_one_account() -> AResult {
 
     assert_that!(&vc_config).is_equal_to(&expected);
 
-    let account_list = registry.account(TEST_ACCOUNT_ID)?;
+    let account_list = registry.get_account(TEST_ACCOUNT_ID)?;
```
Buckram123, line 81: Are we keeping `get_` prefix for the interface?

Buckram123 (approved):
LGTM

framework/contracts/account/tests/create.rs
CyberHoward: we could do just `account` as we changed the name of the actual query to `accounts`

CyberHoward: renamed it

PR #497, opened 2024-10-14
Remove speculoos from dependency

Kayanski:
This PR aims at removing speculoos from the testing framework and replacing it with assertor that has very similar syntax and is well maintained for now (although not stable).

The changes are wide, so a lot of changes are needed

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


CyberHoward:
@Buckram123 and I discussed this and we wanted to use default assertions that come with Rust.

Buckram123:
Closed in favor of #501

PR #500, opened 2024-10-16
Separate xion abstract account from account

Buckram123:
This PR separates xion abstract account implementation from abstract account into new `abstract-xion` crate and removes cosmos-sdk-proto dependency to make this package publishable to crates.io


framework/contracts/account/Cargo.toml
```diff
@@ -23,22 +23,7 @@ crate-type = ["cdylib", "rlib"]
 [features]
 default = ["export"]
 export = []
-xion = [
```
CyberHoward, line 26: ahhh much better :)

CyberHoward (approved):

PR #501, opened 2024-10-16
Remove speculoos from dependency

Buckram123:
This PR removes speculoos from dependency
- [x] framework
- [x] modules
- [x] interchain

There are few assert_that!'s left in modules but they are non-testable because we don't currently have wyndex-bundle


Buckram123:
There are few `assert_that!`'s left in modules, but they are non-testable because we don't currently have wyndex-bundle

Kayanski (approved):
Very nice ! 
The only issue I see is that debugging a test that fails is now more difficult because assertions are not very verbose when they fail, but that's ok I guess

PR #502, opened 2024-10-17
Release 0.24.1 beta.2

Buckram123:
Another version patch but fully released this time! Had to publish polytone with cw v2 in the process


framework/Cargo.toml
```diff
@@ -57,7 +57,7 @@ cw-orch-interchain = { version = "0.7.0" }
 tokio = { version = "1.4", features = ["full"] }
 
 polytone = { package = "abstract-polytone", version = "2.0.0" }
-polytone-note = { package = "abstract-polytone-note", version = "4.0.0" }
```
CyberHoward, line 60: I'm assuming both 4.0.0 and 4.0.1 are cw 2.x?

CyberHoward (approved):

framework/Cargo.toml
Buckram123: I failed during previous polytone deployment, so 4.0.0 is cw v1.x and it was yanked right before 3.0.0 released

PR #503, opened 2024-10-17
Added failing test

Kayanski:
This PR aims at making sure that new modules are installed on an existing account with install_on_subaccount enabled.

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/packages/abstract-client/tests/integration.rs
+++ b/framework/packages/abstract-client/tests/integration.rs
@@ -1420,6 +1420,7 @@ fn retrieve_account_builder_install_missing_modules() -> anyhow::Result<()> {
 
     app_publisher.publish_app::<MockAppI<MockBech32>>()?;
 
+    assert!(!app_publisher.account().module_installed(TEST_MODULE_ID)?);
     let account = client
         .account_builder()
         .namespace(Namespace::new(TEST_NAMESPACE)?)
@@ -1550,5 +1551,45 @@ fn ans_balance() -> anyhow::Result<()> {
         .name_service()
         .balance(&chain.sender_addr(), &AssetEntry::new("mock"))?;
     assert_eq!(balance, Uint128::new(101));
+    Ok(())
+}
+
+// Tests wether using the Account builder with install_on_subaccount on an existing account installs given apps
+#[test]
+fn account_builder_doesnt_install_module_on_existing_account() -> anyhow::Result<()> {
+    let chain = MockBech32::new("mock");
+    let client = AbstractClient::builder(chain.clone()).build_mock()?;
+
+    let app_publisher: Publisher<MockBech32> = client
+        .publisher_builder(Namespace::new(TEST_WITH_DEP_NAMESPACE)?)
+        .build()?;
+
+    let app_dependency_publisher: Publisher<MockBech32> = client
+        .publisher_builder(Namespace::new(TEST_NAMESPACE)?)
+        .build()?;
+
+    app_dependency_publisher.publish_app::<MockAppI<_>>()?;
+    let res = app_publisher.publish_app::<MockAppWithDepI<_>>();
+    assert!(res.is_ok());
+
+    const NEW_NAMESPACE: &str = "new-namespace";
+
+    // We create an account
+    let account = client
+        .account_builder()
+        .namespace(NEW_NAMESPACE.try_into()?)
+        .build()?;
+    assert!(!account.module_installed(TEST_MODULE_ID)?);
+
+    // We now re-use the builder to install a module (the usage is usually in a separate script)
+    client
+        .account_builder()
+        .namespace(NEW_NAMESPACE.try_into()?)
+        .install_on_sub_account(true) // Making sure this is enabled,
+        .install_app::<MockAppWithDepI<MockBech32>>(&MockInitMsg {})?
+        .build()?;
+
+    // We assert the app is installed on the account
+    assert!(account.module_installed(TEST_MODULE_ID)?);
     Ok(())
 }


framework/packages/abstract-client/tests/integration.rs
```diff
@@ -1552,3 +1570,42 @@ fn ans_balance() -> anyhow::Result<()> {
...
+    // We create an account
+    let account = client
+        .account_builder()
+        .namespace(NEW_NAMESPACE.try_into()?)
+        .build()?;
+
+    client.fetch_or_build(NEW_NAMESPACE.try_into()?, |builder| {
+        builder
+            .install_app::<MockAppWithDepI<MockBech32>>(&MockInitMsg {})
+            .unwrap()
```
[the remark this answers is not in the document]
Kayanski: Yeah but in the case you don't want to install app for instance, you will have to return a wrapped Ok, wchich is very ugly ! So that's an oustading question, what do you think ?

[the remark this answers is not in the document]
Buckram123: I would argue that `.unwrap()` is ugly as well! Also handling result is more convenient, than handling the panic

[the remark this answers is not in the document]
Kayanski: One solution I had in mind, is to remove the Result from the install_app method, what do you think ?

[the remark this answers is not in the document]
Buckram123: > One solution I had in mind, is to remove the Result from the install_app method, what do you think ?

Sure, we can move all of the checks inside `build()`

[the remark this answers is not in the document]
Kayanski: Ah this is an idea yes. I was proposing to do unwraps inside the `AccountBuilder::install_app` instead but your suggestion makes sense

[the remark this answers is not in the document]
Kayanski: I decided to go with `unwrap` inside `isntall_app` methods because : 
1. It's very difficult to handle the Vec<Result<ModuleInstallConfig, _>> because of clone bounds on errors mainly
2. It's ok to have that operation panic, because this is mainly parsing error and this is never on the blockchain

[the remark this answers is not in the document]
Buckram123: > I decided to go with `unwrap` inside `isntall_app` methods because :
> 
>     1. It's very difficult to handle the Vec<Result<ModuleInstallConfig, _>> because of clone bounds on errors mainly
> 
>     2. It's ok to have that operation panic, because this is mainly parsing error and this is never on the blockchain

We can update and figure it out later if someone would be unhappy with this panic. Removing results from installs already nice!

framework/packages/abstract-client/tests/integration.rs
```diff
@@ -1552,3 +1570,42 @@ fn ans_balance() -> anyhow::Result<()> {
     assert_eq!(balance, Uint128::new(101));
     Ok(())
 }
+
+// Tests wether using the Account builder with install_on_subaccount on an existing account installs given apps
```
CyberHoward, line 1574: update doc-comment

framework/packages/abstract-client/src/account.rs
```diff
@@ -745,6 +673,14 @@ impl<Chain: CwEnv> Account<Chain> {
         self.module_installed(IBC_CLIENT)
     }
 
+    /// Builder to create a subaccount from this account
+    pub fn sub_account_builder(&self) -> AbstractClientResult<AccountBuilder<Chain>> {
+        let mut builder = AccountBuilder::new(&self.infrastructure()?);
```
Buckram123, line 678: Should we unwrap here as well? If `infrastructure` errors don't think there's anything you can do really

framework/packages/abstract-client/src/publisher.rs
```diff
@@ -3,106 +3,52 @@
...
+        })
+    }
+
+    /// New publisher from account and namespace
+    /// Claim the namespace from the account
+    pub fn new_with_namespace(
+        account: Account<Chain>,
+        namespace: &str,
+    ) -> AbstractClientResult<Self> {
+        account.claim_namespace(namespace)?;
```
Buckram123, line 48: What if it already have claimed namespace, will it error?

Buckram123: I would prefer not, to have more returnable script

Buckram123 (approved):
Small error handling comments

framework/packages/abstract-client/src/account.rs
Kayanski: Yes

framework/packages/abstract-client/src/publisher.rs
Kayanski: What do you think now ?

Kayanski: https://github.com/AbstractSDK/abstract/pull/503/commits/853a4ba5267e3d83e7ddab451f32b4cdc076f29e#diff-0908c6e723f19d3802d488bc582b29bb0da58a0330aaa450fcfe6e70e5bcee13R49

framework/packages/abstract-client/src/publisher.rs
```diff
@@ -45,7 +46,19 @@ impl<Chain: CwEnv> Publisher<Chain> {
...
+            if namespace_info.account_id != account.id()?{
+                return Err(AbstractClientError::NamespaceClaimed{
+                    namespace: namespace.to_string(),
+                    account_id: namespace_info.account_id
+                })
+            }
+        }else{
+            account.claim_namespace(namespace)?;
+        }
+
```
Buckram123, line 61: Nice!

PR #506, opened 2024-10-17
Remove unused fields and variants

Buckram123:
- Removed unused variant from base execute message for the app: `UpdateConfig{}`. Which means BaseExecuteMsg for the app is empty enum and cannot be created. Couldn't replace it to `cosmwasm_std::Never` because it doesn't implement `Serialize`, but we need to serialize execute message of the app

- Removed `registry_address` field for the `GovernanceDetails` methods


framework/packages/abstract-std/src/objects/gov_type.rs
```diff
@@ -81,22 +81,23 @@ pub enum GovAction {
...
+    pub fn verify(self, deps: Deps) -> Result<GovernanceDetails<Addr>, AbstractError> {
         match self {
             GovernanceDetails::Monarchy { monarch } => {
                 let addr = deps.api.addr_validate(&monarch)?;
                 Ok(GovernanceDetails::Monarchy { monarch: addr })
             }
             GovernanceDetails::SubAccount { account } => {
                 let account_addr = deps.api.addr_validate(&account)?;
+
+                let hrp = native_addrs::hrp_from_address(&account_addr);
```
Kayanski, line 93: Love this logic here, not that clean, but smart

framework/packages/abstract-std/src/objects/ownership/README.md
```diff
@@ -12,32 +12,38 @@ Initialize the owner during instantiation using the `initialize_owner` method pr
 
 ```rust
 use cosmwasm_std::{entry_point, DepsMut, Env, MessageInfo, Response, Empty};
-use abstract_std::objects::ownership::OwnershipError;
+use abstract_std::objects::ownership::{initialize_owner, GovOwnershipError, GovernanceDetails};
```
Kayanski, line 15: Oh, you even fixed those, that's good !

Kayanski (approved):
LGTM

CyberHoward:
> * Removed unused variant from base execute message for the app: `UpdateConfig{}`. Which means BaseExecuteMsg for the app is empty enum and cannot be created. Couldn't replace it to `cosmwasm_std::Never` because it doesn't implement `Serialize`, but we need to serialize execute message of the app
Have you tested this with the schema gen? Thinking it might be an edge-case that can cause some issues.

Buckram123:
> > * Removed unused variant from base execute message for the app: `UpdateConfig{}`. Which means BaseExecuteMsg for the app is empty enum and cannot be created. Couldn't replace it to `cosmwasm_std::Never` because it doesn't implement `Serialize`, but we need to serialize execute message of the app
> 
> Have you tested this with the schema gen? Thinking it might be an edge-case that can cause some issues.
> 
> edit: closed pr on accident

```
    "BaseExecuteMsg": {
      "type": "string",
      "enum": []
    },
```
from 
```
    "BaseExecuteMsg": {
      "oneOf": [
        {
          "description": "Updates the base config",
          "type": "object",
          "required": [
            "update_config"
          ],
          "properties": {
            "update_config": {
              "type": "object",
              "additionalProperties": false
            }
          },
          "additionalProperties": false
        }
      ]
}
```
Looks fine, we are not supposed to use it anyway if it cannot be created

PR #509, opened 2024-10-21
Remove `prost` from `abstract-xion` dependency

Buckram123:
Using anybuf instead still doesn't fit into 819 000 bytes


Kayanski:
Closed, because we removed abstract-xion within the workspace, now depending solely on abstract-xion published crate and original github repo (https://github.com/burnt-labs/contracts)

PR #511, opened 2024-10-23
Abstract Client: Admin execute on remote application

Buckram123:
Adding this helper to avoid boilerplate

For some context: https://github.com/AbstractSDK/templates/blob/dd6aafa5a6437ad2a9a44ac75022b606cb878a45/contracts/%7B%7Bapp_name%7D%7D/tests/interchain_integration.rs#L78-L88 This test fails on admin check


Buckram123:
Alternatively we can make `execute` always admin for remote in client?

PR #512, opened 2024-10-24
Update osmosis tests

Buckram123:
Now that osmosis and test-tube updated to cosmwasm v2 we can enable integration and tests back

This PR introduces mock binaries and `mock-deploy` feature with open mnemonic to the abstract, it allows publishing abstract in a mocked chains: local, test tubes or starship


Buckram123:
Depends on https://github.com/AbstractSDK/cw-orchestrator/pull/511

framework/packages/abstract-app/src/interface.rs
```diff
@@ -219,9 +219,19 @@ macro_rules! __cw_orch_interfaces__ {
...
+                    {
+                        cw_orch::build::BuildPostfix::None
+                    }
+                };
                 let wasm_name = env!("CARGO_CRATE_NAME").replace('-', "_");
                 ::cw_orch::prelude::ArtifactsDir::auto(Some(env!("CARGO_MANIFEST_DIR").to_string()))
-                    .find_wasm_path(&wasm_name)
+                    .find_wasm_path_with_build_postfix(&wasm_name, build_postfix)
                     .unwrap()
             }
```
Buckram123, line 236: Not sure about this one, but I don't know how to solve it better in current implementation.
Issue that it will require the module to have exactly `module-deployment` feature. This issue comes out because we generate this code with macro and feature checks is happening after macro gets evaluated inside user crate, and not on `abstract-app` level

CyberHoward: @Buckram123 so this will be removed with the new solution right?

Buckram123: > @Buckram123 so this will be removed with the new solution right?

Drafting this PR for now. But yes, hopefully all magic would be limited in a build script

Buckram123:
Closed in favor #528

PR #518, opened 2024-10-30
Update osmosis tests(with modified wasm)

Buckram123:
Alternative to #512

Here we use build script to change creator, instead of adding a new feature.

From the user perspective.
This approach:
- have to set ABSTRACT_CREATOR environment to seed or public key bytes before compiling the contract
- all of the contract dependencies(for example dex adapter) should be compiled with same ABSTRACT_CREATOR
- No indication if the wasm was built with or without custom `ABSTRACT_CREATOR`

#512 :
- Have to set feature on `abstract-MODULE` and all of the abstract module dependencies (for example dex adapter)
- Have to have same feature name as enabled in `abstract-MODULE` (right now feature named `mock-deployment`)
- Have to add mock build inside [package.metadata.optimizer] section
- Two sets of the wasm, mock deployment is clearly visible

It's not a final, before I finish would be nice to have final approve on one of the methods
TODO, if this method gets approved:
- Figure out where we want to set the ENV. Currently I like setting it in `.cargo/config.toml#env` section, because this file usually public and also will be moved inside docker(so no need to set it separately for the docker), but it's easy to deploy the wrong version this way
- If we rely on custom `ABSTRACT_CREATOR` in tests, tests that use wasms, will fail in release version, are we fine with that, anything we can do with it?


framework/contracts/native/registry/src/contract.rs
```diff
@@ -46,11 +46,9 @@ pub fn instantiate(deps: DepsMut, _env: Env, _info: MessageInfo, msg: Instantiat
...
 
     // Save the abstract namespace to the Abstract admin account
-    NAMESPACES.save(
-        deps.storage,
-        &Namespace::new(ABSTRACT_NAMESPACE)?,
-        &ABSTRACT_ACCOUNT_ID,
-    )?;
+    let namespace = Namespace::new(ABSTRACT_NAMESPACE)?;
+    NAMESPACES.save(deps.storage, &namespace, &ABSTRACT_ACCOUNT_ID)?;
+    REV_NAMESPACES.save(deps.storage, &ABSTRACT_ACCOUNT_ID, &namespace)?;
```
Buckram123, line 51: Not related to this PR, but had to fix it to fetch abstract account by namespace. We didn't save abstract account in REV_NAMESPACE

framework/contracts/native/registry/src/migrate.rs
```diff
@@ -47,7 +51,14 @@ pub fn migrate(deps: DepsMut, env: Env, msg: MigrateMsg) -> VCResult {
...
                 .collect::<StdResult<Vec<_>>>()?;
-            for (namespace, account_id) in namespaces_info {
+            for (namespace, account_id) in namespaces_info
+                .into_iter()
+                // Make sure abstract included
+                .chain(std::iter::once((
+                    Namespace::new(ABSTRACT_NAMESPACE)?,
+                    ABSTRACT_ACCOUNT_ID,
+                )))
+            {
```
Buckram123, line 61: And to ensure it's fixed after migration

framework/packages/abstract-interface/build.rs
```diff
@@ -43,8 +56,54 @@ fn main() {
...
+    let read_dir = fs::read_dir(artifacts_path).expect("You should create an artifacts dir in your crate to export the wasm files along with the cw-orch library");
+
+    // Path to OUT_DIR
+    let out_dir = env::var("OUT_DIR").unwrap();
+
+    let creator = if let Ok(creator) = env::var("ABSTRACT_CREATOR") {
+        let public_key = bip32::Mnemonic::new(&creator, Default::default())
+            .map(|phrase| {
+                let seed = phrase.to_seed("");
+                let derive_path: bip32::DerivationPath = "m/44'/118'/0'/0/0".parse().unwrap();
```
Buckram123, line 68: Do we have to add `coin_type` or `derive_path` as another env? IMO it's an overkill until someone needs it

framework/packages/abstract-interface/src/deployment.rs
```diff
@@ -74,91 +74,7 @@ impl<Chain: CwEnv> Deploy<Chain> for Abstract<Chain> {
...
-        #[cfg(feature = "integration")]
-        use abstract_std::objects::gov_type::GovernanceDetails;
-        #[cfg(feature = "integration")]
-        AccountI::create_default_account(
-            &deployment,
-            GovernanceDetails::Monarchy {
-                monarch: chain.sender_addr().to_string(),
-            },
-        )?;
+        let mut deployment = Self::_deploy(&chain)?;
```
Buckram123, line 77: Was thinking about having second deploy function, like mock_deploy, where we allow changing creator, and in default we check if it matches the original

modules/contracts/adapters/dex/tests/osmosis.rs
```diff
@@ -86,27 +76,45 @@ fn get_pool_token(id: u64) -> String {
     format!("gamm/pool/{}", id)
 }
 
+const MOCK_MNEMONIC: &str = "clip hire initial neck maid actor venue client foam budget lock catalog sweet steak waste crater broccoli pipe steak sister coyote moment obvious choose";
+
```
Buckram123, line 80: we can get it from env, but what do we do if it's not set?

Buckram123:
Closed in favor #528

PR #519, opened 2024-10-30
Fix abstract client fetch methods

Buckram123:
We can fetch from AccountSource instead of always from namespace.

Beyond that noticed that we use two versions of bech32, updated the one in abstract-xion


framework/packages/abstract-client/src/client.rs
```diff
@@ -174,29 +174,37 @@ impl<Chain: CwEnv> AbstractClient<Chain> {
...
-        )
+    /// Fetch an [`Account`] from a given source.
+    ///
+    /// This method is used to retrieve an account from a given source. It will **not** create a new account if the source is invalid.
+    ///
+    /// Sources that can be used are:
+    /// - [`Namespace`]: Will retrieve the account from the namespace if it is already claimed.
+    /// - [`AccountId`]: Will retrieve the account from the account id.
+    /// - App [`Addr`]: Will retrieve the account from an app that is installed on it.
+    pub fn fetch_account<T: Into<AccountSource>>(
```
Kayanski, line 185: this can be removed actually, it doesn't make sense to have that

Kayanski (approved):

framework/packages/abstract-client/src/client.rs
Buckram123: > this can be removed actually, it doesn't make sense to have that

We should remove this or `account_from`, what would you prefer to remove? @CyberHoward

Kayanski: `account_from` could be mistaken for account creation where fetch_account is clear in that sens

Kayanski: Lovely

Kayanski:
This is a breaking change though no ?

Buckram123:
> This is a breaking change though no ?

Yes, but `fetch_or_build_account` is quite hard to use without this change IMO
- Build fn doesn't support environment capturing
- You can't create or fetch account without a namespace

So this method was rather for fast creating or loading a publisher, than for customizable account

PR #520, opened 2024-10-30
Renaming and Reorganizing

Kayanski:
This PR aims at re-organizing the repo to simplify the structure and re-group things that belong together (especially tests)

Changes done: 
1. Rename abstract-testing to `abstract-unit-test-utils`
2. Rename abstract-integration-tests to `abstract-integration-test-utils`
3. Move framework_clone_testing to abstract tests crate
4. Move interchain tests to abstract tests crate
5. Move module clone testing tests to abstract test crate

## Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/Cargo.toml
```diff
@@ -55,6 +56,9 @@ clap = { version = "4.0.32", features = ["derive"] }
 semver = "1.0"
 cw-orch = { version = "0.27.0" }
 cw-orch-interchain = { version = "0.8.1" }
+cw-orch-clone-testing = { version = "0.9.0" }
```
Buckram123, line 59: I assume clone test in framework adds many dependencies in this workspace, are we fine with it?

Buckram123 (approved):
Nice to have abstract-tests package! Sometimes it was hard to find location for framework test, now it will be easier :)

framework/Cargo.toml
```diff
@@ -71,6 +75,7 @@ abstract-standalone = { path = "packages/abstract-standalone" }
 # Keep these as path, creates cirular dependency otherwise
 # Only need to re-publish all contracts if a re-publish of abstract-interface is required
 abstract-interface = { version = "0.24.1-beta.1", path = "packages/abstract-interface" }
+abstract-client = { version = "0.24.1-beta.1", path = "packages/abstract-client" }
```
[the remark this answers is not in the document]
Kayanski: Ok, removing the versions here

framework/Cargo.toml
Kayanski: Yes, that's also an issue. Do you think feature flagging those tests would be a good idea ?

Buckram123 (changes requested):
Even though this structure makes sense, we should value time used for compiling our main workspace `framework` much higher than structure of the tests. And we should keep more time-consuming tests with it's heavy dependencies inside `interchain` or different workspace.

Let's improve our development experience! Wdyt?

Kayanski:
> Even though this structure makes sense, we should value time used for compiling our main workspace `framework` much higher than structure of the tests. And we should keep more time-consuming tests with it's heavy dependencies inside `interchain` or different workspace.
> 
> Let's improve our development experience! Wdyt?

I did this change and then realized that, yes, let's close this PR because of the compile times needed for the framework

PR #522, opened 2024-11-01
Ignore beta versions in modules smart query

Buckram123:
It does make it so the user won't be using prerelease versions as default for `Latest`. But it makes it harder to use and test beta versions


Buckram123:
> For the future (next breaking version, where we might need beta) do you know if there is a better solution that changes the storage layout for instance ?

Not yet, would be nice to study at how versions usually sorted

Kayanski (approved):

PR #523, opened 2024-11-04
Ibc client and ibc host removal of predictable addresses

Buckram123:
We can't guarantee having non-breaking API for those modules, so making predictable addresses doesn't make sense, as it disallows having multiple versions of ibc-client/ibc-host registered at the same time


Kayanski (approved):
LGTM

PR #524, opened 2024-11-04
Rev namespace for abstract

Buckram123:
from 24613ed087ce3fc10ad36ad52c6cdfb797010f9a (see https://github.com/AbstractSDK/abstract/pull/518#discussion_r1822440667)


Kayanski:
Nice catch, do you need a new deployment right away for this contract ? You could publish a patch if you want

PR #525, opened 2024-11-05
Added xion crate instead of local xion

Kayanski:
This PR aims at remove abstract-xion crate and using the published crates.io dependency instead

Blocked by xion publishing their account crate

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Buckram123 (approved):
LGTM, have you checked size of the wasm with this change?

Kayanski:
> LGTM, have you checked size of the wasm with this change?

No I haven't

PR #526, opened 2024-11-05
Minify test artifacts

Buckram123:
We have enabled optimization for dev dependencies. Let's try disabling optimization on it and see if we have enough space


Buckram123:
It makes it failing again instead :open_mouth:

PR #528, opened 2024-11-06
Update osmosis tests, using code id to predict addresses

Buckram123:
This PR aims to refactor predicting addresses to relying on creator of the code id instead of hardcoded values in wasms


framework/packages/abstract-std/src/native_addrs.rs
```diff
@@ -1,59 +1,64 @@
...
+        deps.api
+            .addr_canonicalize(creator_addr.as_str())
+            .map(Into::into)
+    })
+    //             ?,
+    //         )
+    //     }
+    //     Ok(CREATOR_CANON_ADDRESS.clone().unwrap())
+    // }
 }
```
Buckram123, line 54: Commented part breaks on multichain, would also be unsafe in multi-threaded(not applicable to chains). What is ideal outcome for this?

framework/Cargo.toml
```diff
@@ -124,9 +124,9 @@ incremental = false
 [patch.crates-io]
 
 # In case Polytone not released
-# abstract-polytone = { git = "https://github.com/AbstractSDK/polytone.git", branch = "bump/cw2" }
-# abstract-polytone-note = { git = "https://github.com/AbstractSDK/polytone.git", branch = "bump/cw2" }
-# abstract-cw-orch-polytone = { git = "https://github.com/AbstractSDK/polytone.git", branch = "bump/cw2" }
+abstract-polytone = { git = "https://github.com/AbstractSDK/polytone.git", branch = "bump/cw2" }
```
Kayanski, line 127: Re-enabling those ?

framework/contracts/account/src/modules.rs
```diff
@@ -211,7 +214,10 @@ pub fn uninstall_module(
     crate::versioning::remove_as_dependent(deps.storage, &module_id, module_dependencies)?;
 
     // Remove for account if needed
-    let registry = RegistryContract::new(deps.api, env)?;
+    let contract_info = deps
```
Kayanski, line 217: Not using once-cell here ?

framework/contracts/account/src/queries.rs
```diff
@@ -73,8 +76,11 @@ pub fn handle_module_info_query(
 
     let ids_and_addr = res?;
 
-    let registry =
-        RegistryContract::new(deps.api, env).map_err(|e| StdError::generic_err(e.to_string()))?;
+    let contract_info = deps
```
Kayanski, line 79: why so much code repeats !!

Buckram123:
As a nice bonus this PR significantly decreases wasm sizes

Kayanski (changes requested):
That's very good thank you. Some remarks: 
- We can hide and potentially cache more thing from the users (for instance the contract info query that you do EVERYWHERE)
- We need to discuss storing the creator address in a once-cell or something simiar

Kayanski:
> As a nice bonus this PR significantly decreases wasm sizes

Why though ? haha seems weird to me

framework/packages/abstract-std/src/native_addrs.rs
Kayanski: Ideal outcome would be : 
- Working inside tests with multi-chain
- Able to cache this (and more) queries during contract execution to save gas

Kayanski: We could work with once-cell and hashmaps, but not sure this is the best idea

Buckram123: OnceLock makes it safe in multi thread. Don't think we supposed to use hashmap(unless feature flagged?) for storing creator in static.

Buckram123:
> > As a nice bonus this PR significantly decreases wasm sizes
> 
> Why though ? haha seems weird to me

Bech32 generation quite expensive!

framework/Cargo.toml
Buckram123: Oopsie, merged it from previous branch

framework/contracts/account/src/modules.rs
Buckram123: Using just mut statics for now

framework/contracts/account/src/queries.rs
Buckram123: It's not like we can avoid it that much in our native contracts, we need to get contract info and code info. For modules it's abstracted away!

Kayanski: We could have a method that does `Registry::new_from_self_code_id` os something no ?

Buckram123: > We could have a method that does `Registry::new_from_self_code_id` os something no ?

Then we can't re-use code-id

Kayanski: ok got it

Kayanski:
Is this branch now closing 
https://github.com/AbstractSDK/abstract/pull/512
and 
https://github.com/AbstractSDK/abstract/pull/518
?

Buckram123:
> This is wonderful. We need to bump to 0.25 version though as this is VERY breaking

Should we hold a bit and merge it when we ready for 0.25 release, in case there will be more 0.24.x patches?

Kayanski:
> > This is wonderful. We need to bump to 0.25 version though as this is VERY breaking
> 
> Should we hold a bit and merge it when we ready for 0.25 release, in case there will be more 0.24.x patches?

Yes ! This is not pressing for testnet

PR #530, opened 2024-11-07
Registry 0.24.3 release

Buckram123:
"abstract" registration in rev namespaces. See #524


Buckram123:
Do we want git tag at this point?

Kayanski:
> Do we want git tag at this commit?

I think it's good practice in case we want to come back to this change and main is breaking

Kayanski (approved):

Buckram123:
> > Do we want git tag at this commit?
> 
> I think it's good practice in case we want to come back to this change and main is breaking

Done

PR #531, opened 2024-11-07
Ibc client and ibc host deployment updates

Buckram123:
- Checking if version is breaking got fixed.
- Host and Client now uploads and migrate with `upload_and_migrate_if_needed` like other contracts, instead of first uploading and migrating after that
- marked with deprecated `Abstract::migrate_if_needed` and `AbstractIbc::migrate_if_needed`


framework/packages/abstract-interface/src/migrate.rs
```diff
@@ -238,24 +241,16 @@ impl<Chain: CwEnv> AbstractIbc<Chain> {
...
-            .upload_if_needed()?
-            .expect("IBC client wasm might be outdated");
-        self.host
-            .upload_if_needed()?
-            .expect("IBC host wasm might be outdated");
-
-        // Check if version is breaking
-        let version_req = semver::VersionReq::parse(&ibc_client_cw2_version).unwrap();
-        let new_version = semver::Version::parse(::ibc_client::contract::CONTRACT_VERSION).unwrap();
-        if version_req.matches(&new_version) {
```
Kayanski, line 252: Why are we not using the matches here that is already implemented by the semver crate ?

Buckram123: Couldn't figure out what structure of  VersionReq would work the way we want. Closest is `~` but it doesn't count pre-release as breaking versions

Kayanski: Are you sure ? 
https://docs.rs/semver/latest/src/semver/eval.rs.html#3

PR #533, opened 2024-11-11
Added migration function from xion accounts

Kayanski:
This PR aims at modifying the account migration function to be able to migrate xion accounts to the Abstract Account

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/contracts/account/Cargo.toml
Buckram123, line 1: What's going on with tabs in this file? Are you using taplo for formatting?

framework/contracts/account/src/migrate.rs
```diff
@@ -1,19 +1,121 @@
...
-    assert_contract_upgrade(deps.storage, ACCOUNT, version)?;
+    // If we already have an abstract account, we just migrate like normal
+    if assert_contract_upgrade(deps.storage, ACCOUNT, version).is_ok() {
+        set_contract_version(deps.storage, ACCOUNT, CONTRACT_VERSION)?;
+        return Ok(AccountResponse::action("migrate"));
+    }
+    // else this means that we are migrating from another contract, we assert it's `account` and create an abstract account
+    let current_contract_version = get_contract_version(deps.storage)?;
+    if current_contract_version.contract != "account" {
+        return Err(AccountError::Std(StdError::generic_err("Can't migrate something other than `abstract_account`  or `account` to `abstract_account`")));
```
Buckram123, line 43:
```suggestion
        return Err(AccountError::Std(StdError::generic_err("Can't migrate something other than `abstract::account`  or `account` to `abstract_account`")));
```

framework/contracts/account/src/migrate.rs
```diff
@@ -1,19 +1,121 @@
...
+    }
+
+    // Use CW2 to set the contract version, this is needed for migrations
     set_contract_version(deps.storage, ACCOUNT, CONTRACT_VERSION)?;
-    Ok(AccountResponse::action("migrate"))
+
+    let registry = RegistryContract::new(deps.api, &env)?;
```
Buckram123, line 49: Would it be possible to share a function for this and instantiate?

framework/packages/abstract-interface/src/deployment.rs
```diff
@@ -306,6 +306,10 @@ impl<Chain: CwEnv> Abstract<Chain> {
             blob: self.blob.clone(),
         }
     }
+
+    pub fn account_code_id(&self) -> Result<u64, AbstractInterfaceError> {
```
Buckram123, line 310: Do we want code_id method on "account" object instead? Where we load code_id for "abstract::account" from cw orch env state

Buckram123: Right now it will fail because contract id is different depending on account id

Buckram123 (changes requested):
Looks good, small details before we merge it

framework/contracts/account/Cargo.toml
Kayanski: Ah oups, yes, I was experimenting with new taplo settings and it just applied. Will revert (see this task https://linear.app/abstract-sdk/issue/ABS-583/change-taplo-base-configuration)

framework/contracts/account/src/migrate.rs
Kayanski: not really, I removed a lot of code that was un-necessary and all that. I might want to call instantiate directly though maybe ?

framework/packages/abstract-interface/src/deployment.rs
Kayanski: Yeah that's why I put that on Abstract. I wanted a way to get the account id that should be used for creating a new account. This function might want to load the latest code id from chain, that could be better I think

framework/contracts/account/src/migrate.rs
Kayanski: No that's not possible as the xion authenticator is in the way. Copying the code is the best solution here I think

framework/contracts/account/src/migrate.rs
Buckram123: @Kayanski sorry it's not fully resolved, I edited suggestion later

Buckram123 (approved):
LGTM

PR #537, opened 2024-11-15
Fix install modules during instantiation

Buckram123:
Account was trying to query self code_id, before getting created. On chain contract does not exist during instantiation, but it does exist in cw-multi-test's, so we didn't catch it with tests


framework/contracts/account/src/contract.rs
```diff
@@ -230,9 +230,9 @@ pub fn instantiate(
         // Install modules
         let (install_msgs, install_attribute) = _install_modules(
             deps.branch(),
-            &env,
             install_modules,
             simulate_resp.total_required_funds,
+            code_id,
```
Buckram123, line 235: tldr: we use code id from instantiate msg for modules install here. That's basically whole change

adairrr (approved):

adairrr:
LGTM 🛥️

PR #540, opened 2024-11-20
Change the security_disabled flag to !security_enabled

Buckram123:
security_disabled often makes double negative in discussions and in code, see more in ABS-584 description


framework/packages/abstract-interface/src/deployment.rs
```diff
@@ -220,9 +220,9 @@ impl<Chain: CwEnv> Abstract<Chain> {
             &abstract_std::registry::InstantiateMsg {
                 admin: admin.to_string(),
                 #[cfg(feature = "integration")]
-                security_disabled: Some(true),
+                security_enabled: Some(true),
```
Kayanski, line 223:
```suggestion
                security_enabled: Some(false),
```

framework/packages/abstract-interface/src/deployment.rs
```diff
@@ -220,9 +220,9 @@ impl<Chain: CwEnv> Abstract<Chain> {
...
                 #[cfg(not(feature = "integration"))]
-                security_disabled: Some(false),
+                security_enabled: Some(false),
```
Kayanski, line 225:
```suggestion
                security_enabled: Some(true),
```

Kayanski (approved):
Superb,
We also said we wanted to change the way the `integration` feature works no ? 
I don't remember us settling on a behavior though

PR #544, opened 2024-11-21
AccountTrace Serialization was done in the wrong order

Kayanski:
This Pr aims at correcting the display and parsing of account traces that was done in the wrong order.

It also changes the way accounts are serialized as keys.
Rationale: 
The way Prefixer and KeyDeserialize was implemented is not compatible with cosmwasm-std 2, especially for composite keys. This implementation, didn't take into account the fact that the length of Vec<Key> used during serialization needs to be fixed for composite keys to be correctly usable for ranging. 
In this Breaking PR, we fix this issue.

This PR is breaking on the account structure. Everything will break (see module migration testing for instance)

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Buckram123 (approved):
LGTM. But it is state breaking, right? I'm surprised to not see a snapshot file changed here

Kayanski:
> LGTM. But it is state breaking, right? I'm surprised to not see a snapshot file changed here

Because we haven't used multi-hop accounts for now in the repo.
And I think it's fine to push that as a patch because we haven't used it yet

framework/packages/abstract-std/src/objects/account/account_trace.rs
```diff
@@ -174,12 +174,13 @@ impl AccountTrace {
         let acc = if trace == LOCAL {
             Self::Local
         } else {
-            Self::Remote(
-                trace
-                    .split(CHAIN_DELIMITER)
-                    .map(TruncatedChainId::_from_str)
-                    .collect(),
-            )
+            let rev_trace: Vec<_> = trace
```
adairrr, line 177: What about the `from_string` method lines 155-167?

adairrr: (a method which also needs a test case)

adairrr (changes requested):

framework/packages/abstract-std/src/objects/account/account_trace.rs
Kayanski: Put things in common, test case covers everything

framework/packages/abstract-std/src/objects/account/account_trace.rs
```diff
@@ -22,6 +22,7 @@ impl KeyDeserialize for &AccountTrace {
 
     #[inline(always)]
     fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
+        let value = value.into_iter().filter(|b| *b > 32).collect();
```
Kayanski, line 25: This is good and all, but it feels like a hack that works with current functions. Can't we just serialize the AccountTrace to our format and then use a simple deserialization function.
We should also provide a better serialization function because the one we have currently is very weird for our use-case

Buckram123: > This is good, but it feels like a hack that works with current storage. We should also provide a better serialization function because the one we have currently is very weird for our use-case. For example, simply use the Display impl to ser-deser ?

Yes, as we discussed we shouldn't use this type in state and replace it with string

Kayanski: Another solution, for composite keys to work better, we could use a maximum trace length for accounts

framework/packages/abstract-std/src/objects/entry/contract_entry.rs
```diff
@@ -107,7 +107,7 @@ impl<'a> Prefixer<'a> for &ContractEntry {
 
 impl KeyDeserialize for &ContractEntry {
     type Output = ContractEntry;
-    const KEY_ELEMS: u16 = 1;
+    const KEY_ELEMS: u16 = 2;
```
Kayanski, line 110: These changes are mandatory if you want to work with composite keys for these structures with custom cw-storage-plus impl

adairrr (approved):

PR #545, opened 2024-11-25
Oracle Adapter

Kayanski:
Adding the Oracle Adapter to abstract supported adapters

Added Pyth implementation and tests

Based on https://www.notion.so/abstract-money/Price-Oracle-24cb3a64c535427d9a38b0312235f889

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


modules/contracts/adapters/oracle/STATUS.md
```diff
@@ -0,0 +1,11 @@
+# Adapter Status
```
adairrr, line 1: Needs updating

modules/contracts/adapters/oracle/tests/osmosis.rs
adairrr, line 1: Seems that there are quite a few leftover tests?

scripts/wasm-all.sh
```diff
@@ -9,18 +9,18 @@ fi
...
+# # Delete all the current wasms first
+# rm -rf ./artifacts/*.wasm
+# # Optimized builds
+# docker run --rm -v "$(pwd)":/code \
+#   --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+#   --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+#   ${image}
 
-cd $starting_dir
+# cd $starting_dir
```
adairrr, line 23: Why was this changed?

adairrr (changes requested):
The PR specifies that the implementation was performed with the API specified in the Oracle adapter design. 

However, it seems that asset registration and value queries were not implemented. Are those supposed to be implemented on this adapter or elsewhere?

scripts/wasm-all.sh
Kayanski: Oups, will revert

integrations/oracles/pyth/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+            &deps.querier,
+            pyth_address,
+            PriceIdentifier::from_hex(price_id.as_bytes())
+                .map_err(|e| StdError::generic_err(format!("Wrong price id hex format, {e}")))?,
+        )?;
+        let price_feed = price_feed_response.price_feed;
+
+        let current_price = price_feed
+            .get_price_no_older_than(env.block.time.seconds() as i64, no_older_than)
+            .ok_or_else(|| StdError::not_found("Current price is not available"))?;
```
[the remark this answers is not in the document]
Kayanski: The price must be not older than `no_older_than` from the current block time.

Kayanski:
> The PR specifies that the implementation was performed with the API specified in the Oracle adapter design.
> 
> However, it seems that asset registration and value queries were not implemented. Are those supposed to be implemented on this adapter or elsewhere?

To me, asset registration and value queries should not be placed inside the oracle adapter which is there to query prices from price sources using keys (for the moment).
It belongs in a value computation app for instance, because it contains account specific storage that is best handled within an App

Buckram123 (changes requested):
Amazing implementation! Great decision to keep it without state and execute methods, it would be make more sense to keep at module that adapts this adapter. Left few small nits.

Another point that oracle part of the account previously was able to return value of the account(list of assets) in "base asset" - it was evaluating assets/lp tokens, we don't need it anymore? I think it should be possible to get price of list assets/lp-tokens if every asset involved would be tagged by "price_source_key". Not sure that we need to calculate value in "base asset", IMO oracle price is enough, I imagine it would be quite challenging to create and use API that includes both "base asset" and oracle prices

modules/contracts/adapters/oracle/src/state.rs
```diff
@@ -0,0 +1,4 @@
+use abstract_oracle_standard::msg::Config;
+use cw_storage_plus::Item;
+
+pub const CONFIG: Item<Config> = Item::new("config");
```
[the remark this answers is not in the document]
Buckram123: Actually just remove it? It would be as breaking to add this query as to change what it contains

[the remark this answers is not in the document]
Kayanski: Removed

PR #546, opened 2024-11-26
Audit Fixes 1: XION Authentication fix and test

Kayanski:
This Pr aims at adding Auth permissions fix for xion (found during the audit by us)

Changes : 

- Account contract is now always whitelisted, it can execute on itself non-admin actions
- ownership::assert_nested_owner flag, now checks that the contract has admin execution rights in case of a XION Abstract Account
- Removed admin checks for xion in the whitelist function (not used anymore)
- Added a test to make sure it's not possible to bypass the admin check when calling from a module
- Added admin checks on the Auth Method admin methods (original audit issue)

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.

### DIFF
--- a/framework/contracts/account/src/execution.rs
+++ b/framework/contracts/account/src/execution.rs
@@ -10,27 +10,21 @@ use cosmwasm_std::{
 };
 
 use crate::{
-    config::assert_admin,
     contract::{AccountResponse, AccountResult, ADMIN_ACTION_REPLY_ID, FORWARD_RESPONSE_REPLY_ID},
     error::AccountError,
     modules::load_module_addr,
 };
 
 /// Check that sender either whitelisted or governance
-pub(crate) fn assert_whitelisted_or_owner(deps: &mut DepsMut, sender: &Addr) -> AccountResult<()> {
-    #[cfg(feature = "xion")]
-    {
-        if let Some(is_admin) = crate::state::AUTH_ADMIN.may_load(deps.storage)? {
-            // Clear auth if it was set
-            crate::state::AUTH_ADMIN.remove(deps.storage);
-            if is_admin {
-                return Ok(());
-            }
-        }
-    }
+pub(crate) fn assert_whitelisted_or_owner(
+    deps: &mut DepsMut,
+    env: &Env,
+    sender: &Addr,
+) -> AccountResult<()> {
     let whitelisted_modules = WHITELISTED_MODULES.load(deps.storage)?;
     if whitelisted_modules.0.contains(sender)
         || ownership::assert_nested_owner(deps.storage, &deps.querier, sender).is_ok()
+        || sender == env.contract.address
     {
         Ok(())
     } else {
@@ -42,10 +36,11 @@ pub(crate) fn assert_whitelisted_or_owner(deps: &mut DepsMut, sender: &Addr) -> 
 /// Permission: Module
 pub fn execute_msgs(
     mut deps: DepsMut,
+    env: Env,
     msg_sender: &Addr,
     msgs: Vec<CosmosMsg<Empty>>,
 ) -> AccountResult {
-    assert_whitelisted_or_owner(&mut deps, msg_sender)?;
+    assert_whitelisted_or_owner(&mut deps, &env, msg_sender)?;
 
     Ok(AccountResponse::action("execute_module_action").add_messages(msgs))
 }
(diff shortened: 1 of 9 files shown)


framework/contracts/account/src/execution.rs
```diff
@@ -16,20 +16,15 @@ use crate::{
...
-    }
+pub(crate) fn assert_whitelisted_or_owner(
+    deps: &mut DepsMut,
+    env: &Env,
+    sender: &Addr,
+) -> AccountResult<()> {
     let whitelisted_modules = WHITELISTED_MODULES.load(deps.storage)?;
     if whitelisted_modules.0.contains(sender)
         || ownership::assert_nested_owner(deps.storage, &deps.querier, sender).is_ok()
+        || sender == env.contract.address
```
adairrr, line 27: This adds a self check so we should update the fn name to include `or_self`.

framework/packages/abstract-std/src/objects/ownership/gov_ownable.rs
```diff
@@ -213,17 +215,39 @@ pub fn assert_nested_owner(
 ) -> Result<(), GovOwnershipError> {
     let ownership = OWNERSHIP.load(store)?;
     // If current sender is owner of this account - it's the owner
-    if ownership.assert_owner(querier, sender).is_ok() {
+    let owner_assertion = ownership.assert_owner(querier, sender);
+    if owner_assertion.is_ok() {
+        // If this is a xion abstract account, we need to make sure the admin flag is set
```
adairrr, line 220:
```suggestion
        // If this is a self-owned abstract account, we need to make sure the admin flag is set
```

framework/contracts/account/tests/xion.rs
```diff
@@ -123,3 +142,116 @@ fn test_init_sign_arb() {
...
+        pending_owner: None,
+        pending_expiry: None,
+    };
+    OWNERSHIP.save(deps.as_mut().storage, &ownership)?;
+
+    let whitelisted = deps.api.addr_make("whitelisted");
+    let not_whitelisted_yet = deps.api.addr_make("not_whitelisted");
+
+    // We whitelist a module
+    AUTH_ADMIN.save(deps.as_mut().storage, &true)?;
```
adairrr, line 200: Do we need to clear this admin flag before executing the not_whitelisted update messages?

framework/packages/abstract-std/src/objects/ownership/gov_ownable.rs
```diff
@@ -213,17 +215,39 @@ pub fn assert_nested_owner(
...
     // the contract must have an owner
-    top_level_ownership.assert_owner(querier, sender)
+    match top_level_ownership.assert_owner(querier, sender) {
+        Ok(_) => {
+            // If the top level owner is an abstract account, we need to make sure the admin flag is set
+            if let GovernanceDetails::AbstractAccount { address } = top_level_ownership.owner {
+                if let Ok(true) = AUTH_ADMIN.query(querier, address) {
+                    return Ok(());
+                } else {
+                    return Err(crate::objects::ownership::GovOwnershipError::NotOwner);
```
adairrr, line 244: Should we return a more specific error such as `NotAdmin`?

framework/contracts/account/src/execution.rs
adairrr: We remove the `AUTH_ADMIN.remove` call above. Does this change the semantics for authorizated executions on XION?

adairrr: Also - we should have a test that specifically hits this case where the account calls into itself.

adairrr (changes requested):

framework/contracts/account/src/execution.rs
Kayanski: We do have a test for that, in the only test that was added for this PR
Changing the name of the function
yes, the `AUTH_ADMIN.remove` call was a first idea on how to handle this variable that was not removed when we decided to place at the end of the execute function

framework/contracts/account/tests/xion.rs
Kayanski: The Admin flag is cleared at the end of the execute function

framework/packages/abstract-std/src/objects/ownership/gov_ownable.rs
Kayanski: Not really because this logic ensures that it's the owner calling the contract. If a module calls the account, then it's not the owner calling it. So I would stick with that

PR #554, opened 2024-11-27
Do not MERGE : Audit Fixes 10: Insufficient validation of the input URLs

Kayanski:
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


framework/packages/abstract-std/src/objects/validation/verifiers.rs
```diff
@@ -28,6 +30,8 @@ pub fn validate_link(link: Option<&str>) -> Result<(), ValidationError> {
             Err(ValidationError::LinkInvalidFormat {})
         } else if contains_dangerous_characters(link) {
             Err(ValidationError::LinkContainsDangerousCharacters {})
+        } else if let Err(e) = Url::parse(link) {
+            Err(ValidationError::LinkInvalidUrl(e))
```
adairrr, line 34: The reason we didn't do this Url verification from the original audit results last year was because of wasm size increase with minimal benefit. Would you mind checking the size differences before and after?

Kayanski: The increase is ~ 160K, this is huge !
We might not want to merge that PR

Kayanski:
Closed because there is a 160K increase in wasm size because of this change alone

PR #562, opened 2025-01-06
Only update and when update fields are non empty

Kayanski:
### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Buckram123 (approved):
Nice! Took small glance at implementations, and noticed that we can do another small gas optimization, on contract side. `update_dex_registry` loads REGISTERED_DEXES twice in case to_add and to_remove isn't empty.

Kayanski:
> Nice! Took small glance at implementations, and noticed that we can do another small gas optimization, on contract side. `update_dex_registry` loads REGISTERED_DEXES twice in case both `to_add` and `to_remove` isn't empty.

As this function won't be triggered too much, I think it's ok to keep it that way, plus it's nice right now haha

PR #563, opened 2025-01-07
Publish Abstract V0.26.0

Kayanski:
Xion has published their crate on crates.io, it's now possible to publish abstract 0.26.0

### Checklist

- [ ] CI is green.
- [ ] Changelog updated.


Buckram123 (approved):
Nice! Not sure what is going on with formatting of toml

framework/packages/abstract-std/src/account.rs
```diff
@@ -104,7 +104,11 @@ pub mod state {
 }
 
 #[cosmwasm_schema::cw_serde]
-pub struct MigrateMsg {}
+pub struct MigrateMsg {
+    /// This field provides the new code id of the contract
+    /// This is only necessary for migrations from XION accounts.
+    pub code_id: Option<u64>,
+}
```
Buckram123, line 111: Won't it be simpler to have it non-optional?

Buckram123 (approved):
Nice fixes for xion!

framework/packages/abstract-std/src/account.rs
Kayanski: Would it ?

Kayanski: Not really, because in a lot of places, the code id is not easily available inside our code. Plus having a field with information that is not used in the contract (for abstract->abstract migration), feels very weird

Buckram123:
@Kayanski Feel free to merge it, uploaded and registered this wasm in this #564

PR #564, opened 2025-02-14
xion-testnet-2 abstract deployment

Buckram123:


Buckram123:
@Kayanski Just noticed that #563 version of the 0.26.0 abstract is quite different from the tag release. Which version was supposed to be deployed?

Buckram123:
Uploaded account from #563 and registered it in register contract

Kayanski:
> @Kayanski Just noticed that #563 version of the 0.26.0 abstract is quite different from the tag release. Which version was supposed to be deployed?

Oh, really ? Version #563 should be used yes, I haven't updated the tagged release ?

interchain/scripts/state.json
```diff
@@ -82,5 +82,28 @@
...
+      "abstract:account": 95,
+      "abstract:ans-host": 4,
+      "abstract:ibc-client": 8,
+      "abstract:ibc-host": 9,
+      "abstract:module-factory": 6,
+      "abstract:registry": 5,
+      "cw:blob": 3,
+      "polytone:note": 10,
+      "polytone:proxy": 12,
+      "polytone:voice": 11
```
Buckram123, line 97:
```suggestion
      "cw:blob": 3
```
Oopsie polytone code ids slipped here

Kayanski (approved):
Ok, perfect, could you update the sscache CI that is failing here ? It's just a question of changing the sccache version I think (https://github.com/PyO3/maturin-action/issues/338)

PR #567, opened 2025-04-24
Support multiple polytone evm connections on ica client

Buckram123:
This PR aims to add support for multiple polytone connections on ica-client.

Implementation inspired by ibc-client

### DIFF
--- a/framework/contracts/native/ica-client/src/chain_types/evm.rs
+++ b/framework/contracts/native/ica-client/src/chain_types/evm.rs
@@ -1,12 +1,9 @@
 use abstract_ica::EVM_NOTE_ID;
 use abstract_sdk::{
-    feature_objects::{AnsHost, RegistryContract},
-    Resolve,
+    feature_objects::{AnsHost, RegistryContract},  Resolve
 };
 use abstract_std::{
-    ibc::PACKET_LIFETIME,
-    native_addrs,
-    objects::{module::ModuleInfo, ChannelEntry, ContractEntry, TruncatedChainId},
+    ibc::PACKET_LIFETIME, ica_client::state::ICA_INFRA, native_addrs, objects::{module::ModuleInfo, ChannelEntry, ContractEntry, TruncatedChainId}
 };
 use cosmwasm_std::{
     wasm_execute, Addr, Binary, Coin, CosmosMsg, Deps, Env, HexBinary, QuerierWrapper, WasmMsg,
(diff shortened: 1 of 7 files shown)


Buckram123:
Up for discussion: It seems more common to use chain id as number instead of chain-name for evm chains, do we want to replace `chain: TruncatedChainId` parts to
```rust
#[non_exhaustive]
enum IcaChainId {
  Cosmos(TruncatedChainId),
  Evm(u64),
}
```

@Kayanski @adairrr ?

Kayanski:
> Up for discussion: It seems more common to use chain id as number instead of chain-name for evm chains, do we want to replace `TruncatedChainId` parts to
> 
> ```rust
> #[non_exhaustive]
> enum IcaChainId {
>   Cosmos(TruncatedChainId),
>   Evm(u64),
> }
> ```
> 
> @Kayanski @adairrr ?

That's a good idea, Are we using this ICA Client for other purposes than EVM chains though ? Is it necessary to have Cosmos chains in there too ? We do have the IBC Client for that I think already ?

Buckram123:
> > Up for discussion: It seems more common to use chain id as number instead of chain-name for evm chains, do we want to replace `TruncatedChainId` parts to
> > ```rust
> > #[non_exhaustive]
> > enum IcaChainId {
> >   Cosmos(TruncatedChainId),
> >   Evm(u64),
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @Kayanski @adairrr ?
> 
> That's a good idea, Are we using this ICA Client for other purposes than EVM chains though ? Is it necessary to have Cosmos chains in there too ? We do have the IBC Client for that I think already ? Otherwise, that's a very nice idea to have the ICA client work like the IBC client !

You are right, we're only using it for evm chains right now. It could be good to keep it as a multi-chain client as it has more functionality beyond helping evm executions and it is quite different from ibc client: each account controls it's own proxy on remote chain

Buckram123:
> > > Up for discussion: It seems more common to use chain id as number instead of chain-name for evm chains, do we want to replace `TruncatedChainId` parts to
> > > ```rust
> > > #[non_exhaustive]
> > > enum IcaChainId {
> > >   Cosmos(TruncatedChainId),
> > >   Evm(u64),
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > @Kayanski @adairrr ?
> > 
> > 
> > That's a good idea, Are we using this ICA Client for other purposes than EVM chains though ? Is it necessary to have Cosmos chains in there too ? We do have the IBC Client for that I think already ? Otherwise, that's a very nice idea to have the ICA client work like the IBC client !
> 
> You are right, we're only using it for evm chains right now. It could be good to keep it as a multi-chain client as it has more functionality beyond helping evm executions and it is quite different from ibc client: each account controls it's own proxy on remote chain

Moved discussion into ABS-637

